Issue #1182, opened 2019-05-31
Coin should support deflationary policy for third party dappchains


mattkanwisher:
For non loom chains, several people have asked for bitcoin style deflationary scheme. So basically in the genesis coin should support an inflation scheme, that changes over time. 
So for example, someone could configure their chain to have 10% inflation first year, 5% second. Etc . Ideally they just configure the windows of time for rebalance.

mattkanwisher:
Looks like we just have 1 PR Left?
Do we have example Solidity contracts for ETH Mainnet for reference for people.
When do we think we can get this onto a testate?

enlight:
Solidity contracts? What are you talking about?

mattkanwisher:
Perhaps interaction with mainnet solidity isn’t needed in this case

suhasagg:
PR for the same - 
https://github.com/loomnetwork/loomchain/pull/1224

Go-loom PR for the same  -
https://github.com/loomnetwork/go-loom/pull/395

PR #1224, opened 2019-06-13
Coin policy manager

suhasagg:

### DIFF
--- a/app.go
+++ b/app.go
@@ -6,15 +6,16 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/go-kit/kit/metrics"
-	kitprometheus "github.com/go-kit/kit/metrics/prometheus"
-	"github.com/loomnetwork/go-loom"
-	"github.com/loomnetwork/go-loom/plugin"
-	"github.com/loomnetwork/go-loom/types"
 	"github.com/loomnetwork/go-loom/util"
 	"github.com/loomnetwork/loomchain/eth/utils"
-	"github.com/loomnetwork/loomchain/log"
 	"github.com/loomnetwork/loomchain/registry"
+
+	"github.com/go-kit/kit/metrics"
+	kitprometheus "github.com/go-kit/kit/metrics/prometheus"
+	loom "github.com/loomnetwork/go-loom"
+	"github.com/loomnetwork/go-loom/plugin"
+	"github.com/loomnetwork/go-loom/types"
+	"github.com/loomnetwork/loomchain/log"
 	"github.com/loomnetwork/loomchain/store"
 	blockindex "github.com/loomnetwork/loomchain/store/block_index"
 	evmaux "github.com/loomnetwork/loomchain/store/evm_aux"
@@ -253,11 +254,6 @@ type QueryHandler interface {
 	Handle(state ReadOnlyState, path string, data []byte) ([]byte, error)
 }
 
-type OriginHandler interface {
-	ValidateOrigin(input []byte, chainId string, currentBlockHeight int64) error
-	Reset(currentBlockHeight int64)
-}
-
 type KarmaHandler interface {
 	Upkeep() error
 }
@@ -271,17 +267,11 @@ type ChainConfigManager interface {
 	EnableFeatures(blockHeight int64) error
 }
 
-type CoinPolicyManager interface {
-	MintCoins() error
-}
-
 type GetValidatorSet func(state State) (loom.ValidatorSet, error)
 
 type ValidatorsManagerFactoryFunc func(state State) (ValidatorsManager, error)
 
 type ChainConfigManagerFactoryFunc func(state State) (ChainConfigManager, error)
-
-type CoinPolicyManagerFactoryFunc func(state State) (CoinPolicyManager, error)
 
 type Application struct {
 	lastBlockHeader abci.Header
@@ -297,8 +287,6 @@ type Application struct {
 	blockindex.BlockIndexStore
 	CreateValidatorManager   ValidatorsManagerFactoryFunc
 	CreateChainConfigManager ChainConfigManagerFactoryFunc
-	CreateCoinPolicyManager  CoinPolicyManagerFactoryFunc
-	OriginHandler
 	// Callback function used to construct a contract upkeep handler at the start of each block,
 	// should return a nil handler when the contract upkeep feature is disabled.
 	CreateContractUpkeepHandler func(state State) (KarmaHandler, error)
@@ -428,8 +416,6 @@ func (a *Application) BeginBlock(req abci.RequestBeginBlock) abci.ResponseBeginB
 		}
 	}
 
-	a.OriginHandler.Reset(a.curBlockHeader.Height)
-
 	storeTx := store.WrapAtomic(a.Store).BeginTx()
 	state := NewStoreState(
 		context.Background(),
@@ -462,15 +448,6 @@ func (a *Application) BeginBlock(req abci.RequestBeginBlock) abci.ResponseBeginB
 		}
 	}
 
-	coinPolicyManager, err := a.CreateCoinPolicyManager(state)
-	if err != nil {
-		panic(err)
-	}
-	if coinPolicyManager != nil {
-		if err := coinPolicyManager.MintCoins(); err != nil {
-			panic(err)
-		}
-	}
 	storeTx.Commit()
 
 	return abci.ResponseBeginBlock{}
@@ -601,14 +578,6 @@ func (a *Application) processTx(txBytes []byte, isCheckTx bool) (TxHandlerResult
 		a.curBlockHash,
 		a.GetValidatorSet,
 	)
-
-	if isCheckTx {
-		err := a.OriginHandler.ValidateOrigin(txBytes, state.Block().ChainID, state.Block().Height)
-		if err != nil {
-			storeTx.Rollback()
-			return TxHandlerResult{}, err
-		}
-	}
 
 	receiptHandler, err := a.ReceiptHandlerProvider.StoreAt(a.height(), state.FeatureEnabled(EvmTxReceiptsVersion2Feature, false))
 	if err != nil {
--- a/app.go
+++ b/app.go
@@ -268,7 +268,7 @@ type ChainConfigManager interface {
 }
 
 type CoinPolicyManager interface {
-	MintCoins() error
+	ApplyPolicy() error
 }
 
 type GetValidatorSet func(state State) (loom.ValidatorSet, error)
@@ -460,7 +460,7 @@ func (a *Application) BeginBlock(req abci.RequestBeginBlock) abci.ResponseBeginB
 		panic(err)
 	}
 	if coinPolicyManager != nil {
-		if err := coinPolicyManager.MintCoins(); err != nil {
+		if err := coinPolicyManager.ApplyPolicy(); err != nil {
 			panic(err)
 		}
 	}
--- a/app.go
+++ b/app.go
@@ -267,11 +267,17 @@ type ChainConfigManager interface {
 	EnableFeatures(blockHeight int64) error
 }
 
+type CoinPolicyManager interface {
+	MintCoins() error
+}
+
 type GetValidatorSet func(state State) (loom.ValidatorSet, error)
 
 type ValidatorsManagerFactoryFunc func(state State) (ValidatorsManager, error)
 
 type ChainConfigManagerFactoryFunc func(state State) (ChainConfigManager, error)
+
+type CoinPolicyManagerFactoryFunc func(state State) (CoinPolicyManager, error)
 
 type Application struct {
 	lastBlockHeader abci.Header
@@ -287,6 +293,7 @@ type Application struct {
 	blockindex.BlockIndexStore
 	CreateValidatorManager   ValidatorsManagerFactoryFunc
 	CreateChainConfigManager ChainConfigManagerFactoryFunc
+	CreateCoinPolicyManager  CoinPolicyManagerFactoryFunc
 	// Callback function used to construct a contract upkeep handler at the start of each block,
 	// should return a nil handler when the contract upkeep feature is disabled.
 	CreateContractUpkeepHandler func(state State) (KarmaHandler, error)
@@ -444,6 +451,16 @@ func (a *Application) BeginBlock(req abci.RequestBeginBlock) abci.ResponseBeginB
 	}
 	if chainConfigManager != nil {
 		if err := chainConfigManager.EnableFeatures(a.height()); err != nil {
+			panic(err)
+		}
+	}
+
+	coinPolicyManager, err := a.CreateCoinPolicyManager(state)
+	if err != nil {
+		panic(err)
+	}
+	if coinPolicyManager != nil {
+		if err := coinPolicyManager.MintCoins(); err != nil {
 			panic(err)
 		}
 	}
--- a/builtin/plugins/coin/coin.go
+++ b/builtin/plugins/coin/coin.go
@@ -355,7 +355,7 @@ func Mint(ctx contract.Context) error {
 	if mintingAmount.Value.Cmp(loom.NewBigUIntFromInt(0)) == 0 {
 		return nil // No more coins to be minted on block creation
 	}
-	return mint(ctx, loom.UnmarshalAddressPB(policy.MintingAccount), amount)
+	return mint(ctx, loom.UnmarshalAddressPB(policy.MintingAccount), loom.NewBigUIntFromInt(mintingAmount.Value.Int64()))
 }
 
 //First year is started from the checkpoint at which minting started
--- a/builtin/plugins/coin/coin.go
+++ b/builtin/plugins/coin/coin.go
@@ -260,11 +260,11 @@ func Mint(ctx contract.Context) error {
 	totalSupply := loom.NewBigUIntFromInt(int64(policy.TotalSupply))
 	blocksGeneratedPerYear := loom.NewBigUIntFromInt(int64(policy.BlocksGeneratedPerYear))
 	year := blockHeight.Div(blockHeight, blocksGeneratedPerYear)
-	if year == loom.NewBigUIntFromInt(0) {
+	year = year.Add(year, loom.NewBigUIntFromInt(1))
+	if year == loom.NewBigUIntFromInt(1) {
 		amount = totalSupply.Div(totalSupply, blocksGeneratedPerYear)
 	} else {
-		changeRatioNumerator = changeRatioNumerator.Exp(changeRatioNumerator, year, nil)
-		changeRatioDenominator = changeRatioDenominator.Exp(changeRatioDenominator, year, nil)
+		changeRatioDenominator = changeRatioDenominator.Mul(changeRatioDenominator, year)
 		totalSupplyForYear := totalSupply.Mul(totalSupply, changeRatioNumerator)
 		totalSupplyForYear = totalSupplyForYear.Div(totalSupplyForYear, changeRatioDenominator)
 		amount = totalSupplyForYear.Div(totalSupplyForYear, blocksGeneratedPerYear)
--- a/builtin/plugins/coin/coin.go
+++ b/builtin/plugins/coin/coin.go
@@ -284,7 +284,7 @@ func Mint(ctx contract.Context) error {
 		return err
 	}
 	if err == contract.ErrNotFound {
-	//Sets Minting Height - Height at which minting started
+		//Sets Minting Height - Height at which minting started
 		err := ctx.Set(mintingHeightKey, &types.BigUInt{
 			Value: *loom.NewBigUIntFromInt(ctx.Block().Height),
 		})
@@ -319,25 +319,16 @@ func Mint(ctx contract.Context) error {
 	} else {
 		if modulus.Cmp(big.NewInt(1)) == 0 && err != contract.ErrNotFound {
 			//Operator Support to support different inflation ratio patterns for different year
-			if strings.EqualFold("div", operator) {
-				amount, err = ComputeforConsecutiveYearBeginningDivOperator(ctx, baseAmount, changeRatioNumerator,
-					changeRatioDenominator, basePercentage, blocksGeneratedPerYear, amount, loom.NewBigUIntFromInt(relativeYear.Int64()))
-				if err != nil {
-					return errUtil.Wrap(err, "Failed Minting Block with div operator for consecutive year")
-				}
-			} else if strings.EqualFold("exp", operator) {
-				amount, err = ComputeforConsecutiveYearBeginningExpOperator(ctx, baseAmount, changeRatioNumerator,
-					changeRatioDenominator, basePercentage, blocksGeneratedPerYear, amount, loom.NewBigUIntFromInt(relativeYear.Int64()))
-				if err != nil {
-					return errUtil.Wrap(err, "Failed Minting Block Configuration with exp operator for consecutive year")
-				}
-			} else {
-				return errors.New("Invalid operator - Operator should be div or exp")
+			amount, err = ComputeforConsecutiveYearBeginningWithOperator(ctx, baseAmount, changeRatioNumerator,
+				changeRatioDenominator, basePercentage, blocksGeneratedPerYear, amount,
+				loom.NewBigUIntFromInt(relativeYear.Int64()), operator)
+			if err != nil {
+				return errUtil.Wrap(err, "Failed Minting Block with operator for consecutive year")
 			}
 		} else if err == contract.ErrNotFound {
 			amount, err = ComputeforFirstYearBlockHeightgreaterthanOneyear(ctx, baseAmount, basePercentage, blocksGeneratedPerYear)
 			if err != nil {
-				return errUtil.Wrap(err, "Failed Minting Block Configuration for exp operator")
+				return errUtil.Wrap(err, "Failed Minting Block for First Year BlockHeight greater than One year")
 			}
 		} else {
 			amount, err = ComputeforConsecutiveYearinMiddle(ctx)
@@ -396,35 +387,29 @@ func ComputeforFirstYearBlockHeightgreaterthanOneyear(ctx contract.Context, base
 	return amount, nil
 }
 
-//Computes minting amount per year, after applying change ratio to base percentage using div operator
-//Computes inflation by multiplying base Amount with base percentage. Base percentage is computed by multiplying base percentage with change ratio on which div operator is applied according to the year.
-func ComputeforConsecutiveYearBeginningDivOperator(ctx contract.Context, baseAmount *common.BigUInt, changeRatioNumerator *common.BigUInt,
-	changeRatioDenominator *common.BigUInt, basePercentage *common.BigUInt, blocksGeneratedPerYear *common.BigUInt, amount *common.BigUInt, year *common.BigUInt) (*common.BigUInt, error) {
-	changeRatioDenominator = changeRatioDenominator.Mul(changeRatioDenominator, year)
-	basePercentage = basePercentage.Mul(basePercentage, changeRatioNumerator)
-	inflationforYear, err := ComputeInflationForYear(ctx, baseAmount, changeRatioDenominator, basePercentage)
-	if err != nil {
-		return nil, err
-	}
-	amount = inflationforYear.Div(inflationforYear, blocksGeneratedPerYear)
-	//Minting Amount per block computed starting from beginning block of that year and saved to avoid re-computations
-	err = ctx.Set(mintingAmountKey, &types.BigUInt{
-		Value: *amount,
-	})
-	if err != nil {
-		return nil, err
-	}
-	return amount, nil
-}
-
-//Computes minting amount per year, after applying change ratio to base percentage using exp operator
-//Computes inflation by multiplying base Amount with base percentage. Base percentage is computed by multiplying base percentage with change ratio on which exp operator is applied according to the year
-func ComputeforConsecutiveYearBeginningExpOperator(ctx contract.Context, baseAmount *common.BigUInt, changeRatioNumerator *common.BigUInt,
+func ComputeforConsecutiveYearBeginningWithOperator(ctx contract.Context, baseAmount *common.BigUInt,
+	changeRatioNumerator *common.BigUInt,
 	changeRatioDenominator *common.BigUInt, basePercentage *common.BigUInt, blocksGeneratedPerYear *common.BigUInt,
-	amount *common.BigUInt, year *common.BigUInt) (*common.BigUInt, error) {
-	changeRatioNumerator = changeRatioNumerator.Exp(changeRatioNumerator, year, nil)
-	changeRatioDenominator = changeRatioDenominator.Exp(changeRatioDenominator, year, nil)
-	basePercentage = basePercentage.Mul(basePercentage, changeRatioNumerator)
+	amount *common.BigUInt, year *common.BigUInt, operator string) (*common.BigUInt, error) {
+	var err error
+	switch operator {
+	//Computes minting amount per year, after applying change ratio to base percentage using div operator
+	//Computes inflation by multiplying base Amount with base percentage. Base percentage is computed by multiplying base percentage with change ratio on which div operator is applied according to the year.
+	case "div":
+		changeRatioDenominator = changeRatioDenominator.Mul(changeRatioDenominator, year)
+		basePercentage = basePercentage.Mul(basePercentage, changeRatioNumerator)
+	//Computes minting amount per year, after applying change ratio to base percentage using exp operator
+	//Computes inflation by multiplying base Amount with base percentage. Base percentage is computed by multiplying base percentage with change ratio on which exp operator is applied according to the year
+	case "exp":
+		changeRatioNumerator = changeRatioNumerator.Exp(changeRatioNumerator, year, nil)
+		changeRatioDenominator = changeRatioDenominator.Exp(changeRatioDenominator, year, nil)
+		basePercentage = basePercentage.Mul(basePercentage, changeRatioNumerator)
+	default:
+		err = fmt.Errorf("unrecognised operator %v", operator)
+	}
+	if err != nil {
+		return nil, err
+	}
 	inflationforYear, err := ComputeInflationForYear(ctx, baseAmount, changeRatioDenominator, basePercentage)
 	if err != nil {
 		return nil, err
@@ -438,7 +423,6 @@ func ComputeforConsecutiveYearBeginningExpOperator(ctx contract.Context, baseAmo
 		return nil, err
 	}
 	return amount, nil
-
 }
 
 //As minting amount has been computed for year, just derives it from context to save re-computation
--- a/builtin/plugins/coin/coin.go
+++ b/builtin/plugins/coin/coin.go
@@ -352,8 +352,7 @@ func Mint(ctx contract.Context) error {
 //First year is started from the checkpoint at which minting started
 //Computes minting Amount for first year if Blocks generated are less than a year, whether its year beginning or in the middle
 //Divides BaseAmountforfirstyear/Blocks generated per year to get minting Amount per block
-func ComputeforFirstYear(ctx contract.Context, baseAmount *common.BigUInt,
-	blocksGeneratedPerYear *common.BigUInt) (*common.BigUInt, error) {
+func ComputeforFirstYear(ctx contract.Context, baseAmount *common.BigUInt, blocksGeneratedPerYear *common.BigUInt) (*common.BigUInt, error) {
 	amount := baseAmount.Div(baseAmount, blocksGeneratedPerYear)
 	err := ctx.Set(mintingAmountKey, &types.BigUInt{
 		Value: *amount,
@@ -366,8 +365,7 @@ func ComputeforFirstYear(ctx contract.Context, baseAmount *common.BigUInt,
 
 //Computes minting Amount for first year if Block Height is greater than blocks generated in a year, whether its year beginning or in the middle
 //Computes base Amount by taking minting height into Account (Height at which minting started), applies base percentage to compute inflation for that year which is divided by blocks generated per year to get ==> minting Amount per block
-func ComputeforFirstYearBlockHeightgreaterthanOneyear(ctx contract.Context, baseAmount *common.BigUInt,
-	basePercentage *common.BigUInt, blocksGeneratedPerYear *common.BigUInt) (*common.BigUInt, error) {
+func ComputeforFirstYearBlockHeightgreaterthanOneyear(ctx contract.Context, baseAmount *common.BigUInt, basePercentage *common.BigUInt, blocksGeneratedPerYear *common.BigUInt) (*common.BigUInt, error) {
 	var econ Economy
 	err := ctx.Get(economyKey, &econ)
 	if err != nil {
@@ -387,8 +385,7 @@ func ComputeforFirstYearBlockHeightgreaterthanOneyear(ctx contract.Context, base
 	return amount, nil
 }
 
-func ComputeforConsecutiveYearBeginningWithOperator(ctx contract.Context, baseAmount *common.BigUInt,
-	changeRatioNumerator *common.BigUInt, changeRatioDenominator *common.BigUInt, basePercentage *common.BigUInt, blocksGeneratedPerYear *common.BigUInt, amount *common.BigUInt, year *common.BigUInt, operator string) (*common.BigUInt, error) {
+func ComputeforConsecutiveYearBeginningWithOperator(ctx contract.Context, baseAmount *common.BigUInt, changeRatioNumerator *common.BigUInt, changeRatioDenominator *common.BigUInt, basePercentage *common.BigUInt, blocksGeneratedPerYear *common.BigUInt, amount *common.BigUInt, year *common.BigUInt, operator string) (*common.BigUInt, error) {
 	var err error
 	switch operator {
 	//Computes minting amount per year, after applying change ratio to base percentage using div operator
--- a/builtin/plugins/coin/coin.go
+++ b/builtin/plugins/coin/coin.go
@@ -290,7 +290,7 @@ func Mint(ctx contract.Context) error {
 	if year.Cmp(loom.NewBigUIntFromInt(1)) == 0 {
 		//Determines minting amount at the beginning block of year or at block height at which minting is enabled
 		if modulus.Cmp(big.NewInt(1)) == 0 || err == contract.ErrNotFound {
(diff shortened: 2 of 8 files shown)


migrations/coin_policy_migration.go
```diff
@@ -0,0 +1,88 @@
...
+func GenerateCoinPolicyMigrationFn(cfg *config.Config) func(ctx *MigrationContext) error {
+	return func(ctx *MigrationContext) error {
+		//Resolve coin context
+		_, coinCtx, err := resolveCoin(ctx)
+		if err != nil {
+			return err
+		}
+		div := loom.NewBigUIntFromInt(10)
+		div.Exp(div, loom.NewBigUIntFromInt(18), nil)
+		if len(cfg.CoinPolicyMigrationConfig.MintingAccount) == 0 {
```
ParthDesai: Also, check if `cfg.CoinPolicyMigrationConfig` is nil

plugin/coin_policy_manager.go
```diff
@@ -0,0 +1,52 @@
...
+type CoinPolicyManager struct {
+	ctx   contract.Context
+	state loomchain.State
+}
+
+type Policy = ctypes.Policy
+
+// NewCoinPolicyManager attempts to create an instance of CoinPolicyManager.
+func NewCoinPolicyManager(pvm *PluginVM, state loomchain.State) (*CoinPolicyManager,
+	error) {
```
ParthDesai: keep `error)` in same line as function signature

migrations/coin_policy_migration.go
suhasagg: Done.

plugin/coin_policy_manager.go
suhasagg: Done.

migrations/coin_policy_migration.go
ParthDesai: @suhasagg hmm. This validation code has become too long. Let's do one thing, move it as a method for `CoinPolicyMigrationConfig` as in 

```
func (c *CoinPolicyMigrationConfig) IsValid() error {
   // Validate all fields here
}
```

and in here, we only validate if `cfg.CoinPolicyMigrationConfig` is nil or not.

config/config.go
```diff
@@ -104,6 +104,8 @@ type Config struct {
 	//Prometheus
 	PrometheusPushGateway *PrometheusPushGatewayConfig
 
+	CoinPolicyMigrationConfig *CoinPolicyMigrationConfig
```
ParthDesai: Rename this field to `CoinPolicyMigration`

suhasagg: Done.

builtin/plugins/coin/coin_test.go
```diff
@@ -34,6 +35,105 @@ func (m *mockLoomCoinGateway) DummyMethod(ctx contractpb.Context, req *MintToGat
...
+				Owner:   addr1.MarshalPB(),
+				Balance: uint64(31),
+			},
+		},
+		Policy: &Policy{
+			DeflationFactorDenominator: 5,
+			DeflationFactorNumerator:   1,
+			MintingAccount:             addr1.MarshalPB(),
+		},
+		BaseMintingAmount: 50,
```
ParthDesai: I thought `BaseMinitingAccount` is inside Policy data structure?

suhasagg: Base Minting Amount should be of type uint64 when loaded from genesis.
So in init request it should be outside policy data structure.
In policy it is of type BigUInt, so at the time of policy save in context, it is converted to BigUInt and saved.

ParthDesai: No, we need to get `BaseMintingAmount` inside `Policy` data structure. It will be inconsistent to have `BaseMintingAmount` outside, since it is logically part of `Policy`.

Plus, BaseMinting amount probably wont become that large that it wont fit uint64. 

So, let's put `BaseMintingAmount` in `Policy` and make it uint64.

Also, no need to check `< 0` for uint64, only `== 0` should suffice.

config/config.go
```diff
@@ -177,6 +181,14 @@ type PrometheusPushGatewayConfig struct {
 	JobName           string
 }
 
+type CoinPolicyMigrationConfig struct {
+	Enabled                    bool   //Enable Supply of DeflationInfo for modification of existing configuration
+	DeflationFactorNumerator   uint64 //Deflation Factor Numerator
+	DeflationFactorDenominator uint64 //Deflation Factor Denominator
+	BaseMintingAmount          int64  //Base Minting Amount
```
ParthDesai: Convert this to `uint64`

builtin/plugins/coin/coin_test.go
suhasagg: Policy.BaseMintingAMount should be of format - BIgUint.
Overtime, this BaseMintingAMount will aggregate and will become equal to total supply which should be of format - BigUInt 

This is the current - Policy data structure -
message Policy {
    uint64 deflation_factor_numerator = 1 [(gogoproto.customname)="DeflationFactorNumerator"];
    uint64 deflation_factor_denominator = 2 [(gogoproto.customname)="DeflationFactorDenominator"];
    BigUInt base_minting_amount = 3 [(gogoproto.customname)="BaseMintingAmount"];
    Address minting_account = 4 [(gogoproto.customname)="MintingAccount"];
} 

All these calculations should be of format BigUInt- 
policyNumerator := loom.NewBigUIntFromInt(int64(policy.DeflationFactorNumerator))
policyDenominator := loom.NewBigUIntFromInt(int64(policy.DeflationFactorDenominator))
blockHeight := loom.NewBigUIntFromInt(ctx.Block().Height)
depreciation.Mul(policyNumerator, blockHeight)
depreciation.Div(depreciation, policyDenominator)
amount := policy.BaseMintingAmount.Value
amount.Sub(&amount, depreciation)

If I use uint64 for Policy.BaseMintingAmount, I will have to convert to BigUInt in mint function.
``` 
//Conversion to BigUint required here - 
amount := policy.BaseMintingAmount.Value
amount.Sub(&amount, depreciation)
```

Solution 1 -
1)Keep BaseMintingAmount as uint64 in init request to facilitate data load.
2)Keep BaseMintingAmount as as BigUint in policy.

Solution 2 - 
Keep BaseMintingAmount as unit64 in policy and do conversion to BigUint in Mint Function.

config/config.go
suhasagg: Done.

builtin/plugins/coin/coin_test.go
suhasagg: Which Solution will you prefer.

builtin/plugins/coin/coin.go
```diff
@@ -213,6 +253,38 @@ func mint(ctx contract.Context, to loom.Address, amount *loom.BigUInt) error {
...
+		return errUtil.Wrap(err, "Failed to Get PolicyInfo")
+	}
+	var depreciation *common.BigUInt
+	policyNumerator := loom.NewBigUIntFromInt(int64(policy.DeflationFactorNumerator))
+	policyDenominator := loom.NewBigUIntFromInt(int64(policy.DeflationFactorDenominator))
+	blockHeight := loom.NewBigUIntFromInt(ctx.Block().Height)
+	depreciation.Mul(policyNumerator, blockHeight)
+	depreciation.Div(depreciation, policyDenominator)
+	amount := policy.BaseMintingAmount.Value
+	amount.Sub(&amount, depreciation)
```
ParthDesai: Only mint if `amount` is greater than `depreciation`

builtin/plugins/coin/coin_test.go
ParthDesai: We will go with solution 2. Keep in mind that while converting do not multiply by 10^18.

builtin/plugins/coin/coin.go
```diff
@@ -213,6 +249,25 @@ func mint(ctx contract.Context, to loom.Address, amount *loom.BigUInt) error {
...
+	if err != nil {
+		return errUtil.Wrap(err, "Failed to Get PolicyInfo")
+	}
+	var depreciation *common.BigUInt
+	policyNumerator := loom.NewBigUIntFromInt(int64(policy.DeflationFactorNumerator))
+	policyDenominator := loom.NewBigUIntFromInt(int64(policy.DeflationFactorDenominator))
+	blockHeight := loom.NewBigUIntFromInt(ctx.Block().Height)
+	depreciation.Mul(policyNumerator, blockHeight)
+	depreciation.Div(depreciation, policyDenominator)
+	//amount := policy.BaseMintingAmount
```
ParthDesai: Why did you comment out this? 

I said only mint, if `amount` is greater than `depreciation`.

migrations/coin_policy_migration.go
```diff
@@ -0,0 +1,64 @@
+package migrations
+
+import "C"
```
ParthDesai: Remove this.

builtin/plugins/coin/coin.go
suhasagg: It is a temp change for code to compile as Policy data type was modified :), mukul will complete it in short while.

mukuls9971: @ParthDesai he must have done it to compile it, I am completing it

migrations/coin_policy_migration.go
suhasagg: Done.

builtin/plugins/coin/coin.go
```diff
@@ -76,7 +79,40 @@ func (c *Coin) Meta() (plugin.Meta, error) {
 func (c *Coin) Init(ctx contract.Context, req *InitRequest) error {
 	div := loom.NewBigUIntFromInt(10)
 	div.Exp(div, loom.NewBigUIntFromInt(18), nil)
-
+	//Checks if Coin Policy Feature is enabled before loading Coin Monetary Supply Policy
+	if ctx.FeatureEnabled(loomchain.CoinPolicyFeature, false) {
```
enlight: The `Init` only runs once when the contract is deployed, there's not much point in checking for this feature flag here. Better to just skip these validation checks & skip calling `ctx.Set(policyKey, policy)` if `req.Policy == nil`.

plugin/coin_policy_manager.go
```diff
@@ -0,0 +1,51 @@
...
+	ctx := contract.WrapPluginContext(pvm.CreateContractContext(caller, contractAddr, readOnly))
+	return &CoinPolicyManager{
+		ctx:   ctx,
+		state: state,
+	}, nil
+}
+
+//MintCoins method of coin_deflation_Manager will be called from Block
+func (c *CoinPolicyManager) MintCoins() error {
+	err := coin.Mint(c.ctx)
```
enlight: Just `return coin.Mint(c.ctx)`

migrations/coin_policy_migration.go
```diff
@@ -0,0 +1,63 @@
...
+
+type (
+	Policy = ctypes.Policy
+)
+
+var (
+	policyKey = []byte("policy")
+)
+
+func GenerateCoinPolicyMigrationFn(cfg *config.Config) func(ctx *MigrationContext) error {
```
enlight: The input parameters for the migration should be part of the migration tx, you shouldn't be reading the config here. We'll need to update the migration tx handler to support migration input parameters though.

features.go
```diff
@@ -53,9 +53,13 @@ const (
 	// Enables Coin v1.1 contract (also applies to ETHCoin)
 	CoinVersion1_1Feature = "coin:v1.1"
 
+	// Enables Coin Policy Manager Feature
+	CoinPolicyFeature = "coin:policy"
```
enlight: -> `CoinVersion1_2Feature = "coin:v1.2"

builtin/plugins/coin/coin.go
suhasagg: Done

plugin/coin_policy_manager.go
suhasagg: Done

migrations/coin_policy_migration.go
suhasagg: Done

features.go
suhasagg: Done.

migrations/coin_policy_migration.go
suhasagg: Migration Input Parameters are supplied as CLI arguments

cmd/loom/sendtx.go
```diff
@@ -50,16 +55,30 @@ func newMigrationCommand() *cobra.Command {
 			if callerChainID == "" {
 				callerChainID = cli.TxFlags.ChainID
 			}
-			return migrationTx(Id, cli.TxFlags.PrivFile, cli.TxFlags.Algo, callerChainID)
+			if Id == 2 {
+				policy := Policy{}
+				jsonData, err := ioutil.ReadFile(inputFile)
+				err = jsonpb.UnmarshalString(string(jsonData), &policy)
```
ParthDesai: Handle error in case you are not able to read the file.

cmd/loom/sendtx.go
```diff
@@ -50,16 +55,30 @@ func newMigrationCommand() *cobra.Command {
 			if callerChainID == "" {
 				callerChainID = cli.TxFlags.ChainID
 			}
-			return migrationTx(Id, cli.TxFlags.PrivFile, cli.TxFlags.Algo, callerChainID)
+			if Id == 2 {
```
ParthDesai: Shift this thing into another method. And use `switch`, so we can add parameter for other migrations too.

cmd/loom/sendtx.go
suhasagg: Done

cmd/loom/sendtx.go
suhasagg: Done.

cmd/loom/sendtx.go
```diff
@@ -50,16 +53,48 @@ func newMigrationCommand() *cobra.Command {
...
 	setChainFlags(migrationCmd.Flags())
 	return migrationCmd
 }
 
-func migrationTx(migrationId uint32, privFile, algo, callerChainID string) error {
+//Parse Migration Input Parameters depending upon MigrationId
+func parseInputParameters(migrationId uint32, inputFile string) ([]byte, error) {
+	var inputBytes []byte
+	switch migrationId {
+	case 1:
```
ParthDesai: Should be: 

```
case 2:
default:
```

No advantage of having case 1 separately.

suhasagg: Done.

cmd/loom/sendtx.go
```diff
@@ -50,16 +53,46 @@ func newMigrationCommand() *cobra.Command {
...
+		err = jsonpb.UnmarshalString(string(jsonData), &policy)
+		if err != nil {
+			return nil, err
+		}
+		inputBytes, err = proto.Marshal(&policy)
+		if err != nil {
+			return nil, err
+		}
+	default:
+		inputBytes = nil
```
ParthDesai: Default value of `inputBytes` is nil. No need to set it explicitly.

suhasagg: Done

suhasagg: Done.

app.go
```diff
@@ -256,12 +256,18 @@ type ChainConfigManager interface {
 	EnableFeatures(blockHeight int64) error
 }
 
+type CoinPolicyManager interface {
+	MintCoins() error
```
enlight: `MintCoins` -> `ApplyPolicy`

builtin/plugins/coin/coin.go
```diff
@@ -42,16 +43,18 @@ type (
...
-
-	BurnRequest = ctypes.BurnRequest
+	Policy               = ctypes.Policy
+	BurnRequest          = ctypes.BurnRequest
 )
 
 var (
-	ErrSenderBalanceTooLow = errors.New("sender balance is too low")
+	ErrSenderBalanceTooLow   = errors.New("sender balance is too low")
+	InvalidBaseMintingAmount = errors.New("Base Minting Amount should be greater than zero")
```
enlight: Go naming convention for errors is `ErrBlahBlah`

builtin/plugins/coin/coin.go
```diff
@@ -76,7 +79,37 @@ func (c *Coin) Meta() (plugin.Meta, error) {
 func (c *Coin) Init(ctx contract.Context, req *InitRequest) error {
 	div := loom.NewBigUIntFromInt(10)
 	div.Exp(div, loom.NewBigUIntFromInt(18), nil)
-
+	//Checks if Coin Policy Feature is enabled before loading Coin Monetary Supply Policy
```
enlight: This comment is wrong, get rid of it

builtin/plugins/coin/coin.go
```diff
@@ -76,7 +79,37 @@ func (c *Coin) Meta() (plugin.Meta, error) {
...
+		}
+		if req.Policy.MintingAccount == nil {
+			return errors.New("Invalid Minting Account Address")
+		}
+		addr := loom.UnmarshalAddressPB(req.Policy.MintingAccount)
+		if addr.Compare(loom.RootAddress(addr.ChainID)) == 0 {
+			return errors.New("Minting Account Address cannot be Root Address")
+		}
+		deflationFactorNumerator := req.Policy.DeflationFactorNumerator
+		deflationFactorDenominator := req.Policy.DeflationFactorDenominator
```
enlight: What's the point of these temporary vars?

builtin/plugins/coin/coin.go
```diff
@@ -76,7 +79,37 @@ func (c *Coin) Meta() (plugin.Meta, error) {
...
+		policy := &Policy{
+			DeflationFactorNumerator:   deflationFactorNumerator,
+			DeflationFactorDenominator: deflationFactorDenominator,
+			BaseMintingAmount:          req.Policy.BaseMintingAmount,
+			MintingAccount:             req.Policy.MintingAccount,
+		}
+		err := ctx.Set(policyKey, policy)
+		if err != nil {
```
enlight:
```golang
if err := ctx.Set(policyKey, policy); err != nil {
    return err
}
```

cmd/loom/sendtx.go
```diff
@@ -32,6 +32,8 @@ type deployTxFlags struct {
 	Name       string `json:"name"`
 }
 
+type Policy = ctypes.Policy
```
enlight: Don't think it's a good idea to re-export a coin type from a module that has nothing to do with it

migrations/coin_policy_migration.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/loomnetwork/go-loom/plugin/contractpb"
+	"github.com/loomnetwork/loomchain"
+)
+
+var (
+	policyKey = []byte("policy")
+)
+
+type (
+	Policy = ctypes.Policy
```
enlight: Just use `ctypes.Policy`

plugin/coin_policy_manager.go
```diff
@@ -0,0 +1,47 @@
...
+	contract "github.com/loomnetwork/go-loom/plugin/contractpb"
+	"github.com/loomnetwork/loomchain"
+	"github.com/loomnetwork/loomchain/builtin/plugins/coin"
+	regcommon "github.com/loomnetwork/loomchain/registry"
+	"github.com/pkg/errors"
+)
+
+var (
+	// ErrCoinContractNotFound indicates that the Coin contract hasn't been deployed yet.
+	ErrCoinContractNotFound = errors.New("[CoinDeflationManager] CoinContract contract not found")
```
enlight: Description needs to be improved

plugin/coin_policy_manager.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+// CoinPolicyManager implements loomchain.CoinPolicyManager interface
+type CoinPolicyManager struct {
+	ctx   contract.Context
+	state loomchain.State
+}
+
+type Policy = ctypes.Policy
```
enlight: Just use `ctypes.Policy`

app.go
suhasagg: Done.

builtin/plugins/coin/coin.go
suhasagg: Done.

builtin/plugins/coin/coin.go
suhasagg: Done

builtin/plugins/coin/coin.go
suhasagg: Done.

builtin/plugins/coin/coin.go
suhasagg: Done.

cmd/loom/sendtx.go
suhasagg: It is being used at the time of unmarshalling inputParams for migrationId 2.

plugin/coin_policy_manager.go
suhasagg: Done.

plugin/coin_policy_manager.go
suhasagg: Done.

cmd/loom/sendtx.go
ParthDesai: You can just use `ctypes.Policy` directly here.

suhasagg: That's what I have done :)

builtin/plugins/coin/coin.go
```diff
@@ -79,7 +79,7 @@ func (c *Coin) Meta() (plugin.Meta, error) {
 func (c *Coin) Init(ctx contract.Context, req *InitRequest) error {
 	div := loom.NewBigUIntFromInt(10)
 	div.Exp(div, loom.NewBigUIntFromInt(18), nil)
-	//Checks if Coin Policy Feature is enabled before loading Coin Monetary Supply Policy
+	//Checks if Coin Policy is not nil before loading Coin Monetary Supply Policy
```
enlight: This comment is completely useless, remove it

plugin/coin_policy_manager.go
```diff
@@ -41,7 +38,7 @@ func NewCoinPolicyManager(pvm *PluginVM, state loomchain.State) (*CoinPolicyMana
 	}, nil
 }
 
-//MintCoins method of coin_deflation_Manager will be called from Block
-func (c *CoinPolicyManager) MintCoins() error {
-	  return coin.Mint(c.ctx)
+//ApplyPolicy method of coin_deflation_Manager will be called from Block
```
enlight: -> `// ApplyPolicy will apply the current economic policy within the coin contract`

builtin/plugins/coin/coin.go
suhasagg: Done.

plugin/coin_policy_manager.go
suhasagg: Done.

builtin/plugins/coin/coin.go
```diff
@@ -213,6 +242,27 @@ func mint(ctx contract.Context, to loom.Address, amount *loom.BigUInt) error {
...
+	policyNumerator := loom.NewBigUIntFromInt(int64(policy.DeflationFactorNumerator))
+	policyDenominator := loom.NewBigUIntFromInt(int64(policy.DeflationFactorDenominator))
+	blockHeight := loom.NewBigUIntFromInt(ctx.Block().Height)
+	depreciation = policyNumerator.Mul(policyNumerator, blockHeight)
+	depreciation.Div(depreciation, policyDenominator)
+	amount := loom.NewBigUIntFromInt(int64(policy.BaseMintingAmount))
+	if amount.Cmp(depreciation) <= 0 {
+		return nil // No more coins to be minted on block creation
+	}
+	amount.Sub(amount, depreciation)
```
enlight: So the original issues https://github.com/loomnetwork/loomchain/issues/1182 says:
> someone could configure their chain to have 10% inflation first year, 5% second. Etc . Ideally they just configure the windows of time for rebalance.

This code doesn't do that.
Also, here the depreciation starts from the first block, which doesn't seem quite right, it should start applying once this feature is enabled (from the height at which it's enabled).

builtin/plugins/coin/coin.go
```diff
@@ -213,6 +257,95 @@ func mint(ctx contract.Context, to loom.Address, amount *loom.BigUInt) error {
...
+				Value: *amount,
+			})
+			if err != nil {
+				return err
+			}
+		}
+	} else {
+		if modulus.Cmp(big.NewInt(1)) == 0 || err == contract.ErrNotFound {
+			//Operator Support to support different inflation ratio patterns for different year
+			if strings.EqualFold("div", operator) {
```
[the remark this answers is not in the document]
suhasagg: There are two operators - div and exp -->
```
			if strings.EqualFold("div", operator) {
				changeRatioDenominator = changeRatioDenominator.Mul(changeRatioDenominator, year)
				basePercentage = basePercentage.Mul(basePercentage, changeRatioNumerator)
			} else {
				changeRatioNumerator = changeRatioNumerator.Exp(changeRatioNumerator, year, nil)
				changeRatioDenominator = changeRatioDenominator.Exp(changeRatioDenominator, year, nil)
				basePercentage = basePercentage.Mul(basePercentage, changeRatioNumerator)
			}
```

builtin/plugins/coin/coin.go
```diff
@@ -213,6 +257,95 @@ func mint(ctx contract.Context, to loom.Address, amount *loom.BigUInt) error {
...
+	var policy Policy
+	var amount *common.BigUInt
+	err := ctx.Get(policyKey, &policy)
+	if err != nil {
+		return errUtil.Wrap(err, "Failed to Get PolicyInfo")
+	}
+	changeRatioNumerator := loom.NewBigUIntFromInt(int64(policy.ChangeRatioNumerator))
+	changeRatioDenominator := loom.NewBigUIntFromInt(int64(policy.ChangeRatioDenominator))
+	blockHeight := loom.NewBigUIntFromInt(ctx.Block().Height)
+	totalSupply := loom.NewBigUIntFromInt(int64(policy.TotalSupply))
```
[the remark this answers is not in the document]
suhasagg: total Supply means --> Supply of coins in first year.
This is used to determine --> minting amount per block for first year as obtained from formula - Total Supply for first year/Blocks produced in a year.
This is used if minting is to start from first block/at a certain blockheight of chain for first year.

builtin/plugins/coin/coin.go
```diff
@@ -213,6 +257,95 @@ func mint(ctx contract.Context, to loom.Address, amount *loom.BigUInt) error {
...
+			if err != nil {
+				return err
+			}
+		}
+	} else {
+		if modulus.Cmp(big.NewInt(1)) == 0 || err == contract.ErrNotFound {
+			//Operator Support to support different inflation ratio patterns for different year
+			if strings.EqualFold("div", operator) {
+				changeRatioDenominator = changeRatioDenominator.Mul(changeRatioDenominator, year)
+				basePercentage = basePercentage.Mul(basePercentage, changeRatioNumerator)
```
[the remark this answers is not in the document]
suhasagg:
```
   totalSupply = &econ.TotalSupply.Value
   inflationforYear := totalSupply.Mul(totalSupply, basePercentage)
   inflationforYear = inflationforYear.Div(inflationforYear, changeRatioDenominator)
   inflationforYear = inflationforYear.Div(inflationforYear, loom.NewBigUIntFromInt(100))
   amount = inflationforYear.Div(inflationforYear, blocksGeneratedPerYear)
   //Minting Amount per block computed starting from beginning block of that year and saved to avoid re-computations
   err = ctx.Set(mintingAmountKey, &types.BigUInt{
	Value: *amount,
   })
   if err != nil {
      return err
  }
  ```
  This is accomplished by this code, division is done in the end to avoid rounding off loss

File: builtin/plugins/coin/coin.go [hunk 32]
```diff
@@ -213,6 +257,95 @@ func mint(ctx contract.Context, to loom.Address, amount *loom.BigUInt) error {
...
+			} else {
+				changeRatioNumerator = changeRatioNumerator.Exp(changeRatioNumerator, year, nil)
```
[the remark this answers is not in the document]
suhasagg: This else is for exponential operator

builtin/plugins/coin/coin.go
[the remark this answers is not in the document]
ParthDesai: then have it as like this: ```
if string.EqualFold("div", operator) {
} else if string.EqualFold("exp", operator) {
} else {
   // return an error
}
```

File: builtin/plugins/coin/coin.go (hunk 30 above)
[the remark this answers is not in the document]
Comment (ParthDesai) [reply builtin/plugins/coin/coin.go]: @suhasagg why it is coming from policy? 
FYI, Total supply of coin means, coins currently in circulation.

File: builtin/plugins/coin/coin.go (hunk 31 above)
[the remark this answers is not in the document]
Comment (ParthDesai) [reply builtin/plugins/coin/coin.go]: Change your terminology. TotalSupply means something else here.

File: builtin/plugins/coin/coin.go (hunk 29 above)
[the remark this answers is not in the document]
Comment (suhasagg) [reply builtin/plugins/coin/coin.go]: I am performing input validation in operator.
It checks for div and exp explicitly.
Unless, we modify operator via CLI, to change inflation operator stored in policy, it should be fine as we will never encounter a different operator

File: builtin/plugins/coin/coin.go (hunk 30 above)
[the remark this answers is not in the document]
Comment (suhasagg) [reply builtin/plugins/coin/coin.go]: Ok, I am changing Total Supply to Inflation for first year, but for year 1, aren't two equal.
i.e Inflation  = Total Supply, as chain was initiated in that year.

File: builtin/plugins/coin/coin.go (hunk 31 above)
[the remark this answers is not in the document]
Comment (suhasagg) [reply builtin/plugins/coin/coin.go]: Should I rename Total Supply to BaseAmount as inflation is applied on top of BaseAmount.

File: builtin/plugins/coin/coin.go (hunk 29 above)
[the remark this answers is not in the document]
Comment (ParthDesai) [reply builtin/plugins/coin/coin.go]: it's good that you added input validation, but it is required to add that validation here too.

File: builtin/plugins/coin/coin.go (hunk 31 above)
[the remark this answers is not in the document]
Comment (ParthDesai) [reply builtin/plugins/coin/coin.go]: yeah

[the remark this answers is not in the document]
Comment (suhasagg) [reply builtin/plugins/coin/coin.go]: Done.

File: builtin/plugins/coin/coin.go [hunk 33]
```diff
@@ -213,6 +257,95 @@ func mint(ctx contract.Context, to loom.Address, amount *loom.BigUInt) error {
...
+			//Minting Amount per block computed starting from beginning block of that year and saved to avoid re-computations
+			err = ctx.Set(mintingAmountKey, &types.BigUInt{
+				Value: *amount,
+			})
+			if err != nil {
+				return err
+			}
+		}
+	}
+	var amount1 = &types.BigUInt{
```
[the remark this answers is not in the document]
suhasagg: Done.

builtin/plugins/coin/coin.go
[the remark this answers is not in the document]
suhasagg: Validation is added.

builtin/plugins/coin/coin.go
```diff
@@ -213,6 +257,97 @@ func mint(ctx contract.Context, to loom.Address, amount *loom.BigUInt) error {
...
+			}
+		}
+	} else {
+		if modulus.Cmp(big.NewInt(1)) == 0 || err == contract.ErrNotFound {
+			//Operator Support to support different inflation ratio patterns for different year
+			if strings.EqualFold("div", operator) {
+				changeRatioDenominator = changeRatioDenominator.Mul(changeRatioDenominator, year)
+				basePercentage = basePercentage.Mul(basePercentage, changeRatioNumerator)
+			} else if strings.EqualFold("exp", operator) {
+				changeRatioNumerator = changeRatioNumerator.Exp(changeRatioNumerator, year, nil)
```
ParthDesai: In case of `exp` we dont need to change numerator only denominator

ParthDesai: Also, change naming convention. basePercentage is what we read from config, and then depending upon which year it is, we update it

ParthDesai: Check if we try to `div` by zero.

builtin/plugins/coin/coin.go
```diff
@@ -213,6 +257,97 @@ func mint(ctx contract.Context, to loom.Address, amount *loom.BigUInt) error {
...
+			}
+			//Fetches Total Supply from coin contract and applies inflation ratio on total Supply- Happens at
+			//beginning block for that year or block height at which minting is enabled
+			baseAmount = &econ.TotalSupply.Value
+			inflationforYear := baseAmount.Mul(baseAmount, basePercentage)
+			inflationforYear = inflationforYear.Div(inflationforYear, changeRatioDenominator)
+			inflationforYear = inflationforYear.Div(inflationforYear, loom.NewBigUIntFromInt(100))
+			amount = inflationforYear.Div(inflationforYear, blocksGeneratedPerYear)
+			//Minting Amount per block computed starting from beginning block of that year and saved to avoid re-computations
+			err = ctx.Set(mintingAmountKey, &types.BigUInt{
```
ParthDesai: Only mint if amount is greater than zero.

ParthDesai: Also, why are you setting it and then getting it? It's redundant. Just mint and store it.

ParthDesai:
Also, please make sure builds are passing.

builtin/plugins/coin/coin.go
suhasagg: Suppose base percentage is - 10
Change ratio of base percentage is - 2/3 
For consecutive years --> Base percentage will be as follows - 
1) Base percentage - 10 * 2/3 - ~6.6
2) Base percentage - 10 * 4/9 - ~4.4
So, you need to take exponential of both numerator and denominator

suhasagg: I am doing input validation of denominator that it is not equal to zero, so practically denominator will never be zero.
But will add a check, here as well.

builtin/plugins/coin/coin.go
suhasagg: I am already doing this validation here, when minting Amount becomes zero, minting stops and function returns nil  
//Boundary case when minting amount per block becomes zero
	if amount == loom.NewBigUIntFromInt(0) {
		return nil // No more coins to be minted on block creation
	}

suhasagg: I am setting base minting amount in context because of following reason - 
BaseMintingAmount remains constant for a year and changes on yearly basis.
So, I am setting it in contract context so that it does not need to be recomputed.
I am storing in contract context so that storage is permanent, in case chain crashes for a short period in a year, base minting amount is already stored. Though it can be recomputed but why will we want to do same computation again :)

suhasagg:
Please take a look at unit tests, I have covered all the scenarios and boundary cases there.
Suppose we start at a particular blockheight in a year, when base minting Amount becomes zero and more.
Test also verifies that minting amount will become zero after a long period.

builtin/plugins/coin/coin.go
ParthDesai: I am not saying you shouldnt store. I am asking why you are setting it and getting it in same function call.

ParthDesai: Also, where is the code, that check in code that prevents recomputing of minting amount

suhasagg:
```
_, modulus := big.NewInt(ctx.Block().Height).DivMod(big.NewInt(ctx.Block().Height), big.NewInt(int64(policy.BlocksGeneratedPerYear)), big.NewInt(int64(policy.BlocksGeneratedPerYear)))
```
 This computes modulus which determines whether minting is being done at the beginning of year or at a certain blockheight.
 
//Checks if minting amount is computed before or it is being computed for first time
```
	err = ctx.Get(mintingAmountKey, checkMintAmount)
	if err != nil && err != contract.ErrNotFound {
		return err
	}	
```	
If minting is being done for the first time ==> error is contract.ErrNotFound.

```
if modulus.Cmp(big.NewInt(1)) == 0 || err == contract.ErrNotFound 	
```
If modulus is 1, this determines it is beginning of year and minting amount will need to be recomputed or if err == contract.ErrNotFound, minting amount will need to be computed as minting is being done for the first time whether at beginning of year or at a certain blockheight at a certain point in a year.	

Computation code does not run otherwise and minting amount once computed is stored in contract context from where it is derived for the rest of the year and recomputed as described above.

builtin/plugins/coin/coin.go
```diff
@@ -213,6 +257,97 @@ func mint(ctx contract.Context, to loom.Address, amount *loom.BigUInt) error {
...
+	basePercentage := loom.NewBigUIntFromInt(int64(policy.BasePercentage))
+	operator := policy.Operator
+	_, modulus := big.NewInt(ctx.Block().Height).DivMod(big.NewInt(ctx.Block().Height), big.NewInt(int64(policy.BlocksGeneratedPerYear)), big.NewInt(int64(policy.BlocksGeneratedPerYear)))
+	year := blockHeight.Div(blockHeight, blocksGeneratedPerYear)
+	year = year.Add(year, loom.NewBigUIntFromInt(1))
+	var checkMintAmount = &types.BigUInt{
+		Value: *loom.NewBigUIntFromInt(0),
+	}
+	//Checks if minting amount is computed before or it is being computed for first time
+	err = ctx.Get(mintingAmountKey, checkMintAmount)
```
ParthDesai: Where are you using this `checkMintAmount`?  Also, if you find the amount set in context, you are still going to do entire calculation right?

ParthDesai: Also, make this calculation more modular, so it's easy to understand.

suhasagg:
Write up is posted in coin-deflation group

builtin/plugins/coin/coin.go
```diff
@@ -209,10 +254,223 @@ func mint(ctx contract.Context, to loom.Address, amount *loom.BigUInt) error {
...
+			Value: *loom.NewBigUIntFromInt(ctx.Block().Height),
+		})
+		if err != nil {
+			return err
+		}
+	}
+	var mintingHeight = &types.BigUInt{
+		Value: *loom.NewBigUIntFromInt(0),
+	}
+	err1 := ctx.Get(mintingHeightKey, mintingHeight)
```
ParthDesai: Why `err1`?

builtin/plugins/coin/coin.go
```diff
@@ -209,10 +254,223 @@ func mint(ctx contract.Context, to loom.Address, amount *loom.BigUInt) error {
...
+	})
+	if err != nil {
+		return nil, err
+	}
+	return amount, nil
+}
+
+//Computes minting amount per year, after applying change ratio to base percentage using div operator
+//Computes inflation by multiplying base Amount with base percentage. Base percentage is computed by multiplying base percentage with change ratio on which div operator is applied according to the year.
+func ComputeforConsecutiveYearBeginningDivOperator(ctx contract.Context, baseAmount *common.BigUInt, changeRatioNumerator *common.BigUInt,
```
ParthDesai: Write a common function, which than use switch case to fan out to different operator.

builtin/plugins/coin/coin.go
```diff
@@ -209,10 +254,223 @@ func mint(ctx contract.Context, to loom.Address, amount *loom.BigUInt) error {
...
+	err := ctx.Set(mintingAmountKey, &types.BigUInt{
+		Value: *amount,
+	})
+	if err != nil {
+		return nil, err
+	}
+	return amount, nil
+}
+
+//Computes minting Amount for first year if Block Height is greater than blocks generated in a year, whether its year beginning or in the middle
```
ParthDesai: Why would we need to measure if Block height is greater than blocks generated in a year.

ParthDesai: Also, keep in mind that, we will start counting year from the height plugin is enabled.

suhasagg: Algorithm Description for calculate base minting Amount for first year.

Two cases ==>

Case 1
Minting starts when first block of blockchain is produced.
In this case there will not be any total supply.
So, minting Amount per block will be computed as Base Amount for first year/Blocks generated in a year

Case 2
Minting starts at a height which is greater than blocks generated in a year.
If block height is greater than one year, when minting starts.
In this case, there will be some total supply accumulated for running period of blockchain, so minting amount per block will be computed as follows -
Base percentage for first year will be applied on top of total supply to get inflation for first year.
This inflation computed will be divided by Blocks generated in a year to get minting amount per block.

suhasagg: Have taken that into account already, you can see there is a minting height which is computed when minting happens for the first time in blockchain, rest of the computations and year computations are relative to this.
```
    //Sets Minting Height - Height at which minting started
        err := ctx.Set(mintingHeightKey, &types.BigUInt{
            Value: *loom.NewBigUIntFromInt(ctx.Block().Height),
        })
        if err != nil {
            return err
        }
    }
    var mintingHeight = &types.BigUInt{
        Value: *loom.NewBigUIntFromInt(0),
    }
    err1 := ctx.Get(mintingHeightKey, mintingHeight)
    if err1 != nil {
        return err1
    }
   
```

Above is code for the same

```
relativeHeight := big.NewInt(ctx.Block().Height).Sub(big.NewInt(ctx.Block().Height), big.NewInt(mintingHeight.Value.Int64()))
```

```
relativeYear := relativeHeight.Div(relativeHeight, blocksGeneratedPerYear.Int)
```

relativeHeight and relativeYear computation

builtin/plugins/coin/coin.go
suhasagg: I need to track previous error in later part of the code, so need to use this new error variable here.
First err is for mintingAmount, Second err is for mintingHeight

builtin/plugins/coin/coin.go
```diff
@@ -209,10 +254,202 @@ func mint(ctx contract.Context, to loom.Address, amount *loom.BigUInt) error {
...
+//Computes base Amount by taking minting height into Account (Height at which minting started), applies base percentage to compute inflation for that year which is divided by blocks generated per year to get ==> minting Amount per block
+func ComputeforFirstYearBlockHeightgreaterthanOneyear(ctx contract.Context, baseAmount *common.BigUInt, basePercentage *common.BigUInt, blocksGeneratedPerYear *common.BigUInt) (*common.BigUInt, error) {
```
ParthDesai: Change name to something simpler.

suhasagg: Done

builtin/plugins/coin/coin.go
suhasagg: Done.

enlight:
Will revisit this at a later date.

Issue #1190, opened 2019-05-31
JSON RPC eth_getReceiptByTransaction not following the standard


eduardonunesp:
The command [eth_getReceiptByTransaction](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_gettransactionreceipt) which returns the receipt for a given transaction isn't following the standard for fields `contractAddress` and `to`. 

The issues on fields:
 * `to` - the field isn't available on the results for the command `eth_getReceiptByTransaction`, also the field should be `null` when it's a contract creation transaction.
 * `contractAddress` - the field is available, the problem is the field should be with the contract address only if the transaction was a contract creation, otherwise is `null`.

> According to the description above both fields are available on the response even the value is `null`

> There are errors on `blockscout` because it can confirms the type of the input

Issue #1157, opened 2019-05-24
Error on retrieve transaction receipt for go contracts on Asia 1


eduardonunesp:
Found some incongruencies on `asia1` chain, the issues are found by the `indexer` of the `blockscout` on the process of indexing, the cases found are related to issues to retrieve the receipt.

Url tested: `https://test-z-asia1.dappchains.com/eth`

## Case 1: Failed to get a receipt from the transaction 

```bash
ethquery blockbyhash 0x960ef6c24536626553580884cf9d21a109785778daca46077549f7bd275e332a -v
{"id": 0, "jsonrpc":"2.0", "method": "eth_getBlockByHash", "params": ["0x960ef6c24536626553580884cf9d21a109785778daca46077549f7bd275e332a",true]}
{
  "result": {
    "number": "0x2f13e",
    "hash": "0x960ef6c24536626553580884cf9d21a109785778daca46077549f7bd275e332a",
    "parentHash": "0x0f1321bff6fd38fcd770c4b5b2dd5337f2beb8d2c9bc507919ddce68612c128c",
    "nonce": "0x0000000000000000",
    "sha3Uncles": "0x0000000000000000000000000000000000000000000000000000000000000000",
    "logsBloom": "0x0",
    "transactionsRoot": "0x0000000000000000000000000000000000000000000000000000000000000000",
    "stateRoot": "0x0000000000000000000000000000000000000000000000000000000000000000",
    "receiptsRoot": "0x0000000000000000000000000000000000000000000000000000000000000000",
    "miner": "0x81578447947c18989b08cb17f9dc613fcd1169e6",
    "difficulty": "0x0",
    "totalDifficulty": "0x0",
    "extraData": "0x0",
    "size": "0x0",
    "gasLimit": "0x0",
    "gasUsed": "0x0",
    "timestamp": "0x5ce674ab",
    "transactions": [
      {
        "hash": "0x851e00a8dd97a548b8e8650e80eeeb74f6d8211322c659ae8da09315aeb12fdc",
        "nonce": "0x13",
        "blockHash": "0x960ef6c24536626553580884cf9d21a109785778daca46077549f7bd275e332a",
        "blockNumber": "0x2f13e",
        "transactionIndex": "0x0",
        "from": "0x1950b39864e23a4e98489bf00553d0ac53e15469",
        "to": "0x152c4e874a569a61e75b9b30ade7b6a265a3720e",
        "value": "0x0",
        "gasPrice": "0x0",
[...]

PR #1226, opened 2019-06-13
Returns tx hashs for all transactions, returns receipts for all valid hashes.

Sriep:

Addresses https://github.com/loomnetwork/loomchain/issues/1157
Should also provide a fix for https://github.com/loomnetwork/loomchain/issues/1190 `eth_getTransactionReceipt`
Replaces https://github.com/loomnetwork/loomchain/pull/1213

Allows accessing receipts using both Tendermint and loom tx hashes. 
In block, inquiries will display tendermint tx hash if no loom tx hash is available, for example, if the transaction failed.
Will return a receipt if the input is a valid hash (starts `0x`) and the loomchian is ok. If the hash cannot be identified as either a loom-tx-hash or tendermint-tx-hash, an empty receipt with status ~ failed is returned.


app.go
```diff
@@ -550,7 +551,7 @@ func (a *Application) DeliverTx(txBytes []byte) abci.ResponseDeliverTx {
 	r, err := a.processTx(txBytes, false)
 	if err != nil {
 		log.Error(fmt.Sprintf("DeliverTx: %s", err.Error()))
-		return abci.ResponseDeliverTx{Code: 1, Log: err.Error()}
+		return abci.ResponseDeliverTx{Code: 1, Log: err.Error(), Data: ttypes.Tx(txBytes).Hash()}
```
enlight: This is a consensus/replay breaking change, the `Data` field is part of the block hash, so you can't change it in `DeliverTx` without a feature flag, though in theory you should be able to change it in `CheckTx` without breaking anything.

receipts/chain/chain.go
```diff
@@ -61,7 +61,7 @@ func (sr *StateDBReceipts) GetReceipt(state loomchain.ReadOnlyState, txHash []by
 	receiptState := store.PrefixKVReader(common.ReceiptPrefix, state)
 	txReceiptProto := receiptState.Get(txHash)
 	txReceipt := types.EvmTxReceipt{}
-	if txReceiptProto == nil {
+	if txReceiptProto == nil || len(txReceiptProto) == 0 {
```
enlight: `len(txReceiptProto) == 0` is all you need, `len(nil) == 0`

app.go
Sriep: Removed. It's possible to calculate the tx hash in advance with `"github.com/tendermint/tendermint/types".Tx(tx-input-Bytes).Hash()`.

Sriep:
Changes so the transaction does not return the transaction hash on failed transactions. The tendermint [transaction hash](https://godoc.org/github.com/tendermint/tendermint/types#Tx.Hash) can be calculated before the transaction is run. It is Sum256 of input bytes, https://godoc.org/crypto/sha256#Sum256 .

rpc/query_server.go
```diff
@@ -749,36 +756,44 @@ func (s *QueryServer) EthGetTransactionReceipt(hash eth.Data) (resp eth.JsonTxRe
...
-
-		for i := 0; i < len(txReceipt.Logs); i++ {
-			txReceipt.Logs[i].BlockTime = timestamp
-		}
+	height := int64(txReceipt.BlockNumber)
+	blockResult, err := s.BlockStore.GetBlockByHeight(&height)
+	if err != nil {
+		// return empty response if cannot find hash
+		resp.Status = eth.EncInt(int64(StatusTxFail))
+		return resp, nil
```
enlight: Should probably just return the error in this case, the idea is to return a fake receipts for failed txs, if the block doesn't exist it's something else entirely.

rpc/query_server.go
```diff
@@ -749,36 +756,44 @@ func (s *QueryServer) EthGetTransactionReceipt(hash eth.Data) (resp eth.JsonTxRe
...
+	}
+	if int32(len(blockResult.Block.Data.Txs)) <= txReceipt.TransactionIndex {
+		// return empty response if cannot find hash
+		resp.Status = eth.EncInt(int64(StatusTxFail))
+		return resp, nil
```
enlight: Here too it probably makes more sense to return an error rather than a fake receipt.

rpc/query_server.go
```diff
@@ -749,36 +756,44 @@ func (s *QueryServer) EthGetTransactionReceipt(hash eth.Data) (resp eth.JsonTxRe
 	if err != nil {
 		return resp, err
 	}
+
 	txReceipt, err := r.GetReceipt(snapshot, txHash)
 	if err != nil && errors.Cause(err) != common.ErrTxReceiptNotFound {
-		return resp, err
+		// return empty response if cannot find hash
+		resp.Status = eth.EncInt(int64(StatusTxFail))
+		return resp, nil
```
enlight: Looks like `LevelDbReceipts.GetReceipt` won't ever return `common.ErrTxReceiptNotFound`, it'll return `ErrNotFound` instead, so it doesn't looks like `getReceiptByTendermintHash` will ever get called below.

Sriep: I should look for a way of making less confusing, but the code holds together.
 `r.GetReceipt(snapshot, txHash)` looks for the hash both in the leveldb and the chain.
```go
	case ReceiptHandlerLevelDb:
		receipt, err := r.leveldbReceipts.GetReceipt(txHash)
		// In case receipts has been upgraded from V1 to V2 try getting the receipt from the chain.
		if err != nil {
			var chainErr error
			receipt, chainErr = r.chainReceipts.GetReceipt(state, txHash)
			if chainErr != nil {
				return receipt, errors.Wrap(chainErr, err.Error())
			}
		}
		return receipt, nil
```
`levledbReceipts.GetReceipt` tends to return `leveldb: not found`, while `r.chainReceipts.GetReceipt` returns `common.ErrTxReceiptNotFound`. Its the `r.chainReceipts.GetReceipt` error that ends up being the cause.

Sriep: Changed it so it's more explicit that GetReceipt returns `common.ErrTxReceiptNotFound`.

Issue #1177, opened 2019-05-30
eth_sendRawTransaction modified to handle Ethereum transactions 


Sriep:
`eth_sendRawTransaction`, expects the input to be a TM tx, which isn't very useful because Web3 compatible clients don't send TM txs. We want that endpoint to accept Eth txs from any Web3 compatible client.

Sriep:
Fixed by https://github.com/loomnetwork/loomchain/pull/1227 and https://github.com/loomnetwork/go-loom/pull/391

enlight:
Resolved

PR #1227, opened 2019-06-13
eth_sendRawTransaction modified to support ethereum transactions

Sriep:

Fix for https://github.com/loomnetwork/loomchain/issues/1177
Changes to support  [Ethereum transactions](https://github.com/loomnetwork/go-ethereum/blob/loomchain/core/types/transaction.go#L38)
Requires https://github.com/loomnetwork/go-loom/pull/391
Changes to allow access to `auth.Config` information to determine Chain id. https://github.com/loomnetwork/loomchain/issues/1348
Add new feature flag `EthTxFeature` to activate ethereum transaction functionality.


rpc/tendermint_handler.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+func ethereumToTendermintTx(txBytes []byte) (ttypes.Tx, error) {
+	ethTx := &vm.EthTx{
+		EthereumTransaction: txBytes,
+	}
+
+	var err error
+	msg := &vm.MessageTx{}
+	msg.Data, err = proto.Marshal(ethTx)
```
enlight: Don't need this, can just do `msg.Data = txBytes`

Sriep:
Creating a new Tx id was so that `vm.handler.go.ProcessTx` knew how to unwrap the MessageTx.Data object. Is it an `DeployTx`, `CallTx` or `rpl-wrapped-Ethereum-Transaction object`
Alternatives to a new Tx id for an ethereum-transaction are.
1) Somewhere, maybe `VerifySolidity66Byte` the transaction self-modifies itself so that messageTx.Data field is converted into a DeplyTx or CallTx, losing the singing information just keeping the Code/Input data.
2) In `vm.handler.go.ProcessTx` if unmarshalling messageTx.Data as a DeplyTx or CallTx fails with an error, we have a stab as rlp.Decoding it. Only returning en error if both attempts fail. I'll code this for the moment.

rpc/tendermint_handler.go
```diff
@@ -0,0 +1,152 @@
...
+	if err := rlp.DecodeBytes(txBytes, &tx); err != nil {
+		return nil, err
+	}
+
+	if tx.To() != nil {
+		msg.To = EthToLoomAddress(*tx.To()).MarshalPB()
+	}
+
+	chainConfig := utils.DefaultChainConfig()
+	ethSigner := types.MakeSigner(&chainConfig, blockNumber)
```
enlight: This picks a signer based on the block number (and you chose 5 for whatever reason), which doesn't make much sense since we're not running an Ethereum node, should figure out which concrete signer implementation we should be using and use that.

auth/eth.go
```diff
@@ -24,3 +43,33 @@ func verifyTron(tx SignedTx) ([]byte, error) {
...
+		return nil, err
+	}
+
+	chainConfig := utils.DefaultChainConfig()
+	ethSigner := etypes.MakeSigner(&chainConfig, big.NewInt(1))
+	from, err := etypes.Sender(ethSigner, &tx)
+	if err != nil {
+		return nil, err
+	}
+	return from.Bytes(), err
```
enlight: Also check that `tx.To()` matches `msg.To`, and that `NonceTx.Sequence` matches the `tx` nonce.

auth/eth.go
```diff
@@ -24,3 +43,33 @@ func verifyTron(tx SignedTx) ([]byte, error) {
 	}
 	return tronAddr.Bytes(), nil
 }
+
+func verifyEthereumTransacton(signedTx SignedTx) ([]byte, error) {
```
enlight: -> `verifyEthTx`

eduardonunesp:
https://github.com/loomnetwork/loomchain/issues/1348

auth/eth.go
```diff
@@ -3,12 +3,31 @@
...
 
-func verifySolidity66Byte(tx SignedTx) ([]byte, error) {
+const (
+	ethID = uint32(4)
+)
+
+func VerifySolidity66Byte(tx SignedTx) ([]byte, error) {
+	if tx.Signature == nil {
+		return verifyEthereumTransacton(tx)
+	}
```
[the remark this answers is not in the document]
Sriep: I moved this check outside top level middleware function.
```go
		if tx.Id == 4 || signedTx.Signature == nil {
			if !state.FeatureEnabled(loomchain.EthTxFeature, false) {
				return r, errors.New("ethereum transactions feature not enabled")
			}
		}
```

rpc/tendermint_handler.go
Sriep: The block number field is used to see if we are before or after the [`EIP155` fork](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md). 
The block number corresponding to that fork is defined in the ChainConfig, we have set it to [`big.NewInt(0)`](https://github.com/loomnetwork/loomchain/blob/master/evm/evm.go#L263) .
So the blockNumber parameter can be any non-negative integer.
I'll use the value from `chainConfig` to make it more explicit what is going on here.

Sriep:
Fix for issue 1377 & 1371.

rpc/query_server.go
```diff
@@ -1005,7 +1038,11 @@ func (s *QueryServer) EthGetTransactionCount(local eth.Data, block eth.BlockHeig
...
 	}
-	address, err := eth.DecDataToAddress(s.ChainID, local)
+	localBytes, err := eth.DecDataToBytes(local)
+	if err != nil {
+		return eth.Quantity("0x0"), err
+	}
+	address, err := s.localToEthAccount(localBytes)
 	if err != nil {
 		return eth.Quantity("0x0"), err
 	}
```
enlight: Should just call `auth.Nonce(snapshot, address)` here instead of `s.Nonce`

rpc/eth_rpc_funcs.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	if tx.To() != nil {
+		msg.To = loom.Address{
+			ChainID: chainID,
+			Local:   tx.To().Bytes(),
+		}.MarshalPB()
+	}
+
+	ethChainID, err := evmcompat.ToEthereumChainID(chainID)
```
Sriep: As ethChainId only ever depends on `SendRawTransactionPRCFunc.chainID` can we save in  `SendRawTransactionPRCFunc` rather than calculate every time?

enlight: Yeah, you're right, I'll fix that

Sriep: Also ethSigner := etypes.NewEIP155Signer(ethChainID) couple lines below

Issue #620, opened 2019-01-15
Implement a more space efficient IAVL store


enlight:
The current IAVL store uses up too much disk space by saving ever single version to disk. Some space can be reclaimed by deleting old versions from the IAVL store, however this appears to be very slow, as well as unreliable (with some keys going missing after old versions are deleted, even though they are still supposedly referenced by newer versions). So we need a better IAVL store that ideally doesn't need pruning at all,

Requirements:
- New IAVL store should only write to disk every N blocks, and store the intermediary versions in memory.
- New IAVL store must generate the same app hashes as the current IAVL store when replaying blocks.
- New IAVL store must write a new version to disk when processing the first block after startup.
  This is to make sure that if an app hash mismatch occurs shortly after deploying a new
  build we can rollback the app store to a version that can still be correctly processed
  by the new build.

Pulling this off will likely require forking the `iavl` package to create a hybrid MemDB & LevelDB backed IAVL tree.

Sriep:
name | time | disk size | compact | ratio time | ratio size | ratio compact
-- | -- | -- | -- | -- | -- | --
nieve | 42.753808385 | 143878761 | 143437026 | 100 | 100 | 100
prune2 | 29.573169054 | 30786828 | 17272351 | 69 | 21 | 12
2xDB | 19.184488022 | 65335520 | 65475391 | 44 | 45 | 45
nieve | 48.419747945 | 143365210 | 142958643 | 100 | 100 | 100
prune2 | 37.845223968 | 23948230 | 17220831 | 78 | 16 | 12
Periodic | 38.798430762 | 36314539 | 18650820 | 80 | 25 | 13
CacheDepend | 35.99396027 | 29455756 | 19671531 | 74 | 20 | 13

Sriep:
The 2xDB method forgets orphans along with the in-memory DB, rather than deletes them.  This process is much quicker, almost half the speed of other methods, but the disk datbase is 3-4 times the size; Suggesting that some orphans cannot be just forgotten but need to be explicitly deleted.
There is some suggestion that the CacheDependant method can give better uncompacted DB-size figures. It aims to reduce writes to disk by keeping as much work as available RAM allows in memory. 

First two methods available in all branches.
nieve - saves to disk every block. Keeps all versions.
prune - saves to disk every block. Deletes all but latest two versions.
prune acts as a control and allows the tests across the two branches to be better compared. Need to add seed to random fuzz-tests.
From `iavl-hack` branch.
[...]

PR #1229, opened 2019-06-13
new iavl store

Sriep:

Addresses https://github.com/loomnetwork/loomchain/issues/620
Needs https://github.com/loomnetwork/iavl/pull/4


Sriep:
## IAVLStore
Now two sorts of IAVL store focused on how database changes are saved to memory.  This is controlled by the type of Batch file used.
* Old version saves each database action, individually to disk.
* [New version,](https://github.com/loomnetwork/loomchain/blob/iavl-620/store/differencedb.go#L7) save only the differences since the last save, to disk.

## Parameters
[Parameters](https://github.com/loomnetwork/loomchain/blob/iavl-620/store/iavlstore.go#L231) for creating an IAVL store.
* maxVersions Every block the MutableTree.DeleteVersion method is used to restrict the number of versions maintained in the app.db to the last `maxVersions`; This can be modified by versionFrequency.
* saveFrequecny For most blocks we version the database in memory. However once every `saveFrequency` blocks we save the iavl-tree to the database on disk.
* versionFrequency Every `versionFrequency` block is protected from being deleted as a result of a non-zero `maxVersions`.  So maxVersions=2 and versionFrequency=20 will result in the app.db containing every 20th version plus the latest two versions.
* minCacheSize This is the size of the cache. If `maxCacheSize > minCacheSize` this is the size the cache is reset to after each save.
* maxCacheSize If `maxCacheSize <= minCacheSize`  this is ignored. If `maxCacheSize > minCacheSize`,  when iavl-saveVersions is called if the cahce size exceeds `maxCacheSize` then the tree is saved to disk and the cache reduced in size to minCacheSize. Otherwise, the iavl-tree is versioned in memory.  This setting takes priority over `saveFrequecny`.

Sriep:
Wants https://github.com/loomnetwork/go-loom/pull/397 for unit tests.

PR #1230, opened 2019-06-14
Tests illustrating DPOSv2 bugs

gakonst:
This PR adds 2 tests illustrating the bugs found on DposV2 -- these are not present on the currently runnning v3.

## Migration: 
On line: https://github.com/loomnetwork/loomchain/blob/e70bf8c95bd262dc0f2f838682756c2a1e184696/builtin/plugins/dposv2/dpos.go#L2364, the function creates indices for storing index counter with a unique key combined of validator and delegator addresses. If the validator of delegation is plasma validator, it will set validator to . This causes the issue. If a user has multiple plasma-x delegations, he or she will end up having multiple plasma-0 delegations with the same index 1 during the migration and when the migration is done, some delegations will get lost.

## Redelegate
Suppose a user has two delegations on validator A and validator B, and he redelegates less than full amount from validator B to validator A, the delegation amount on validator A will get overwritten by the new delegation amount.

Both were found by @pathornteng

### DIFF
--- a/builtin/plugins/dposv2/dpos_test.go
+++ b/builtin/plugins/dposv2/dpos_test.go
@@ -2438,12 +2438,12 @@ func TestDedoublingDelegation(t *testing.T) {
 
 // after we migrate we want to have all delegations that were in plasma-* nodes, on plasma-0
 func TestPlasmaDelegationMigration(t *testing.T) {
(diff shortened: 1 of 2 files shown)


enlight:
Don't think there's much point in merging this in, will just keep the branch around for a bit longer.

PR #1245, opened 2019-06-19
Split Hsm config out of loom.yml

atchapcyp:
Added function and test ParseConfigFileName for specify configuration file.

### DIFF
--- a/cmd/loom/common/config.go
+++ b/cmd/loom/common/config.go
@@ -1,6 +1,7 @@
 package common
 
 import (
+	"log"
 	"path/filepath"
 
 	"github.com/loomnetwork/loomchain/config"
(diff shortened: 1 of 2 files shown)


pathornteng:
Currently, we have duplicate `ParseConfig` functions, one in `cmd/loom/common` and another one in `config/config`. We should remove the one in `config/config` and then change this line https://github.com/loomnetwork/loomchain/blob/fe27e028672fe46a779d4dfa2c85979465d305a0/rpc/query_server.go#L158 to call `ParseConfig` in `cmd/loom/common`

enlight:
> 
> 
> Currently, we have duplicate `ParseConfig` functions, one in `cmd/loom/common` and another one in `config/config`. We should remove the one in `config/config` and then change this line
> 
> https://github.com/loomnetwork/loomchain/blob/fe27e028672fe46a779d4dfa2c85979465d305a0/rpc/query_server.go#L158
> to call `ParseConfig` in `cmd/loom/common`

You got that backwards, `config/config` should have the canonical `ParseConfig`, nothing should depend on `cmd/loom/common` outside of `cmd/loom/*`

config/config.go
```diff
@@ -306,7 +306,10 @@ func ParseConfig() (*Config, error) {
 	if err != nil {
 		return nil, err
 	}
-
+	conf.HsmConfig, err = ParseHSMConfig()
+	if err != nil {
+		return nil, err
+	}
```
enlight: If the hsm config file doesn't exist we should use the HSM config from loom.yml, so this should look something like:
```golang
hsmCfg, err = ParseHSMConfig()
if err == nil {
  conf.HsmConfig = hsmCfg
} else if _, notFound := err.(viper.ConfigFileNotFoundError); !notFound {
  // loom_hsm config file exists but couldn't be loaded
  return nil, errors.Wrap(err, "failed to load loom_hsm config")
}
```

config/config.go
```diff
@@ -298,15 +298,17 @@ func ParseConfig() (*Config, error) {
...
-		return nil, err
-	}
+	v.ReadInConfig()
 	conf := DefaultConfig()
 	err := v.Unmarshal(conf)
 	if err != nil {
 		return nil, err
 	}
-
+	//hsmDefault := hsmpv.DefaultConfig()
```
pathornteng: Remove this comment

config/config_test.go
```diff
@@ -35,3 +38,94 @@ func TestConfig(t *testing.T) {
...
+	t.Log(conf.HsmConfig)
+	if err != nil {
+		t.Error(err)
+	}
+	assert.Equal(t, conf, actual)
+	_ = os.Remove(exampleLoom + ".yaml")
+}
+
+func TestHsmFileConfigNotFound(t *testing.T) {
+	// var err error
```
pathornteng: remove this comment

Issue #1241, opened 2019-06-18
Make output of `chain-cfg list-features` cmd more concise


enlight:
Currently the output of `loom chain-cfg list-features -u http://plasma.dappchains.com:80` is overly verbose JSON... makes it difficult to see all the features at a glance. We should change the output of the command to look like this instead:

```
name         | status  | validators | height    | vote % | build
---------------------------------------------------------------------
mw:deploy-wl | ENABLED | 4          | 2251050   | 80     | 1028
```

Features should be displayed in lexicographical order.

enlight:
Resolved

PR #1251, opened 2019-06-20
change format style of list-features command

atchapcyp:
Changing format style of chain-cfg list-features command from JSON to table.
#1241

### DIFF
--- a/cmd/loom/chainconfig/cmd.go
+++ b/cmd/loom/chainconfig/cmd.go
@@ -217,7 +217,7 @@ func ListFeaturesCmd() *cobra.Command {
 				return err
 			}
 
-			type MaxLength struct {
+			type maxLength struct {
 				Name        int
 				Status      int
 				Validators  int
@@ -226,7 +226,7 @@ func ListFeaturesCmd() *cobra.Command {
 				BuildNumber int
 			}
 
-			ml := MaxLength{Name: 4, Status: 6, Validators: 10, Height: 6, Percentage: 6, BuildNumber: 5}
+			ml := maxLength{Name: 4, Status: 6, Validators: 10, Height: 6, Percentage: 6, BuildNumber: 5}
 			for _, value := range resp.Features {
 				if len(value.Name) >= ml.Name {
 					ml.Name = len(value.Name)
--- a/cmd/loom/chainconfig/cmd.go
+++ b/cmd/loom/chainconfig/cmd.go
@@ -226,22 +226,13 @@ func ListFeaturesCmd() *cobra.Command {
 				BuildNumber int
 			}
 
-			ml := maxLength{Name: 4, Status: 6, Validators: 10, Height: 6, Percentage: 6, BuildNumber: 5}
+			ml := maxLength{Name: 4, Status: 7, Validators: 10, Height: 6, Percentage: 6, BuildNumber: 5}
 			for _, value := range resp.Features {
-				if len(value.Name) >= ml.Name {
+				if len(value.Name) > ml.Name {
 					ml.Name = len(value.Name)
 				}
-				if len(value.Status.String()) >= ml.Status {
-					ml.Status = len(value.Status.String())
-				}
-				if uintLength(value.BlockHeight) >= ml.Height {
+				if uintLength(value.BlockHeight) > ml.Height {
 					ml.Height = uintLength(value.BlockHeight)
-				}
-				if uintLength(value.Percentage) >= ml.Percentage {
-					ml.Percentage = uintLength(value.Percentage)
-				}
-				if uintLength(value.BuildNumber) >= ml.BuildNumber {
-					ml.BuildNumber = uintLength(value.BuildNumber)
 				}
 			}
 			fmt.Printf(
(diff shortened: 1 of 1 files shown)


cmd/loom/chainconfig/cmd.go
```diff
@@ -214,11 +216,42 @@ func ListFeaturesCmd() *cobra.Command {
 			if err != nil {
 				return err
 			}
-			out, err := formatJSON(&resp)
-			if err != nil {
-				return err
+
+			type MaxLength struct {
```
pathornteng: rename it to `maxLength` as it is not used externally

cmd/loom/chainconfig/cmd.go
```diff
@@ -214,11 +216,42 @@ func ListFeaturesCmd() *cobra.Command {
...
+				if uintLength(value.Percentage) >= ml.Percentage {
+					ml.Percentage = uintLength(value.Percentage)
+				}
+				if uintLength(value.BuildNumber) >= ml.BuildNumber {
+					ml.BuildNumber = uintLength(value.BuildNumber)
+				}
+			}
+
+			for i, value := range resp.Features {
+				if i == 0 {
```
pathornteng: Move the header out of loop so that we don't have to check `i==0`

cmd/loom/chainconfig/cmd.go
```diff
@@ -214,11 +216,42 @@ func ListFeaturesCmd() *cobra.Command {
...
+					fmt.Printf("%-*s | %-*s | %-*s | %-*s | %-*s | %-*s\n", ml.Name, "name", ml.Status, "status", ml.Validators, "validators", ml.Height, "height", ml.Percentage, "vote %", ml.BuildNumber, "build")
```
pathornteng: One line should not exceed 125 characters. You can do this
```
fmt.Printf(
	"%-*s | %-*s | %-*s | %-*s | %-*s | %-*s\n",
	ml.Name, "name", ml.Status, "status",
	ml.Validators, "validators", ml.Height, "height",
	ml.Percentage, "vote %", ml.BuildNumber, "build",
)
```

cmd/loom/chainconfig/cmd.go
```diff
@@ -214,11 +216,42 @@ func ListFeaturesCmd() *cobra.Command {
...
+					fmt.Printf(strings.Repeat("-", ml.Name+ml.Status+ml.Validators+ml.Height+ml.Percentage+ml.BuildNumber+15) + "\n")
+				}
+				fmt.Printf("%-*s | %-*s | %-*d | %-*d | %-*d | %-*d\n", ml.Name, value.Name, ml.Status, value.Status, ml.Validators, len(value.Validators), ml.Height, value.BlockHeight, ml.Percentage, value.Percentage, ml.BuildNumber, value.BuildNumber)
```
pathornteng: Ditto

cmd/loom/chainconfig/cmd.go
```diff
@@ -214,11 +216,42 @@ func ListFeaturesCmd() *cobra.Command {
...
+				}
+				if uintLength(value.BuildNumber) >= ml.BuildNumber {
+					ml.BuildNumber = uintLength(value.BuildNumber)
+				}
+			}
+
+			for i, value := range resp.Features {
+				if i == 0 {
+					fmt.Printf("%-*s | %-*s | %-*s | %-*s | %-*s | %-*s\n", ml.Name, "name", ml.Status, "status", ml.Validators, "validators", ml.Height, "height", ml.Percentage, "vote %", ml.BuildNumber, "build")
+					fmt.Printf(strings.Repeat("-", ml.Name+ml.Status+ml.Validators+ml.Height+ml.Percentage+ml.BuildNumber+15) + "\n")
```
pathornteng: Ditto

cmd/loom/chainconfig/cmd.go
```diff
@@ -214,11 +216,42 @@ func ListFeaturesCmd() *cobra.Command {
...
+				Percentage  int
+				BuildNumber int
+			}
+
+			ml := MaxLength{Name: 4, Status: 6, Validators: 10, Height: 6, Percentage: 6, BuildNumber: 5}
+			for _, value := range resp.Features {
+				if len(value.Name) >= ml.Name {
+					ml.Name = len(value.Name)
+				}
+				if len(value.Status.String()) >= ml.Status {
```
pathornteng: the condition should be just `>` not `>=`, so that we can save one operation

pathornteng: we don't need to check `Status` as it is alway `PENDING` or`WAITING` or `ENABLED`

cmd/loom/chainconfig/cmd.go
```diff
@@ -214,11 +216,42 @@ func ListFeaturesCmd() *cobra.Command {
...
+					ml.Status = len(value.Status.String())
+				}
+				if uintLength(value.BlockHeight) >= ml.Height {
+					ml.Height = uintLength(value.BlockHeight)
+				}
+				if uintLength(value.Percentage) >= ml.Percentage {
```
pathornteng: Same as `Status`

cmd/loom/chainconfig/cmd.go
```diff
@@ -214,11 +216,42 @@ func ListFeaturesCmd() *cobra.Command {
...
+					ml.Percentage = uintLength(value.Percentage)
+				}
+				if uintLength(value.BuildNumber) >= ml.BuildNumber {
```
pathornteng: BuildNumber is not gonna exceed 10000 soon, so it is safe to set 5. No need to check it.

PR #1252, opened 2019-06-20
Add per contract tx limiter

### LINKED ISSUE #1235: Add per contract tx limiter
We have a simple global tx limiter in https://github.com/loomnetwork/loomchain/blob/master/throttle/tx_limiter_middleware.go that rate limits all accounts and all txs. We need to add another tx limiter that limits how many CallTx(s) can be sent to an EVM contract within a pre-configured block range. The new limiter should tie into the UserDeployerWhitelist contract, each tier specified in that contract should have a corresponding block range & max tx count.

The new rate limiter should only run in CheckTx, similar to the global limiter.

mukuls9971:
- [ ] I assign all copyright to Loom Network for the code in the pull request]\
resolve #1235


throttle/contract_tx_limiter_middleware.go
```diff
@@ -0,0 +1,124 @@
...
+			return res, errors.Wrap(err, "throttle: unwrap nonce Tx")
+		}
+
+		var tx loomchain.Transaction
+		if err := proto.Unmarshal(nonceTx.Inner, &tx); err != nil {
+			return res, errors.New("throttle: unmarshal tx")
+		}
+
+		// Should not be initialized at every checkTx
+		if TxLimiter == nil {
```
ParthDesai: Why do we have to initialize this here. We could initialize this when we are initializing whole thing

throttle/contract_tx_limiter_middleware.go
```diff
@@ -0,0 +1,124 @@
...
+			var msg vm.MessageTx
+			if err := proto.Unmarshal(tx.Data, &msg); err != nil {
+				return res, errors.Wrapf(err, "unmarshal message tx %v", tx.Data)
+			}
+			var tx vm.CallTx
+			if err := proto.Unmarshal(msg.Data, &tx); err != nil {
+				return res, errors.Wrapf(err, "unmarshal call tx %v", msg.Data)
+			}
+			if tx.VmType == vm.VMType_EVM {
+
```
ParthDesai: Remove this space.

throttle/contract_tx_limiter_middleware.go
```diff
@@ -0,0 +1,124 @@
...
+func (txl *contractTxLimiter) isAccountLimitReached(contractAddr loom.Address, curBlockHeight int64) bool {
+	tier, ok := txl.contractToTierMap[contractAddr.String()]
+	if !ok {
+		return false
+	}
+	blockTxns, ok := txl.contractToBlockTrx[contractAddr.String()]
+	if !ok {
+		return false
+	}
+	minBlockHeight := curBlockHeight - int64(tier.BlockRange) + 1
```
ParthDesai: handle edge cases here, like `curBlockHeight` is less than `BlockRange`

ParthDesai: Also, I feel this algo can be written in more concise format. will get back to you with exact idea.

throttle/contract_tx_limiter_middleware.go
mukuls9971: we need context for initialization, so we needed state.. where should I initialize ?

throttle/contract_tx_limiter_middleware.go
mukuls9971: done

throttle/contract_tx_limiter_middleware.go
mukuls9971: even if curBlockHeight is less than BlockRange, still there wont be any issue ... it is int64 ..

throttle/contract_tx_limiter_middleware.go
ParthDesai: you need to dynamically load contract to tier map.

throttle/contract_tx_limiter_middleware.go
```diff
@@ -0,0 +1,111 @@
...
+			if err != nil {
+				return res, errors.Wrap(err, "throttle: context creation")
+			}
+			contractToTierMap, err := udw.GetContractTierMapping(ctx)
+			if err != nil {
+				return res, errors.Wrap(err, "throttle: contractToTierMap creation")
+			}
+			TxLimiter = &contractTxLimiter{contractToTierMap: contractToTierMap}
+		}
+		if tx.Id == callId {
```
ParthDesai: bail out early in case `tx.Id != callId`

mukuls9971: it is already the end of program where next is being called... do it require to mention this case explicitly.

ParthDesai: I am saying if `tx.Id != callID` then just call `return`

mukuls9971: done

builtin/plugins/user_deployer_whitelist/user_deployer_whitelist.go
```diff
@@ -326,9 +329,11 @@ func (uw *UserDeployerWhitelist) SetTierInfo(ctx contract.Context, req *SetTierI
 		return ErrNotAuthorized
 	}
 	tier := &Tier{
-		TierID: req.TierID,
-		Fee:    req.Fee,
-		Name:   req.Name,
+		TierID:     req.TierID,
```
ParthDesai: Validate blockrange and maxtx value. Both should be greater than 0.

throttle/contract_tx_limiter_middleware.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+}
+
+// NewContractTxLimiterMiddleware add another tx limiter that limits how many CallTx(s) can be sent to an EVM contract within a pre-configured block range
+func NewContractTxLimiterMiddleware(
+	createUserDeployerWhitelistCtx func(state loomchain.State) (contractpb.Context, error),
+) loomchain.TxMiddlewareFunc {
+	TxLimiter = &contractTxLimiter{
+		contractToBlockTrx: make(map[string]map[int64]int64, 0),
+		lastUpdated:        time.Now().Unix() - refreshInterval, // updation time is fifteen minutes
```
ParthDesai: This is not hardcoded to fifteen minutes. Take this as config parameter.

throttle/contract_tx_limiter_middleware.go
```diff
@@ -0,0 +1,123 @@
...
+		}
+		var tx loomchain.Transaction
+		if err := proto.Unmarshal(nonceTx.Inner, &tx); err != nil {
+			return res, errors.New("throttle: unmarshal tx")
+		}
+		if tx.Id != callId {
+			return next(state, txBytes, isCheckTx)
+		}
+		// updated only if tx.Id == callId
+		if TxLimiter.lastUpdated+refreshInterval <= time.Now().Unix() {
```
ParthDesai: Need to have two condition here.

`TxLimiter.contractToTierMap == nil || TxLimiter.lastUpdated+refreshInterval <= time.Now().Unix()`

ParthDesai: Also, let's execute this code only in case tx is evm tx.

builtin/plugins/user_deployer_whitelist/user_deployer_whitelist.go
mukuls9971: done

throttle/contract_tx_limiter_middleware.go
mukuls9971: done

throttle/contract_tx_limiter_middleware.go
mukuls9971: done

throttle/contract_tx_limiter_middleware.go
```diff
@@ -13,7 +13,27 @@ import (
 	"github.com/pkg/errors"
 )
 
-const refreshInterval int64 = 15 * 60
+type TxLConfig struct {
```
ParthDesai: Rename it to `TxLimiterConfig`

throttle/contract_tx_limiter_middleware.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+func DefaultContractTxLimiterConfig() *ContractTxLimiterConfig {
+	return &ContractTxLimiterConfig{
+		RefreshInterval: 15 * 60,
+	}
+}
+
+// Clone returns a deep clone of the config.
+func (c *ContractTxLimiterConfig) Clone() *ContractTxLimiterConfig {
```
ParthDesai: No need for this, as struct only contains primitive fields

throttle/contract_tx_limiter_middleware.go
```diff
@@ -0,0 +1,143 @@
...
+
+type ContractTxLimiterConfig struct {
+	// Enables the middleware
+	Enabled bool
+	// Number of seconds each refresh lasts
+	RefreshInterval int64
+}
+
+func DefaultContractTxLimiterConfig() *ContractTxLimiterConfig {
+	return &ContractTxLimiterConfig{
```
ParthDesai: Set `Enabled` to `false` for sake of clarity.

throttle/middleware_test_helper.go
```diff
@@ -71,3 +71,57 @@ func throttleMiddlewareHandler(ttm loomchain.TxMiddlewareFunc, state loomchain.S
...
 	)
 }
+
+func throttleMiddlewareHandlerCheckTx(ttm loomchain.TxMiddlewareFunc, state loomchain.State, tx auth.SignedTx, ctx context.Context) (loomchain.TxHandlerResult, error) {
+	return ttm.ProcessTx(
+		state.WithContext(ctx),
+		tx.Inner,
+		func(state loomchain.State, txBytes []byte, isCheckTx bool) (res loomchain.TxHandlerResult, err error) {
+
+			var nonceTx loomAuth.NonceTx
```
ParthDesai: Why do we need to simulate this entire thing? We only care if tx reaches `next` or not

throttle/contract_tx_limiter_middleware_test.go
```diff
@@ -0,0 +1,124 @@
...
+	txSignedEVM2 := mockSignedTx(t, uint64(3), callId, vm.VMType_EVM, contractAddr)
+	cfg := DefaultContractTxLimiterConfig()
+	contractTxLimiterMiddleware := NewContractTxLimiterMiddleware(cfg,
+		func(state loomchain.State) (contractpb.Context, error) {
+			return contractpb.WrapPluginContext(udwContext), nil
+		},
+	)
+
+	ctx := context.WithValue(state.Context(), loomAuth.ContextKeyOrigin, owner)
+	_, err = throttleMiddlewareHandlerCheckTx(contractTxLimiterMiddleware, state, txSignedEVM, ctx)
```
ParthDesai: You also need to handle cases where chain height is varying.

throttle/contract_tx_limiter_middleware_test.go
```diff
@@ -0,0 +1,124 @@
...
+	require.NoError(t, err)
+	_, err = throttleMiddlewareHandlerCheckTx(contractTxLimiterMiddleware, state, txSignedEVM1, ctx)
```
ParthDesai: Since, we care only about whether or not we are reaching next func, you can directly invoke the middleware here. 

something like: 

```
allowed := false

middleware.ProcessTx(...., func(state loomchain.State, txBytes []byte, isCheckTx bool) (res loomchain.TxHandlerResult, err error) {
  allowed = true
  return loomchain.TxHandlerResult{}, nil
})

// Check if allowed should be true or false

```

throttle/contract_tx_limiter_middleware.go
```diff
@@ -0,0 +1,146 @@
...
+	contractToTierMap map[string]udw.Tier
+	// track of no. of txns in previous blocks per contract
+	contractToBlockTrx map[string]map[int64]int64
+	// to refresh the contractToTierMap when configuration changes
+	lastUpdated int64
+}
+
+var TxLimiter *contractTxLimiter
+
+func (txl *contractTxLimiter) isAccountLimitReached(contractAddr loom.Address, curBlockHeight int64) bool {
```
enlight: This is overly complicated. For each contract address keep track of the start of the block range, and a count of txs within that range, each time a tx comes in increment the counter, when the block range ends reset the counter and set the start of the block range to the current height.

throttle/contract_tx_limiter_middleware.go
```diff
@@ -0,0 +1,146 @@
...
+			return res, errors.Wrapf(err, "unmarshal call tx %v", msg.Data)
+		}
+		if msgTx.VmType == vm.VMType_EVM {
+			if TxLimiter.contractToTierMap == nil ||
+				TxLimiter.lastUpdated+cfg.RefreshInterval <= time.Now().Unix() {
+				ctx, err := createUserDeployerWhitelistCtx(state)
+				if err != nil {
+					return res, errors.Wrap(err, "throttle: context creation")
+				}
+				contractToTierMap, err := udw.GetContractTierMapping(ctx)
```
enlight: I don't like the idea of loading tier info for every contract ever deployed, only a small subset of them are likely to get called. Should load the tier info per-contract on an as-needed basis and cache it for some configurable interval.

throttle/contract_tx_limiter_middleware.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	clone := *c
+	return &clone
+}
+
+type contractTxLimiter struct {
+	// contract_address to limiting parametres structure
+	contractToInfoMap map[string]udw.ContractInfo
+	// track of no. of txns in previous blocks per contract
+	contractToBlockTrx map[string]blockTxn
```
ParthDesai: take this as pointer, so we wouldn't need to copy it around in `updateState`

throttle/contract_tx_limiter_middleware.go
```diff
@@ -0,0 +1,166 @@
...
+				// TxLimiter.lastUpdated = time.Now().Unix()
+			}
+			contractAddr := loom.UnmarshalAddressPB(msg.To)
+			//check if contract in list
+			contractInfo, ok := TxLimiter.contractToInfoMap[contractAddr.String()]
+			if !ok {
+				return loomchain.TxHandlerResult{}, ErrContractNotWhitelisted
+			}
+			// if its tierInfo is empty and its expiry is reached, update tier info
+			if contractInfo.TierInfo.BlockRange == 0 ||
```
ParthDesai: Keep tierID -> tierInfo mapping in different data structure, otherwise it will be lots of duplicate data.

ParthDesai: Also, I feel keeping expiry time for every contract is a bit overhead. What we need is to have expiry time for tierID -> tierInfo structure. So, that if we change tier information, we could reload it automatically.

throttle/contract_tx_limiter_middleware.go
```diff
@@ -0,0 +1,178 @@
...
+				ctx, err := createUserDeployerWhitelistCtx(state)
+				if err != nil {
+					return res, errors.Wrap(err, "throttle: context creation")
+				}
+				contractToTierMap, err := udw.GetContractTierMapping(ctx)
+				if err != nil {
+					return res, errors.Wrap(err, "throttle: contractToTierMap creation")
+				}
+				TxLimiter.contractToTierMap = contractToTierMap
+				// TxLimiter.lastUpdated = time.Now().Unix()
```
ParthDesai: Remove this comment.

throttle/contract_tx_limiter_middleware.go
```diff
@@ -0,0 +1,178 @@
...
+			}
+			contractAddr := loom.UnmarshalAddressPB(msg.To)
+			//check if contract in list
+			tierID, ok := TxLimiter.contractToTierMap[contractAddr.String()]
+			if !ok {
```
ParthDesai: Return from here, instead of giving an error. 

There could be contracts deployed by other means.

throttle/contract_tx_limiter_middleware.go
```diff
@@ -0,0 +1,178 @@
...
+		var msg vm.MessageTx
+		if err := proto.Unmarshal(tx.Data, &msg); err != nil {
+			return res, errors.Wrapf(err, "unmarshal message tx %v", tx.Data)
+		}
+		var msgTx vm.CallTx
+		if err := proto.Unmarshal(msg.Data, &msgTx); err != nil {
+			return res, errors.Wrapf(err, "unmarshal call tx %v", msg.Data)
+		}
+		if msgTx.VmType == vm.VMType_EVM {
+			if TxLimiter.contractToTierMap == nil {
```
ParthDesai: Have a expiry time for this map too.

throttle/contract_tx_limiter_middleware.go
```diff
@@ -0,0 +1,187 @@
...
+				ctx, err := createUserDeployerWhitelistCtx(state)
+				if err != nil {
+					return res, errors.Wrap(err, "throttle: context creation")
+				}
+				contractToTierMap, err := udw.GetContractTierMapping(ctx)
+				if err != nil {
+					return res, errors.Wrap(err, "throttle: contractToTierMap creation")
+				}
+				TxLimiter.contractToTierMap = contractToTierMap
+				TxLimiter.contractLastUpdated = time.Now().Unix()
```
ParthDesai: Rename this to `contractDataLastUpdated`

throttle/contract_tx_limiter_middleware.go
```diff
@@ -0,0 +1,187 @@
...
+					if er != nil {
+						return res, errors.Wrap(err, "throttle: context creation")
+					}
+					tierInfo, er := udw.GetTierInfo(ctx, tierID)
+					if er != nil {
+						return res, errors.Wrap(err, "throttle: getTierInfo issue")
+					}
+					TxLimiter.tierMap[tierID] = tierInfo
+				}
+				TxLimiter.tierLastUpdated = time.Now().Unix()
```
ParthDesai: Rename this to `tierDataLastUpdated`

throttle/contract_tx_limiter_middleware.go
```diff
@@ -0,0 +1,187 @@
...
+				TxLimiter.contractLastUpdated = time.Now().Unix()
+			}
+			contractAddr := loom.UnmarshalAddressPB(msg.To)
+			//check if contract in list
+			contractTierID, ok := TxLimiter.contractToTierMap[contractAddr.String()]
+			if !ok {
+				return next(state, txBytes, isCheckTx)
+			}
+			// update tierMap if expired
+			if TxLimiter.tierLastUpdated+cfg.TierRefreshInterval < time.Now().Unix() {
```
ParthDesai: this should be `cfg.TierDataRefreshInterval`

throttle/contract_tx_limiter_middleware.go
```diff
@@ -0,0 +1,187 @@
...
+		var msg vm.MessageTx
+		if err := proto.Unmarshal(tx.Data, &msg); err != nil {
+			return res, errors.Wrapf(err, "unmarshal message tx %v", tx.Data)
+		}
+		var msgTx vm.CallTx
+		if err := proto.Unmarshal(msg.Data, &msgTx); err != nil {
+			return res, errors.Wrapf(err, "unmarshal call tx %v", msg.Data)
+		}
+		if msgTx.VmType == vm.VMType_EVM {
+			if TxLimiter.contractToTierMap == nil || TxLimiter.contractLastUpdated+cfg.ContractRefreshInterval < time.Now().Unix() {
```
ParthDesai: This should be `cfg.ContractDataRefreshInterval`

throttle/contract_tx_limiter_middleware.go
```diff
@@ -0,0 +1,187 @@
...
+	blockTx.txn++
+}
+
+// NewContractTxLimiterMiddleware add another tx limiter that limits how many CallTx(s) can be sent to an EVM contract within a pre-configured block range
+func NewContractTxLimiterMiddleware(cfg *ContractTxLimiterConfig,
+	createUserDeployerWhitelistCtx func(state loomchain.State) (contractpb.Context, error),
+) loomchain.TxMiddlewareFunc {
+	TxLimiter = &contractTxLimiter{
+		contractToBlockTrx: make(map[string]*blockTxn, 0),
+		tierMap:            make(map[udw.TierID]udw.Tier, 0),
```
ParthDesai: No need to initialise this, here.

throttle/contract_tx_limiter_middleware.go
```diff
@@ -0,0 +1,187 @@
...
+				TxLimiter.contractDataLastUpdated = time.Now().Unix()
+			}
+			contractAddr := loom.UnmarshalAddressPB(msg.To)
+			//check if contract in list
+			contractTierID, ok := TxLimiter.contractToTierMap[contractAddr.String()]
+			if !ok {
+				return next(state, txBytes, isCheckTx)
+			}
+			// update tierMap if expired
+			if TxLimiter.tierDataLastUpdated+cfg.TierDataRefreshInterval < time.Now().Unix() {
```
ParthDesai: You need to add condition here: `TxLimiter.tierMap == nil`

throttle/contract_tx_limiter_middleware.go
mukuls9971: done

throttle/contract_tx_limiter_middleware.go
```diff
@@ -0,0 +1,191 @@
...
+				TxLimiter.contractToTierMap = contractToTierMap
+				TxLimiter.contractDataLastUpdated = time.Now().Unix()
+			}
+			contractAddr := loom.UnmarshalAddressPB(msg.To)
+			//check if contract in list
+			contractTierID, ok := TxLimiter.contractToTierMap[contractAddr.String()]
+			if !ok {
+				return next(state, txBytes, isCheckTx)
+			}
+			if TxLimiter.tierMap == nil {
```
ParthDesai: Let's make this bit uniform.

1. Create a static method in the contract, to read tierMap from the contract.
2. Have a condition here similar to `contractToTierMap` like `txLimiter.tierMap == nil || TxLimiter.tierDataLastUpdated+cfg.TierDataRefreshInterval < time.Now().Unix()`

builtin/plugins/user_deployer_whitelist/user_deployer_whitelist.go
```diff
@@ -360,4 +372,49 @@ func RecordEVMContractDeployment(ctx contract.Context, deployerAddress, contract
 	return nil
 }
 
+func GetTierMap(ctx contract.StaticContext) map[TierID]Tier {
```
ParthDesai: Dont ignore unmarshal error. Change return type to map and error

throttle/contract_tx_limiter_middleware.go
```diff
@@ -0,0 +1,180 @@
...
+	blockTx, ok := txl.contractToBlockTrx[contractAddr.String()]
+	if !ok {
+		txl.contractToBlockTrx[contractAddr.String()] = &blockTxn{1, curBlockHeight}
+		return
+	}
+	// reset blockTxn if block is out of range
+	tierID := txl.contractToTierMap[contractAddr.String()]
+	tier := txl.tierMap[tierID]
+
+	if blockTx.blockHeight < curBlockHeight-int64(tier.BlockRange) {
```
ParthDesai: Blockheight isnt always going to be sequential in reality. 

What you need to do is keep track of current block range.  And when you get current block height check to see if current block height falls under current block range, if yes you need to increament the tx count, otherwise migrate to block range for current block height and set number of tx to 1.

throttle/contract_tx_limiter_middleware.go
```diff
@@ -0,0 +1,184 @@
...
+func (txl *contractTxLimiter) isAccountLimitReached(contractAddr loom.Address, curBlockHeight int64) bool {
+	blockTx, ok := txl.contractToBlockTrx[contractAddr.String()]
+	if !ok {
+		return false
+	}
+	// if execution reaches here => tierID and tier are valid
+	tierID := txl.contractToTierMap[contractAddr.String()]
+	tier := txl.tierMap[tierID]
+	// reset contractToBlockTrx if curBlock is out of range
+	if blockTx.blockHeight <= curBlockHeight-int64(tier.BlockRange) {
```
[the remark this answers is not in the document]
mukuls9971: @Parth we have to reset state while isAccountLimitReached, a case: when after ratelimiting txns, a valid out-of-blockrange transaction comes,

about blockheight change, better we change the name from blockheight to blockRangeLowerBound, because changing the logic can create other edge cases

throttle/contract_tx_limiter_middleware.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+}
+
+func (txl *contractTxLimiter) updateState(contractAddr loom.Address, curBlockHeight int64) {
+	blockTx, ok := txl.contractToBlockTrx[contractAddr.String()]
+	tierID := txl.contractToTierMap[contractAddr.String()]
+	tier := txl.tierMap[tierID]
+	if !ok || blockTx.blockHeight <= curBlockHeight-int64(tier.BlockRange) {
+		// resetting the blockHeight to lower bound of range instead of curblockheight
+		rangeStart := (((curBlockHeight - 1) / int64(tier.BlockRange)) * int64(tier.BlockRange)) + 1
```
enlight: Isn't this equivalent to `rangeStart := curBlockHeight`?

throttle/contract_tx_limiter_middleware.go
```diff
@@ -0,0 +1,178 @@
...
+		}
+		var msg vm.MessageTx
+		if err := proto.Unmarshal(tx.Data, &msg); err != nil {
+			return res, errors.Wrapf(err, "unmarshal message tx %v", tx.Data)
+		}
+		var msgTx vm.CallTx
+		if err := proto.Unmarshal(msg.Data, &msgTx); err != nil {
+			return res, errors.Wrapf(err, "unmarshal call tx %v", msg.Data)
+		}
+		if msgTx.VmType == vm.VMType_EVM {
```
enlight: Get rid of one level of nesting by doing this:
```golang
if msgTx.VmType != vm.VMType_EVM {
  return next(state, txBytes, isCheckTx)
}
```

throttle/contract_tx_limiter_middleware.go
```diff
@@ -0,0 +1,178 @@
...
+				}
+				contractToTierMap, err := udw.GetContractTierMapping(ctx)
+				if err != nil {
+					return res, errors.Wrap(err, "throttle: contractToTierMap creation")
+				}
+				TxLimiter.contractToTierMap = contractToTierMap
+				TxLimiter.contractDataLastUpdated = time.Now().Unix()
+			}
+			contractAddr := loom.UnmarshalAddressPB(msg.To)
+			//check if contract in list
```
enlight: A useful comment would be
```golang
// contracts the limiter doesn't know about shouldn't be throttled
```

throttle/contract_tx_limiter_middleware.go
```diff
@@ -0,0 +1,178 @@
...
+					return res, errors.Wrap(err, "throttle: context creation")
+				}
+				TxLimiter.tierMap, err = udw.GetTierMap(ctx)
+				if err != nil {
+					return res, errors.Wrap(err, "throttle: GetTierMap error")
+				}
+				TxLimiter.tierDataLastUpdated = time.Now().Unix()
+			}
+
+			// check if tier corresponding to contract available in tierMap
```
enlight: `check if` -> `ensure that`

throttle/contract_tx_limiter_middleware.go
```diff
@@ -0,0 +1,178 @@
...
+	clone := *c
+	return &clone
+}
+
+type contractTxLimiter struct {
+	// contract_address to limiting parametres structure
+	contractToTierMap       map[string]udw.TierID
+	contractDataLastUpdated int64
+	// track of no. of txns in previous blocks per contract
+	contractToBlockTrx  map[string]*blockTxn
```
enlight: `contractToBlockTrx` -> `contractStatsMap`

throttle/contract_tx_limiter_middleware.go
```diff
@@ -0,0 +1,178 @@
...
+	tierMap             map[udw.TierID]udw.Tier
+	tierDataLastUpdated int64
+}
+
+type blockTxn struct {
```
enlight: `blockTxn` -> `contractStats`

throttle/contract_tx_limiter_middleware.go
mukuls9971: @enlight it is lower bound of current block-range, fo for 1-10 with a block range of 10 => it is 1, for 11-20 it is 11 and so on .. @ParthDesai

throttle/contract_tx_limiter_middleware.go
mukuls9971: done

throttle/contract_tx_limiter_middleware.go
mukuls9971: done

throttle/contract_tx_limiter_middleware.go
mukuls9971: done

builtin/plugins/user_deployer_whitelist/user_deployer_whitelist.go
```diff
@@ -360,4 +372,49 @@ func RecordEVMContractDeployment(ctx contract.Context, deployerAddress, contract
 	return nil
 }
 
+func GetTierMap(ctx contract.StaticContext) (map[TierID]Tier, error) {
+	tierMap := make(map[TierID]Tier, 0)
+	for _, rangeEntry := range ctx.Range([]byte(tierKeyPrefix)) {
+		var tier Tier
+		if err := proto.Unmarshal(rangeEntry.Value, &tier); err != nil {
+			return nil, errors.Wrap(err, "Marshalling error")
```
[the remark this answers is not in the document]
mukuls9971: done

throttle/contract_tx_limiter_middleware.go
ParthDesai: @enlight this is more on the line to store range start, instead of storing `curBlockHeight` at time of recalculation. Advantage here, is more uniform ranges, instead of ranges depending upon incoming blocks (as we could not guarantee blocks will come sequentially)

throttle/contract_tx_limiter_middleware.go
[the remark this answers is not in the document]
Parth: @ParthDesai **

PR #1253, opened 2019-06-20
Binance Dex Integration

phoorichet:

### DIFF
--- a/builtin/plugins/coin/coin.go
+++ b/builtin/plugins/coin/coin.go
@@ -12,6 +12,8 @@ import (
 	"github.com/loomnetwork/go-loom/types"
 	"github.com/loomnetwork/go-loom/util"
 	"github.com/loomnetwork/loomchain"
+
+	errUtil "github.com/pkg/errors"
 )
 
 const (
@@ -45,8 +47,7 @@ type (
 )
 
 var (
-	ErrSenderBalanceTooLow    = errors.New("sender balance is too low")
-	ErrFailedToResolveGateway = errors.New("failed to resolve gateway contract")
+	ErrSenderBalanceTooLow = errors.New("sender balance is too low")
 )
 
 var (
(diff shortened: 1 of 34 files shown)


builtin/plugins/coin/coin.go
```diff
@@ -107,33 +106,29 @@ func (c *Coin) Init(ctx contract.Context, req *InitRequest) error {
...
 // MintToGateway adds loom coins to the loom coin Gateway contract balance, and updates the total supply.
 func (c *Coin) MintToGateway(ctx contract.Context, req *MintToGatewayRequest) error {
-	gatewayAddr, err := ctx.Resolve("loomcoin-gateway")
-	if err != nil {
-		return errUtil.Wrap(err, "failed to mint Loom coin")
-	}
-
-	if ctx.Message().Sender.Compare(gatewayAddr) != 0 {
-		return errors.New("not authorized to mint Loom coin")
+	gatewayAddrs := resolveGatwayAddress(ctx, "loomcoin-gateway", "binance-gateway")
```
enlight: Not sure what the plan here is. You expect there to be two coin contracts? If there's just one it'll always mint to `loomcoin-gateway`, so seems pointless to attempt to resolve the address of the `binance-gateway`. If there are two coin contracts then we should just change how this works so we store the gateway address in the coin contract state, and then have an owner-only method to set it.

phoorichet: I think there should be only one coin contract. The issue is `binance-gateway` is trying to mint/burn LoomCoin when the amount user requested is insufficient. So the request get rejected since only `loomcoin-gateway` is able to perform those operations. 

If the only authorized contract on LoomCoin is only `loomcoin-gateway`, then probably we need to have some interface for `binance-gateway` to let `loomcoin-gateway` mint/burn LoomCoin.

enlight: I guess we'll have to add a list of accounts to the coin contract that are allowed to mint, that list will include both loomcoin-gateway and binance-gateway. Anyway, we can probably leave this alone for now, this won't get shipped until after the generic Binance integration, so we can deal with this later.

enlight: I've reverted the changes to coin for now. Will find a better way to do this later.

phoorichet: OK.

PR #1261, opened 2019-06-24
Don't set status to 0x0 on error

Sriep:

Possible fix for returning status="0x0" for succesful transactions.
For go-transactions it can't find the receipt, but the most of the information for the tendermint blockstore is good, including the status. So no need to change this.
Could also hide the error in [getReceiptByTendermintHash](https://github.com/loomnetwork/loomchain/blob/master/rpc/query_server.go#L1110).


rpc/query_server.go
```diff
@@ -764,13 +764,8 @@ func (s *QueryServer) EthGetTransactionReceipt(hash eth.Data) (resp eth.JsonTxRe
...
 	}
 	if err != nil {
-		resp, err = getReceiptByTendermintHash(snapshot, s.BlockStore, r, txHash)
-		if err != nil {
-			// return empty response if cannot find hash
-			resp.Status = eth.EncInt(int64(StatusTxFail))
-			return resp, nil
-		}
-		return resp, err
+		resp, _ = getReceiptByTendermintHash(snapshot, s.BlockStore, r, txHash)
```
enlight: If you're just going to ignore the error from `getReceiptByTendermintHash` then there's no point in `getReceiptByTendermintHash` returning an error at all. Furthermore, in `getReceiptByTendermintHash` if `blockStore.GetTxResult(hash)` doesn't error out then you can look at `txResults.Result.Code` to see if the tx succeeded or not and set the tx receipt status to match.

Sriep: Moved the error ignoring into `getReceiptByTendermintHash` as there are some errors that should probably not be ignored.

eduardonunesp:
LGTM

PR #1265, opened 2019-06-25
Add on-chain validator information tracking 

atchapcyp:
Here is the list of method we need to implement on `ChainConfig` contract.

-`SetValidatorInfo` <- To let validator set their information before start-up node
-`GetValidatorInfo` <- To inquire validator information
-`ListValidatorsInfo` <-To get all information from validators

method SetBuildNumber or SetValidatorInfo must only be use in ChainConfig version 1.2

About the CLI command :

- `loom chain-cfg get-validator-info`
- `loom chain-cfg set-validator-info`
- `loom chain-cfg list-validators-info`
For printing out build number of each validator, and percentage

[WIP]ChainConfig routine to automatically update their own build number

### DIFF
--- a/builtin/plugins/chainconfig/chainconfig.go
+++ b/builtin/plugins/chainconfig/chainconfig.go
@@ -582,16 +582,13 @@ func (c *ChainConfig) SetValidatorInfo(ctx contract.Context, req *SetValidatorIn
 	if !isValidator {
 		return ErrNotAuthorized
 	}
-	return setValidatorInfo(ctx, senderAddr, req.BuildNumber)
-}
 
-func setValidatorInfo(ctx contract.Context, addr loom.Address, buildNumber uint64) error {
 	validator := &ValidatorInfo{
-		Address:     addr.MarshalPB(),
-		BuildNumber: buildNumber,
+		Address:     senderAddr.MarshalPB(),
+		BuildNumber: req.BuildNumber,
 		UpdatedAt:   uint64(time.Now().Unix()),
 	}
-	return ctx.Set(validatorInfoKey(addr), validator)
+	return ctx.Set(validatorInfoKey(senderAddr), validator)
 }
 
 func (c *ChainConfig) GetValidatorInfo(ctx contract.StaticContext, req *GetValidatorInfoRequest) (*GetValidatorInfoResponse, error) {
--- a/builtin/plugins/chainconfig/chainconfig.go
+++ b/builtin/plugins/chainconfig/chainconfig.go
@@ -562,6 +562,7 @@ func (c *ChainConfig) SetValidatorInfo(ctx contract.Context, req *SetValidatorIn
 	if !ctx.FeatureEnabled(loomchain.ChainCfgVersion1_2, false) {
 		return ErrFeatureNotEnabled
 	}
+
 	if req.BuildNumber == 0 {
 		return ErrInvalidRequest
 	}
--- a/builtin/plugins/chainconfig/chainconfig.go
+++ b/builtin/plugins/chainconfig/chainconfig.go
@@ -1,8 +1,6 @@
 package chainconfig
 
 import (
-	"fmt"
-
 	"github.com/gogo/protobuf/proto"
 	loom "github.com/loomnetwork/go-loom"
 	cctypes "github.com/loomnetwork/go-loom/builtin/types/chainconfig"
@@ -75,6 +73,9 @@ var (
 	ErrFeatureNotSupported = errors.New("[ChainConfig] feature is not supported in the current build")
 	// ErrFeatureNotFound indicates that a feature does not exist
 	ErrFeatureNotFound = errors.New("[ChainConfig] feature not found")
+	// ErrFeatureNotEnabled indacates that a feature has not been enabled
+	// by majority of validators, and has not been activated on the chain.
+	ErrFeatureNotEnabled = errors.New("[ChainConfig] feature not enabled")
 )
 
 const (
@@ -557,20 +558,28 @@ func removeFeature(ctx contract.Context, name string) error {
 var Contract plugin.Contract = contract.MakePluginContract(&ChainConfig{})
 
 func (c *ChainConfig) SetValidatorInfo(ctx contract.Context, req *SetValidatorInfoRequest) error {
+
+	if !ctx.FeatureEnabled(loomchain.ChainCfgVersion1_2, false) {
+		return ErrFeatureNotEnabled
+	}
 	if req.BuildNumber == 0 {
 		return ErrInvalidRequest
 	}
-
 	senderAddr := ctx.Message().Sender
-	fmt.Println(senderAddr.String())
-	// validators, err := getCurrentValidators(ctx)
-	// for _, validator := range validators {
-	// 	if validator.Compare(senderAddr) == 0 {
-
-	// 	} else {
-	// 		return ErrNotAuthorized
-	// 	}
-	// }
+	validators, err := getCurrentValidators(ctx)
+	if err != nil {
+		return err
+	}
+	isValidator := false
+	for _, validator := range validators {
+		if validator.Compare(senderAddr) == 0 {
+			isValidator = true
+			break
+		}
+	}
+	if !isValidator {
+		return ErrNotAuthorized
+	}
 	return setValidatorInfo(ctx, senderAddr, req.BuildNumber)
 }
 
@@ -606,4 +615,22 @@ func getValidatorInfo(ctx contract.StaticContext, addr loom.Address) (*Validator
 		return nil, errors.Wrap(err, "failed to retrieve chainconfig validatorinfo")
 	}
 	return &validatorInfo, nil
+}
+
+// ListValidatorInfo return info validators
+func (c *ChainConfig) ListValidatorInfo(ctx contract.StaticContext, req *ListValidatorInfoRequest) (*ListValidatorInfoResponse, error) {
+
+	validatorRange := ctx.Range([]byte(validatorInfoPrefix))
+	validators := []*ValidatorInfo{}
+	for _, m := range validatorRange {
+		var v ValidatorInfo
+		if err := proto.Unmarshal(m.Value, &v); err != nil {
+			return nil, errors.Wrapf(err, "unmarshal validators %s", string(m.Key))
+		}
+		validators = append(validators, &v)
+	}
+
+	return &ListValidatorInfoResponse{
+		Validators: validators,
+	}, nil
 }
--- a/builtin/plugins/chainconfig/chainconfig.go
+++ b/builtin/plugins/chainconfig/chainconfig.go
@@ -1,8 +1,6 @@
 package chainconfig
 
 import (
-	"time"
-
 	"github.com/gogo/protobuf/proto"
 	loom "github.com/loomnetwork/go-loom"
 	cctypes "github.com/loomnetwork/go-loom/builtin/types/chainconfig"
@@ -560,12 +558,11 @@ func removeFeature(ctx contract.Context, name string) error {
 var Contract plugin.Contract = contract.MakePluginContract(&ChainConfig{})
 
 func (c *ChainConfig) SetValidatorInfo(ctx contract.Context, req *SetValidatorInfoRequest) error {
-	if !ctx.FeatureEnabled(loomchain.ChainCfgVersion1_2, false) {
-		return ErrFeatureNotEnabled
-	}
-
 	if req.BuildNumber == 0 {
 		return ErrInvalidRequest
+	}
+	if !ctx.FeatureEnabled(loomchain.ChainCfgVersion1_2, false) {
+		return ErrFeatureNotEnabled
 	}
 	senderAddr := ctx.Message().Sender
 	validators, err := getCurrentValidators(ctx)
@@ -586,7 +583,7 @@ func (c *ChainConfig) SetValidatorInfo(ctx contract.Context, req *SetValidatorIn
 	validator := &ValidatorInfo{
 		Address:     senderAddr.MarshalPB(),
 		BuildNumber: req.BuildNumber,
-		UpdatedAt:   uint64(time.Now().Unix()),
+		UpdatedAt:   uint64(ctx.Now().Unix()),
 	}
 	return ctx.Set(validatorInfoKey(senderAddr), validator)
 }
@@ -609,7 +606,7 @@ func (c *ChainConfig) GetValidatorInfo(ctx contract.StaticContext, req *GetValid
 	}, nil
 }
 
-// ListValidatorsInfo return info validators
+// ListValidatorsInfo returns the build number for each validators
 func (c *ChainConfig) ListValidatorsInfo(ctx contract.StaticContext, req *ListValidatorsInfoRequest) (*ListValidatorsInfoResponse, error) {
 	validatorRange := ctx.Range([]byte(validatorInfoPrefix))
 	validators := []*ValidatorInfo{}
--- a/builtin/plugins/chainconfig/chainconfig.go
+++ b/builtin/plugins/chainconfig/chainconfig.go
@@ -16,21 +16,27 @@ import (
 )
 
 type (
-	InitRequest           = cctypes.InitRequest
-	ListFeaturesRequest   = cctypes.ListFeaturesRequest
-	ListFeaturesResponse  = cctypes.ListFeaturesResponse
-	GetFeatureRequest     = cctypes.GetFeatureRequest
-	GetFeatureResponse    = cctypes.GetFeatureResponse
-	AddFeatureRequest     = cctypes.AddFeatureRequest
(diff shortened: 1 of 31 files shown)


builtin/plugins/chainconfig/chainconfig.go
```diff
@@ -542,3 +556,82 @@ func removeFeature(ctx contract.Context, name string) error {
 }
 
 var Contract plugin.Contract = contract.MakePluginContract(&ChainConfig{})
+
+func (c *ChainConfig) SetValidatorInfo(ctx contract.Context, req *SetValidatorInfoRequest) error {
+
```
pathornteng: Remove this empty line

builtin/plugins/chainconfig/chainconfig.go
```diff
@@ -542,3 +556,82 @@ func removeFeature(ctx contract.Context, name string) error {
...
+		}
+	}
+	if !isValidator {
+		return ErrNotAuthorized
+	}
+	return setValidatorInfo(ctx, senderAddr, req.BuildNumber)
+}
+
+func setValidatorInfo(ctx contract.Context, addr loom.Address, buildNumber uint64) error {
+
```
pathornteng: Ditto

builtin/plugins/chainconfig/chainconfig.go
```diff
@@ -542,3 +556,82 @@ func removeFeature(ctx contract.Context, name string) error {
...
+	validator := &ValidatorInfo{
+		Address:     addr.MarshalPB(),
+		BuildNumber: buildNumber,
+	}
+	return ctx.Set(validatorInfoKey(addr), validator)
+}
+
+func (c *ChainConfig) GetValidatorInfo(ctx contract.StaticContext, req *GetValidatorInfoRequest) (*GetValidatorInfoResponse, error) {
+
```
pathornteng: Ditto

builtin/plugins/chainconfig/chainconfig.go
```diff
@@ -542,3 +556,82 @@ func removeFeature(ctx contract.Context, name string) error {
...
+	if err != nil {
+		return nil, err
+	}
+	return &GetValidatorInfoResponse{
+		Validator: validatorInfo,
+	}, nil
+}
+
+func getValidatorInfo(ctx contract.StaticContext, addr loom.Address) (*ValidatorInfo, error) {
+
```
pathornteng: Ditto

builtin/plugins/chainconfig/chainconfig.go
```diff
@@ -542,3 +556,82 @@ func removeFeature(ctx contract.Context, name string) error {
...
+	err := ctx.Get(validatorInfoKey(addr), &validatorInfo)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to retrieve chainconfig validatorinfo")
+	}
+	return &validatorInfo, nil
+}
+
+// ListValidatorsInfo return info validators
+func (c *ChainConfig) ListValidatorsInfo(ctx contract.StaticContext, req *ListValidatorsInfoRequest) (*ListValidatorsInfoResponse, error) {
+
```
pathornteng: Ditto

chainconfig/chainconfig_routine.go
```diff
@@ -90,3 +90,16 @@ func (cc *ChainConfigRoutine) run() {
...
+
+func (cc *ChainConfigRoutine) init() {
+	if cc.node.IsValidator() {
+		dappClient := client.NewDAppChainRPCClient(cc.chainID, cc.cfg.DAppChainWriteURI, cc.cfg.DAppChainReadURI)
+		chainConfigClient, err := NewChainConfigClient(dappClient, cc.address, cc.signer, cc.logger)
+		if err != nil {
+			cc.logger.Error("Failed to create ChainConfigClient", "err", err)
+		} else {
+			// NOTE: errors are logged by the client, no need to log again
+			_ = chainConfigClient.SetBuildNumber(cc.buildNumber)
```
pathornteng: if error is ignored, we can just do
```
chainConfigClient.SetBuildNumber(cc.buildNumber)
```

cmd/loom/chainconfig/cmd.go
```diff
@@ -324,6 +327,99 @@ func RemoveFeatureCmd() *cobra.Command {
...
+	}
+	cli.AddContractStaticCallFlags(cmd.Flags(), &flags)
+	return cmd
+}
+
+const listValidatorsInfoCmdExample = `
+loom chain-cfg list-validator-info 
+`
+
+func ListValidatorsInfoCmd() *cobra.Command {
```
pathornteng: Could you please make this function similar to `ListFeatures`?

cmd/loom/chainconfig/cmd.go
```diff
@@ -324,6 +327,99 @@ func RemoveFeatureCmd() *cobra.Command {
...
+		},
+	}
+	cli.AddContractCallFlags(cmd.Flags(), &flags)
+	cmdFlags := cmd.Flags()
+	cmdFlags.Uint64Var(&buildNumber, "build", 0, "Set a validator's information ")
+	return cmd
+}
+
+const getValidatorInfoCmdExample = `
+loom chain-cfg get-validator-info --key
```
pathornteng: change `--key` to some fake address like `default:0x7262d4c97c7B93937E4810D289b7320e9dA82857`

pathornteng:
When the node starts, it will try to `SetBuildNumber` only once. If it fails, the validator information won't be updated until the next restart. We can fix this by keep calling `SetBuildNumber` every interval. What do you think? @enlight

builtin/plugins/chainconfig/chainconfig.go
```diff
@@ -542,3 +558,78 @@ func removeFeature(ctx contract.Context, name string) error {
...
+		return nil, err
+	}
+	return &GetValidatorInfoResponse{
+		Validator: validatorInfo,
+	}, nil
+}
+
+func getValidatorInfo(ctx contract.StaticContext, addr loom.Address) (*ValidatorInfo, error) {
+	var validatorInfo ValidatorInfo
+	err := ctx.Get(validatorInfoKey(addr), &validatorInfo)
```
enlight: It's better not to return an error if the validator info isn't found:
```golang
  return &ValidatorInfo{}, nil
} else if err != nil {
  return nil, errors.Wrap(err, "failed to retrieve validator info")
}
```
Makes the client side code a bit less messy.

builtin/plugins/chainconfig/chainconfig_test.go
```diff
@@ -175,6 +179,31 @@ func (c *ChainConfigTestSuite) TestFeatureFlagEnabledSingleValidator() {
...
+	pctx.SetFeature(loomchain.ChainCfgVersion1_2, true)
+	err = chainconfigContract.SetValidatorInfo(ctx, &SetValidatorInfoRequest{
+		BuildNumber: 0,
+	})
+	require.Equal(ErrInvalidRequest, err)
+	err = chainconfigContract.SetValidatorInfo(ctx, &SetValidatorInfoRequest{
+		BuildNumber: buildNumber,
+	})
+	require.NoError(err)
+	getValidatorInfo, err := chainconfigContract.GetValidatorInfo(ctx, &GetValidatorInfoRequest{Address: addr1.MarshalPB()})
```
enlight: `getValidatorInfo` -> `validatorInfo`

builtin/plugins/chainconfig/chainconfig_test.go
```diff
@@ -405,6 +439,30 @@ func (c *ChainConfigTestSuite) TestFeatureFlagEnabledFourValidators() {
...
+	require.Equal(ErrInvalidRequest, err)
+	err = chainconfigContract.SetValidatorInfo(contractpb.WrapPluginContext(pctx.WithSender(addr1)), &SetValidatorInfoRequest{
+		BuildNumber: buildNumber,
+	})
+	require.NoError(err)
+	err = chainconfigContract.SetValidatorInfo(contractpb.WrapPluginContext(pctx.WithSender(addr2)), &SetValidatorInfoRequest{
+		BuildNumber: buildNumber,
+	})
+	require.NoError(err)
+	getValidatorInfo, err := chainconfigContract.GetValidatorInfo(ctx, &GetValidatorInfoRequest{Address: addr1.MarshalPB()})
```
enlight: `getValidatorInfo` -> `validatorInfo`

features.go
```diff
@@ -49,6 +49,9 @@ const (
 	// Enables usage of ctx.Validators() in ChainConfig contract.
 	ChainCfgVersion1_1 = "chaincfg:v1.1"
 
+	// Enables usage of Set ValidatorInfo in ChainConfig contract.
```
enlight: -> `// Enables validator build number tracking via the ChainConfig contract.`

cmd/loom/chainconfig/cmd.go
```diff
@@ -324,6 +328,136 @@ func RemoveFeatureCmd() *cobra.Command {
...
+	}
+	cli.AddContractStaticCallFlags(cmd.Flags(), &flags)
+	return cmd
+}
+
+const sumValidatorsInfoCmdExample = `
+loom chain-cfg sum-validators-info 
+`
+
+func SumValidatorsInfoCmd() *cobra.Command {
```
enlight: Don't need a separate command for the summary, just print the summary at the end of the validator list printed out by the `list-validators-info` command

cmd/loom/chainconfig/cmd.go
```diff
@@ -324,6 +328,136 @@ func RemoveFeatureCmd() *cobra.Command {
...
+}
+
+const listValidatorsInfoCmdExample = `
+loom chain-cfg list-validators-info 
+`
+
+func ListValidatorsInfoCmd() *cobra.Command {
+	var flags cli.ContractCallFlags
+	cmd := &cobra.Command{
+		Use:     "list-validators-info",
```
enlight: `list-validators-info` -> `list-validators`

chainconfig/chainconfig_routine.go
```diff
@@ -84,9 +84,22 @@ func (cc *ChainConfigRoutine) run() {
...
+
+func (cc *ChainConfigRoutine) init() {
+	if cc.node.IsValidator() {
+		dappClient := client.NewDAppChainRPCClient(cc.chainID, cc.cfg.DAppChainWriteURI, cc.cfg.DAppChainReadURI)
+		chainConfigClient, err := NewChainConfigClient(dappClient, cc.address, cc.signer, cc.logger)
+		if err != nil {
+			cc.logger.Error("Failed to create ChainConfigClient", "err", err)
+		} else {
+			// NOTE: errors are logged by the client, no need to log again
+			chainConfigClient.SetBuildNumber(cc.buildNumber)
```
enlight: This should be done once in `run()`, I don't think we need this separate `init()` function.

enlight:
> When the node starts, it will try to `SetBuildNumber` only once. If it fails, the validator information won't be updated until the next restart. We can fix this by keep calling `SetBuildNumber` every interval. What do you think?

The node should keep calling `SetBuildNumber` every interval until the call succeeds. However, before calling `SetBuildNumber` it should check if the current build number on-chain is already correct or not, if the build number stored in ChainConfig already matches the build number of the node then it shouldn't call `SetBuildNumber` again.

cmd/loom/chainconfig/cmd.go
```diff
@@ -324,6 +328,136 @@ func RemoveFeatureCmd() *cobra.Command {
...
+		Short:   "list validator informations",
```
enlight: -> `Show info stored for each validator`

cmd/loom/chainconfig/cmd.go
```diff
@@ -324,6 +328,136 @@ func RemoveFeatureCmd() *cobra.Command {
...
+		Example: listValidatorsInfoCmdExample,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			var resp cctype.ListValidatorsInfoResponse
+			err := cli.StaticCallContractWithFlags(&flags, chainConfigContractName, "ListValidatorsInfo",
+				&cctype.ListValidatorsInfoRequest{}, &resp)
+			if err != nil {
+				return err
+			}
+			out, err := formatJSON(&resp)
```
enlight: Display this info as a table, similar to how you've done the feature list.

```
validator | build
-----------------
0xblah    | 1023
```

atchapcyp: Should we arrange list as alphabetically by validator or the build number ?

enlight: Lets order by build number, lowest to highest

chainconfig/chainconfig_routine.go
```diff
@@ -85,6 +85,23 @@ func (cc *ChainConfigRoutine) run() {
 			} else {
 				// NOTE: errors are logged by the client, no need to log again
 				_ = chainConfigClient.VoteToEnablePendingFeatures(cc.buildNumber)
+				if !isBuildNumberEqual {
+					validatorInfo, err := chainConfigClient.GetValidatorInfo()
+					if err != nil {
+						cc.logger.Error("Failed to retreive build number", "err", err)
+					} else {
+						buildNumber := validatorInfo.Validator.GetBuildNumber()
```
pathornteng: There is a chance that `validatorInfo.Validator` is nil. We need to check this before calling  `validatorInfo.Validator.GetBuildNumber()`

builtin/plugins/chainconfig/chainconfig.go
```diff
@@ -542,3 +558,70 @@ func removeFeature(ctx contract.Context, name string) error {
...
 var Contract plugin.Contract = contract.MakePluginContract(&ChainConfig{})
+
+func (c *ChainConfig) SetValidatorInfo(ctx contract.Context, req *SetValidatorInfoRequest) error {
+	if !ctx.FeatureEnabled(loomchain.ChainCfgVersion1_2, false) {
+		return ErrFeatureNotEnabled
+	}
+
+	if req.BuildNumber == 0 {
+		return ErrInvalidRequest
+	}
```
enlight: Do this first.

builtin/plugins/chainconfig/chainconfig.go
```diff
@@ -542,3 +558,70 @@ func removeFeature(ctx contract.Context, name string) error {
...
+		}
+	}
+	if !isValidator {
+		return ErrNotAuthorized
+	}
+
+	validator := &ValidatorInfo{
+		Address:     senderAddr.MarshalPB(),
+		BuildNumber: req.BuildNumber,
+		UpdatedAt:   uint64(time.Now().Unix()),
```
enlight: Never use `time.Now()` in a contract, it will differ between nodes, thus changing persistent state in a slightly different way on every node in the cluster, since the state change is inconsistent the nodes will never reach consensus and the chain will halt. Though it's strange the e2e test hasn't failed.

You want to use `ctx.Now()` instead, that's the block timestamp, which is going to be consistent across the cluster.

builtin/plugins/chainconfig/chainconfig.go
```diff
@@ -542,3 +558,70 @@ func removeFeature(ctx contract.Context, name string) error {
...
+		return &GetValidatorInfoResponse{}, nil
+	} else if err != nil {
+		return nil, errors.Wrap(err, "failed to retrieve validator info")
+	}
+	return &GetValidatorInfoResponse{
+		Validator: &validatorInfo,
+	}, nil
+}
+
+// ListValidatorsInfo return info validators
```
enlight: -> `// ListValidatorsInfo returns the build number for each validator`

chainconfig/chainconfig_routine.go
```diff
@@ -85,6 +85,25 @@ func (cc *ChainConfigRoutine) run() {
 			} else {
 				// NOTE: errors are logged by the client, no need to log again
 				_ = chainConfigClient.VoteToEnablePendingFeatures(cc.buildNumber)
+				if !isBuildNumberEqual {
+					validatorInfo, err := chainConfigClient.GetValidatorInfo()
+					if err != nil {
+						cc.logger.Error("Failed to retreive build number", "err", err)
+					} else if validatorInfo.Validator == nil {
+						chainConfigClient.SetBuildNumber(0) // Set build number to 0 when validator never vote before
```
enlight: This isn't going to work right now since `SetValidatorInfo` will error out if you give it a zero build number, but that's not what you want to do here anyway. The way this should work is if `validatorInfo.Validator == nil` then assume that `buildNumber` is zero and therefore call `SetBuildNumber(cc.buildNumber)` just like you would if the build didn't match.

chainconfig/chainconfig_routine.go
```diff
@@ -85,6 +85,25 @@ func (cc *ChainConfigRoutine) run() {
...
+					} else {
+						buildNumber := validatorInfo.Validator.GetBuildNumber()
+						if buildNumber < cc.buildNumber {
```
enlight: This should probably be `buildNumber != cc.buildNumber`

PR #1267, opened 2019-06-25
Dump block txs scanner

pathornteng:
This PR adds a command to scan and dump txs starting from a specific block height.
```
loom dump-block-txs-scanner --height 12345 --uri http://host:port --update-height --to 0x0000 --from 0x000
```


enlight:
Will keep this PR around for reference, sometimes we need to build one-off scanners like these for migrations.

Issue #1268, opened 2019-06-25
JSON RPC should expose the Ethereum stored on account by calling `eth_getBalance` on address


eduardonunesp:
The `JSON RPC` call for `eth_getBalance` isn't exposing the value correctly for a given address, although it works on the previous protobuf interface it didn't work for `JSON RPC` new interface needing return the Ethereum for a given account.

For instance, the call below doesn't return the correct value

```bash
>> {"id": 0, "jsonrpc":"2.0", "method": "eth_getBalance", "params": ["0x53c22427e7D0061A6A83F4fc3184b25033b9b0FB"]}
<< {
  "result": "0x0",
  "jsonrpc": "2.0",
  "id": 0
}
``` 

And the address above has some value stored on plasmachain, as we can see on command `loom gateway account`

```bash
$ loom gateway account default:0x53c22427e7D0061A6A83F4fc3184b25033b9b0FB -u http://plasma.dappchains.com:80
{
  "DAppChainAddress": "0x53c22427e7D0061A6A83F4fc3184b25033b9b0FB",
  "EthereumAddress": "0x86D4859bE772b5c3445A9AA7e82ad930369309f7",
  "LOOM": "0 (0)",
  "ETH": "0 (15184681712962962)"
}```

Sriep:
Implemented in https://github.com/loomnetwork/loomchain/pull/1269

PR #1269, opened 2019-06-26
eth_getBalance call ethCoin method

Sriep:

Implement https://github.com/loomnetwork/loomchain/issues/1268

### DIFF
--- a/cmd/loom/loom.go
+++ b/cmd/loom/loom.go
@@ -1141,7 +1141,6 @@ func deployContract(
 }
 
 type contextFactory func(state loomchain.State) (contractpb.Context, error)
-type staticContextFactory func(state loomchain.State) (contractpb.StaticContext, error)
 
 func getContractCtx(pluginName string, vmManager *vm.Manager) contextFactory {
 	return func(state loomchain.State) (contractpb.Context, error) {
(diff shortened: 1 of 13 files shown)


plugin/vm.go
```diff
@@ -333,3 +333,13 @@ func NewInternalContractContext(contractName string, pluginVM *PluginVM) (contra
 	readOnly := false
 	return contractpb.WrapPluginContext(pluginVM.CreateContractContext(caller, contractAddr, readOnly)), nil
 }
+
+
+func NewInternalStaticContractContext(contractName string, pluginVM *PluginVM) (contractpb.StaticContext, error) {
```
enlight: Add `readOnly` param to `NewInternalContractContext`, no need to have a separate function for this.

cmd/loom/loom.go
```diff
@@ -1211,6 +1222,32 @@ func initQueryService(
...
+				state,
+				createRegistry(state),
+				&loomchain.DefaultEventHandler{},
+				log.Default,
+				newABMFactory,
+				receiptWriter,
+				receiptReader,
+			), nil
+		},
+	)
```
enlight: Much of this is not needed to query EthCoin, see `createAddressMapperCtx` in `QueryServer` for an example of a minimal Go contract context. I don't think you need to change anything in `loom.go`.

PR #1275, opened 2019-06-29
Show_Features_Enabled_By_Migration

suhasagg:
This enables feature to be listed via list-features command

Sets feature in chainconfig contract context at the time of migration.
This enables feature to be listed via list-features command

### DIFF
--- a/builtin/plugins/chainconfig/chainconfig.go
+++ b/builtin/plugins/chainconfig/chainconfig.go
@@ -187,7 +187,7 @@ func (c *ChainConfig) AddFeature(ctx contract.Context, req *AddFeatureRequest) e
 }
 
 
-func AddSpecficFeature(ctx contract.Context, name string, blockHeight uint64) error {
+func AddSpecificFeature(ctx contract.Context, name string, blockHeight uint64) error {
 	if name == "" {
 		return ErrInvalidRequest
 	}
--- a/builtin/plugins/chainconfig/chainconfig.go
+++ b/builtin/plugins/chainconfig/chainconfig.go
@@ -187,25 +187,29 @@ func (c *ChainConfig) AddFeature(ctx contract.Context, req *AddFeatureRequest) e
 }
 
 //Adds specific feature with name provided
-func AddSpecificFeature(ctx contract.Context, name string, blockHeight uint64) error {
+func SyncEnabledFeature(ctx contract.Context, name string, blockHeight uint64) error {
+	var feature Feature
 	if name == "" {
 		return ErrInvalidRequest
 	}
-	if found := ctx.Has(featureKey(name)); found {
-		return ErrFeatureAlreadyExists
-	}
-	feature := Feature{
-		Name:        name,
-		Status:      FeatureEnabled,
-		BlockHeight: blockHeight,
+	if err := ctx.Get(featureKey(name), &feature); err == nil {
+		feature.Status = FeatureEnabled
+	} else {
+		if err != contract.ErrNotFound {
+			return err
+		} else {
+			feature = Feature{
+				Name:        name,
+				Status:      FeatureEnabled,
+				BlockHeight: blockHeight,
+			}
+		}
 	}
 	if err := ctx.Set(featureKey(name), &feature); err != nil {
 		return err
 	}
 	return nil
 }
-
-
 
 // RemoveFeature should be called by the contract owner to remove features.
 // NOTE: Features can only be removed before they're activated by the chain.
@@ -517,7 +521,6 @@ func enableFeature(ctx contract.Context, name string) error {
 
 	return ctx.Set(featureKey(name), &feature)
 }
-
 
 func addFeature(ctx contract.Context, name string, buildNumber uint64, autoEnable bool) error {
 	if name == "" {
--- a/builtin/plugins/chainconfig/chainconfig.go
+++ b/builtin/plugins/chainconfig/chainconfig.go
@@ -186,7 +186,7 @@ func (c *ChainConfig) AddFeature(ctx contract.Context, req *AddFeatureRequest) e
 	return nil
 }
 
-
+//Adds specific feature with name provided
 func AddSpecificFeature(ctx contract.Context, name string, blockHeight uint64) error {
 	if name == "" {
 		return ErrInvalidRequest
--- a/builtin/plugins/chainconfig/chainconfig.go
+++ b/builtin/plugins/chainconfig/chainconfig.go
@@ -186,6 +186,27 @@ func (c *ChainConfig) AddFeature(ctx contract.Context, req *AddFeatureRequest) e
 	return nil
 }
 
+
+func AddSpecficFeature(ctx contract.Context, name string, blockHeight uint64) error {
+	if name == "" {
+		return ErrInvalidRequest
+	}
+	if found := ctx.Has(featureKey(name)); found {
+		return ErrFeatureAlreadyExists
+	}
+	feature := Feature{
+		Name:        name,
+		Status:      FeatureEnabled,
+		BlockHeight: blockHeight,
+	}
+	if err := ctx.Set(featureKey(name), &feature); err != nil {
+		return err
+	}
+	return nil
+}
+
+
+
 // RemoveFeature should be called by the contract owner to remove features.
 // NOTE: Features can only be removed before they're activated by the chain.
 func (c *ChainConfig) RemoveFeature(ctx contract.Context, req *RemoveFeatureRequest) error {
@@ -496,6 +517,7 @@ func enableFeature(ctx contract.Context, name string) error {
 
 	return ctx.Set(featureKey(name), &feature)
 }
+
 
 func addFeature(ctx contract.Context, name string, buildNumber uint64, autoEnable bool) error {
 	if name == "" {
--- a/cmd/loom/loom.go
+++ b/cmd/loom/loom.go
@@ -838,7 +838,7 @@ func loadApp(
 		CreateRegistry: createRegistry,
 		Migrations: map[int32]tx_handler.MigrationFunc{
 			1: migrations.DPOSv3Migration,
-			3: migrations.ConsolidateFeaturesMigration,
+			2: migrations.ConsolidateFeaturesMigration,
 		},
 	}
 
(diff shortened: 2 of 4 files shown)


suhasagg:
Same change will be required for migration ID 2 as well i.e Coin Policy Migration

pathornteng:
This change is not going to work on the production since this migration function has already been called. The place that the logic should be implemented is in `plugin/chain_config_manager.go`. In `plugin/chain_config_manager.go` a function `ConsolidateFeatures` should be implemented and called every begin block like `EnableFeatures`. The example is https://github.com/loomnetwork/loomchain/blob/00dcc1eebbe9e9cac20a67b9dbe2c3b6668215a5/app.go#L434
You will need to implement a function `chainconfig.ListEnabledFeatures` in `ChainConfig` contract. `chainconfig.ListEnabledFeatures` returns a list of enabled features in `ChainConfig`. So you can compare the enabled features in the contract and the enabled features in store state and then add the enabled features found only in store state to `ChainConfig` contract. In order to do so, a function `chainconfig.AddFeature` is needed to be implemented to directly add the enabled features to the contract.
This change needs a feature flag (probably ChainConfigv.1.3) to activate.

ParthDesai:
@pathornteng I think it's better if we just write a migration to consolidate the list once and for all, instead calling `ConsolidateFeatures` in begin block. And implementation of future migrations can be modified to add code to modify chainconfig context.

Let me know what do you think?

pathornteng:
> @pathornteng I think it's better if we just write a migration to consolidate the list once and for all, instead of calling `ConsolidateFeatures` in begin block. And implementation of future migrations can be modified to add code to modify chainconfig context.
> 
> Let me know what do you think?

Just checked the code, we don't really call `SetFeature` anywhere beside of DPOSv3 migration function, so I think your solution works and is better in term of performance. Yeah I think let's go for it but we need to make sure that in all the future migration functions also add the feature to `ChainConfig` context.

ParthDesai:
cool. There is another migration we are writing that uses `StoreState.SetFeature`, but its not been merged yet. So, @suhasagg let's modify it to also set feature in `chainconfig` contract

suhasagg:
Ok, I will complete migration to consolidate features.
Have set feature in chainconfig contract in both migration Ids.

Suhas

ParthDesai:
@suhasagg no need to make any changes to migration already completed.

suhasagg:
Consolidate Features Migration has been added.
Consolidate Features Migration will consolidate features stored in chainconfig contract and loomchain.state (These features stored in state are enabled via migration)

suhasagg:
For current and further migrations, Migrations which enable features in loomchain.state, these features will be stored in chaniconfig contract context as well so that they can be displayed by list features command of chainconfig plugin.

cmd/loom/loom.go
```diff
@@ -838,6 +838,7 @@ func loadApp(
 		CreateRegistry: createRegistry,
 		Migrations: map[int32]tx_handler.MigrationFunc{
 			1: migrations.DPOSv3Migration,
+			3: migrations.ConsolidateFeaturesMigration,
```
ParthDesai: Why this is numbered as 3?

migrations/consolidatefeatures_migration.go
```diff
@@ -0,0 +1,29 @@
...
+	cctypes "github.com/loomnetwork/go-loom/builtin/types/chainconfig"
+)
+
+func ConsolidateFeaturesMigration(ctx *MigrationContext) error {
+	// Pull data from ChainConfig
+	_, chainconfigCtx, err := resolveChainConfig(ctx)
+	if err != nil {
+		return err
+	}
+	featuresFromState := ctx.state.Range([]byte(featurePrefix))
```
ParthDesai: In case feature key does not exists, only then you directly set feature key with feature struct.

migrations/dposv3_migration.go
```diff
@@ -3,21 +3,35 @@ package migrations
...
+	cctypes "github.com/loomnetwork/go-loom/builtin/types/chainconfig"
 	"github.com/loomnetwork/go-loom/plugin/contractpb"
 	"github.com/loomnetwork/loomchain"
 	"github.com/loomnetwork/loomchain/builtin/plugins/dposv2"
 	"github.com/loomnetwork/loomchain/builtin/plugins/dposv3"
 	"github.com/loomnetwork/loomchain/config"
 )
 
+const (
+	featurePrefix = "ft"
```
ParthDesai: Import this from chainconfig plugin

migrations/dposv3_migration.go
```diff
@@ -55,6 +69,13 @@ func DPOSv3Migration(ctx *MigrationContext) error {
 
 	// Switch over to DPOSv3
 	ctx.State().SetFeature(loomchain.DPOSVersion3Feature, true)
+	// Set feature in chainconfig Contract
+	var feature cctypes.Feature
+	feature.Status = cctypes.Feature_ENABLED
+	feature.BlockHeight = uint64(chainconfigCtx.Block().Height)
+	if err := chainconfigCtx.Set(featureKey(loomchain.DPOSVersion3Feature), &feature); err != nil {
```
ParthDesai: Instead of this, write a method in chainconfig contract to enable feature.

migrations/dposv3_migration.go
```diff
@@ -67,3 +88,12 @@ func resolveDPOSv2(ctx *MigrationContext) (loom.Address, contractpb.Context, err
...
 }
+
+func resolveChainConfig(ctx *MigrationContext) (loom.Address, contractpb.Context, error) {
+	chainconfigCtx, err := ctx.ContractContext("chainconfig")
+	if err != nil {
+		return loom.Address{}, nil, err
+	}
+	chainconfigAddr := chainconfigCtx.ContractAddress()
+	return chainconfigAddr, chainconfigCtx, nil
+}
```
ParthDesai: Add a new line.

cmd/loom/loom.go
suhasagg: 2nd is reserved for Coin Policy Migration. It will be merged in from Coin Policy PR.

migrations/consolidatefeatures_migration.go
suhasagg: Done.

migrations/dposv3_migration.go
suhasagg: Done.

builtin/plugins/chainconfig/chainconfig.go
```diff
@@ -186,6 +186,27 @@ func (c *ChainConfig) AddFeature(ctx contract.Context, req *AddFeatureRequest) e
 	return nil
 }
 
+//Adds specific feature with name provided
+func AddSpecificFeature(ctx contract.Context, name string, blockHeight uint64) error {
```
ParthDesai: Name should be along the line of `SyncEnabledFeature`.

Also, in case feature already exists, toggle its status to `Enabled`, otherwise create a new feature and store it.

suhasagg: Have one question, is it necessary migration will always enable a feature? it might enable a feature, disable another feature,  so depending on what was done at the time of migration, we need to preserve that state right ..
So in case feature already exists, it might have enabled/disabled state

ParthDesai: We are querying enabled features from state only.

migrations/consolidatefeatures_migration.go
```diff
@@ -0,0 +1,31 @@
...
+	for _, m := range featuresFromState {
```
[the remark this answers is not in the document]
suhasagg: Done.

builtin/plugins/chainconfig/chainconfig.go
suhasagg: Done

suhasagg: One question, though we are querying for enabled features , should not we store all the features enabled/disabled at the time of migration in chainconfig context, so that state and context are in sync.

enlight:
The consolidation should happen on read, it shouldn't require duplicating the data. You add a new `EnabledFeatures` method to `ReadOnlyState`:
```
type ReadOnlyState interface {
  // ...
  FeatureEnabled(string, bool) bool
  EnabledFeatures() []string
}
`EnabledFeatures()` should return all the enabled features in the store state.
Then in `ChainConfig.ListFeatures` you consolidate the `ctx.EnabledFeatures()` with the ones in the contract.

**Comment by suhasagg** [comment]
```
func (c *ChainConfigManager) EnableFeatures(blockHeight int64) error {
	features, err := chainconfig.EnableFeatures(c.ctx, uint64(blockHeight), c.build)
	if err != nil {
		// When an unsupported feature has been activated by the rest of the chain
		// panic to prevent the node from processing any further blocks until it's
		// upgraded to a new build that supports the feature.
		if err == chainconfig.ErrFeatureNotSupported {
			panic(err)
		}
		return err
	}
	for _, feature := range features {
		c.state.SetFeature(feature.Name, true)
	}
	return nil
}
```
In chainconfigManager, features which are enabled are getting stored in chainconfig context as well as loomchain.State.
So, I thought features should be in sync in both loomchain.state and chaincfg context, isn't there redundancy here as well.

enlight:
> In chainconfigManager, features which are enabled are getting stored in chainconfig context as well as loomchain.State.
> So, I thought features should be in sync in both loomchain.state and chaincfg context, isn't there redundancy here as well.

Yes, there is some data redundancy with enabled features that couldn't be avoided, but we don't need to make it worse.

suhasagg:
How do I access store state in a go contract method where only contract context is available.

pathornteng:
> How do I access store state in a go contract method where only contract context is available.

You need to implement `EnabledFeatures() []string` and expose it to contract context as Vadim mentioned. You have to implement this function interface many places both in go-loom and loomchain. Try searching `FeatureEnabled` in go-loom and loomchain and you will see all the places that you need to add  `EnabledFeatures` interface. The actual implement will be in `app.go` as a method of `StoreState`. Once the interface and the function have been added, any contract can have access to this `EnabledFeatures` using `ctx.EnabledFeatures()`.

suhasagg:
For features enabled via migration we don't have properties blockheight and build number, should we ignore these properties while consolidating, do we need them in later stages of blockchain.

enlight:
@suhasagg For features enabled via migration just assume the build number & height are zero.

suhasagg:
@Vadim can a feature be disabled at the time of migration say disabling outdated dpos feature and enabling a new version feature and we need to consolidate this feature as well in output of chain cfg List Features?

enlight:
@suhasagg No, at this point in time features can only be enabled, not disabled.

PR #1276, opened 2019-06-30
downtime slashing

aupiff:
Implements slashing which is triggered by four consecutive downtime periods during which a validator exceeds the `maximumDowntimePercentage`.

Downtime slashing only occurs once per downtime period, when the `ShiftDowntimeWindow` function is  called & `(uint64(currentHeight)%state.Params.DowntimePeriod) == 0`.


builtin/plugins/dposv3/dpos.go
```diff
@@ -1349,19 +1354,42 @@ func ShiftDowntimeWindow(ctx contract.Context, currentHeight int64, candidates [
 	if state.Params.DowntimePeriod != 0 && (uint64(currentHeight)%state.Params.DowntimePeriod) == 0 {
 		ctx.Logger().Info("DPOS ShiftDowntimeWindow", "block", currentHeight)
 
+		downtimeSlashingEnabled := ctx.FeatureEnabled(loomchain.DPOSVersion3_4, false)
+		maximumMissedBlocksBig := CalculateFraction(state.Params.MaxDowntimePercentage.Value, loom.BigUInt{big.NewInt(int64(state.Params.DowntimePeriod))})
```
pathornteng: Once this feature has been deployed, `state.Params.MaxDowntimePercentage` is going to be nil. So we need check it before dereferencing it.

builtin/plugins/dposv3/dpos.go
```diff
@@ -1349,19 +1354,42 @@ func ShiftDowntimeWindow(ctx contract.Context, currentHeight int64, candidates [
...
+				slash := true;
+				for i := uint64(0); i < 4; i++ {
+					if (maximumMissedBlocks < (statistic.RecentlyMissedBlocks >> (16*i) & 0xFFFF)) {
+						slash = false;
+						break;
+					}
+				}
+
+				if slash {
+					SlashInactivity(ctx, candidateAddress)
```
pathornteng: `error` is ignored here

builtin/plugins/dposv3/util.go
```diff
@@ -107,6 +107,15 @@ func validateFee(fee uint64) error {
 	return nil
 }
 
+func validatePercentage(percentageBig loom.BigUInt) error {
+	percentage := percentageBig.Int.Int64()
+	if percentage < 0 || percentage > 10000 {
```
pathornteng: Can you make `10000` a constant?

builtin/plugins/dposv3/dpos.go
```diff
@@ -1349,15 +1355,53 @@ func ShiftDowntimeWindow(ctx contract.Context, currentHeight int64, candidates [
...
+				}
+
+				if slash {
+					if err := SlashInactivity(ctx, candidateAddress); err != nil {
+						return err
+					}
+
+					// load updated statistic that was modified by the call to
+					// SlashInactivity
+					statistic, err = GetStatistic(ctx, candidateAddress)
```
aupiff: To avoid this statistic-reloading, I might make `SlashInactivity` simply return a modified statistic.

builtin/plugins/dposv3/dpos.go
```diff
@@ -1349,15 +1355,53 @@ func ShiftDowntimeWindow(ctx contract.Context, currentHeight int64, candidates [
...
 					continue
 				}
 				return err
 			}
 
+			if downtimeSlashingEnabled {
+				slash := true;
+				downtime := getDowntimeRecord(ctx, statistic)
+				for i := uint64(0); i < 4; i++ {
+					if (maximumMissedBlocks >= downtime.Periods[i]) {
```
pathornteng: Shouldn't we check only the last period? It seems that a validator is being slashed 4 times for a period that exceeds the maximumMissedBlock.

aupiff: No, the documents explain that if ANY of the last four periods include fewer missed blocks than the maximum, then there is no slashing. it's `slash := true` by default. It's a very lenient policy, very similar to what cosmos does.

pathornteng: aha, my bad, this is where we unslash not slash

builtin/plugins/dposv3/dpos.go
```diff
@@ -1349,15 +1355,53 @@ func ShiftDowntimeWindow(ctx contract.Context, currentHeight int64, candidates [
...
+			statistic, err := GetStatistic(ctx, candidateAddress)
 			if err != nil {
 				if err == contract.ErrNotFound {
 					continue
 				}
 				return err
 			}
 
+			if downtimeSlashingEnabled {
+				slash := true;
```
enlight: What's this semi-colon doing here? :)

builtin/plugins/dposv3/dpos.go
```diff
@@ -1349,15 +1355,53 @@ func ShiftDowntimeWindow(ctx contract.Context, currentHeight int64, candidates [
...
+				downtime := getDowntimeRecord(ctx, statistic)
+				for i := uint64(0); i < 4; i++ {
+					if (maximumMissedBlocks >= downtime.Periods[i]) {
+						slash = false;
+						break;
```
enlight: Semi-colons

builtin/plugins/dposv3/dpos.go
enlight: Or you can pass the statistic into `SlashInactivity` and let it modify it.

aupiff: yeah, that's what I meant.

builtin/plugins/dposv3/dpos.go
```diff
@@ -1450,45 +1484,34 @@ func getDowntimeRecord(ctx contract.StaticContext, statistic *ValidatorStatistic
 }
 
 // only called for validators, never delegators
-func SlashInactivity(ctx contract.Context, validatorAddr []byte) error {
+func SlashInactivity(ctx contract.Context, statistic *ValidatorStatistic) (*ValidatorStatistic, error) {
```
enlight: Seeing as the input `statistic` is modified directly there's no need to return it.

aupiff: good point, made change.

builtin/plugins/dposv3/dpos.go
```diff
@@ -1450,45 +1484,34 @@ func getDowntimeRecord(ctx contract.StaticContext, statistic *ValidatorStatistic
...
-
-	// If slashing percentage is less than current total slash percentage, do
-	// not further increase total slash percentage during this election period
-	if slashPercentage.Cmp(&statistic.SlashPercentage.Value) < 0 {
-		return nil
-	}
-
+func slash(ctx contract.Context, statistic *ValidatorStatistic, slashPercentage loom.BigUInt) error {
 	updatedAmount := common.BigZero()
 	updatedAmount.Add(&statistic.SlashPercentage.Value, &slashPercentage)
```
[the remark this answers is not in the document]
aupiff: thanks for pointing this out--it should never happen but I did remove the guard that was in place before, which was to lenient. I should replace it with a guard that maxes the slashing out at 100%.

builtin/plugins/dposv3/dpos.go
```diff
@@ -2191,6 +2219,33 @@ func (c *DPOS) SetSlashingPercentages(ctx contract.Context, req *SetSlashingPerc
 	return saveState(ctx, state)
 }
 
+func (c *DPOS) SetMaxDowntimePercentage(ctx contract.Context, req *SetMaxDowntimePercentageRequest) error {
+	if !ctx.FeatureEnabled(loomchain.DPOSVersion3_4, false) {
+		return errors.New("DPOS v3.4 is not enabled")
+	}
```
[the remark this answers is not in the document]
aupiff: added a check

pathornteng:
I think we should also make e2e test work. Should we stop slashing after the validator is jailed? 
Also we now have two special commands for e2e tests that maybe useful for making e2e test more reliable.
`wait_for_block_height_to_increase [node_number] [number_of_blocks_to_increase]`
`wait_node_to_start [node_number] [max_waiting_time_in_seconds]`

e2e/dpos-downtime.toml
```diff
@@ -121,3 +126,11 @@
...
   Condition = "contains"
   Expected = ["2000000000000000000"]
+
+# Node 1 staked 1.25M in order to become a validator, if the node was properly
+# slashed, the node's self-delegation should now be a bit less than 1.25M so the
+# figure "125" should not be present in the response.
+[[TestCases]]
+  RunCmd = "{{ $.LoomPath }} dpos3 check-delegation {{index $.NodeAddressList 1}} {{index $.NodeAddressList 1}}"
+#  Condition = "excludes"
+#  Excluded = ["125"]
```
enlight: Why is this commented out?

enlight: Ah ok, flaky tests presumably. @pathornteng can probably fix them up.

aupiff: yep, flakey tests -- supplemented w/ unit tests.

aupiff:
> I think we should also make e2e test work. Should we stop slashing after the validator is jailed?
> Also we now have two special commands for e2e tests that maybe useful for making e2e test more reliable.
> `wait_for_block_height_to_increase [node_number] [number_of_blocks_to_increase]`
> `wait_node_to_start [node_number] [max_waiting_time_in_seconds]`

Slashing does stop after jailing. Good to know about those two new e2e tests.

pathornteng:
Since this PR https://github.com/loomnetwork/loomchain/pull/1327 has been merged and it is using DPOSv3.4 feature flag, not sure if we have to bump up the feature flag version of this PR or they can use the same version.

enlight:
We should add a toggle for the slashing, similar to the one we have for jailing, then in `ShiftDowntimeWindow` use the toggle var to figure out if slashing should happen. And the rest of the changes can just use the same `dpos:v3.4` feature flag.

Issue #490, opened 2018-12-03
Solidity contracts should not have names


Sriep:
Currently, we can create Solidity contracts with a name by filling in the DeployTx.Name field when deploying a contract.

enlight:
We'll add a config option to allow named EVM contracts so the tests can keep working.

enlight:
Resolved by https://github.com/loomnetwork/loomchain/pull/1282

Issue #328, opened 2018-09-03
DeployTxHandler.ProcessTx() silently ignores contract registration errors


enlight:
If a named contract fails to register the tx should fail.
```
reg := h.CreateRegistry(state)
reg.Register(tx.Name, addr, caller) // this may return an error that no one checks
```

enlight:
Somewhat related issue https://github.com/loomnetwork/loomchain/issues/490

enlight:
This will require a feature flag.

PR #1282, opened 2019-07-01
Add EVM Contract Name option

atchapcyp:
Since, We don't need solidity contract to have name and still need to compatible with `loom-js`.
We should add `EVMConfig` field to config file to be an option.

 This also need to check with the feature flag to avoid consensus break.

ref: https://github.com/loomnetwork/loomchain/issues/490, https://github.com/loomnetwork/loomchain/issues/328


cmd/loom/loom.go
```diff
@@ -1133,9 +1134,12 @@ func deployContract(
 		return err
 	}
 
-	err = registry.Register(contractCfg.Name, addr, addr)
-	if err != nil {
-		return err
+	conf, err := common.ParseConfig()
```
pathornteng: We should not reparse the config file here. The config should be passed as a function argument.

cmd/loom/loom.go
```diff
@@ -1133,9 +1134,12 @@ func deployContract(
...
+	if !state.FeatureEnabled(loomchain.EvmContractNameFeature, false) || conf.EvmConfig.AllowNamedEvmContract || vmType == lvm.VMType_PLUGIN {
```
pathornteng: The correct logic should be 
```
if EvmConfig.AllowNamedEvmContract || vmType == lvm.VMType_PLUGIN {
	err = registry.Register(contractCfg.Name, addr, addr)
	// if CheckRegistryErrorFeature is enabled, handle the error
	if err != nil && state.FeatureEnabled(loomchain.CheckRegistryErrorFeature, false) { 
		return r, err
	}
}
```

cmd/loom/loom.go
```diff
@@ -1133,9 +1136,11 @@ func deployContract(
 		return err
 	}
 
-	err = registry.Register(contractCfg.Name, addr, addr)
-	if err != nil {
-		return err
+	if chainCfg.EvmConfig.AllowNamedEvmContract || vmType == lvm.VMType_PLUGIN {
```
enlight: This `deployContract` function doesn't need to be changed at all, it only gets called when a new chain is initialized to deploy contracts specified in the chain genesis, and we can assume whoever writes the genesis knows what they're doing.

config/config.go
```diff
@@ -134,7 +134,8 @@ type Config struct {
 
 	Auth *auth.Config
 
-	EvmStore *evm.EvmStoreConfig
+	EvmStore  *evm.EvmStoreConfig
+	EvmConfig *evm.EvmConfig
```
enlight: Don't need a whole subsection, just add a top level setting named `AllowNamedEVMContracts` that defaults to `false`.

vm/handler.go
```diff
@@ -78,8 +79,13 @@ func (h *DeployTxHandler) ProcessTx(
...
 
-	reg := h.CreateRegistry(state)
-	reg.Register(tx.Name, addr, caller)
+	if h.AllowNamedEVMContract || tx.VmType == VMType_PLUGIN {
+		reg := h.CreateRegistry(state)
+		err := reg.Register(tx.Name, addr, caller)
+		if err != nil && state.FeatureEnabled(loomchain.CheckRegistryErrorFeature, false) { // if feature enable is on then handle the error
+			return r, err
+		}
+	}
```
enlight: This should go before `vm, err := h.Manager.InitVM(tx.VmType, state)`:
```golang
version1_1 := state.FeatureEnabled(loomchain.DeployTxVersion1_1Feature, false)
if version1_1 && (tx.VmType == VMType_EVM) && (len(tx.Name) > 0) && !h.AllowNamedEVMContracts {
  return r, errors.New("named evm contracts are not allowed")
}
```

And then the second part can be this:
```golang
if err := reg.Register(tx.Name, addr, caller); err != nil && version1_1 {
  return r, err
}
```

config/config.go
```diff
@@ -722,6 +724,14 @@ EvmStore:
   NumCachedRoots: {{.EvmStore.NumCachedRoots}}
 {{end}}
 
+{{if .EvmConfig -}}
+#
+# EvmConfig
+#
+EvmConfig:
+  AllowNamedEvmContract: {{ .EvmConfig.AllowNamedEvmContract }}
```
enlight: Move this setting to the `Here be dragons` section

PR #1290, opened 2019-07-03
Switch back to certusone/yubihsm-go branch

pathornteng:
We have already upstreamed the commit we need. The upstream branch and our branch are almost the same, so switching back to the upstream branch.


enlight:
Lets make sure this dependency is locked to a specific git revision.

pathornteng:
We need to switch back go-loom and transfer-gateway branches before merging.

Note: jailing offline validator e2e test is still flaky. Need to figure out why.

Issue #1287, opened 2019-07-02
Make dpos3 subcommands more comprehensible 


pathornteng:
Currently, `loom dpos3` subcommands print raw response from the server. The output is not very comprehensible because they are not formatted. 

Some of the fields that need to be reformatted
1. Addresses -> hex string
2. LockTime -> iso date time
3. LockTimeTier -> (`TIER_ZERO`=2 weeks, `TIER_ONE`=3 months,`TIER_TWO`= 6 months, `TIER_THREE`=1 year)
3. Amount/Fee -> base 10 number system (string)
4. Index 0 -> need a flag to indicate that this is rewards

enlight:
Here's one of the first ones we should clean up:
```
loom dpos3 downtime-record
```
The output should look like this:
```
name        | address | period 1 | period 2 | period 3 | period 4
-------------------------------------------------------------
MyValidator | 0xblah | 123 | 0 | 234 | 135
```

PR #1305, opened 2019-07-05
Change dpos3 downtime record response 

atchapcyp:
This PR change dpos3 downtime-record response from JSON to table.
Subtask of  https://github.com/loomnetwork/loomchain/issues/1287

### DIFF
--- a/cmd/loom/dposV3_commands.go
+++ b/cmd/loom/dposV3_commands.go
@@ -554,7 +554,7 @@ func DowntimeRecordCmdV3() *cobra.Command {
 							Address:        loom.UnmarshalAddressPB(d.GetValidator()).Local.String(),
 							Name:           c.Candidate.GetName(),
 							DownTimeRecord: d,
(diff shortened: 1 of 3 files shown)


enlight:
Need make some tweaks:
- Sort validators in alphabetical order.
- Add another column called `jailed`, the value should be `Y` or `N` to indicate whether a validator has been jailed.
- Numbers in the period columns should be right-aligned.

cmd/loom/dposV3_commands.go
```diff
@@ -1076,3 +1134,11 @@ func NewDPOSV3Command() *cobra.Command {
 	)
 	return cmd
 }
+
+func boolToString(b bool) string {
+	if b == true {
```
pathornteng: Simplify this one
```
if b {
  return "Y"
}
return "N"
```

pathornteng: Second thought, just use `strconv.FormatBool(b)`

Issue #1304, opened 2019-07-05
Add an address format option flag to addressmapper commands 


pathornteng:
We should add an option flag `--hex` to these addressmapper commands 
```
loom addressmapper get-mapping 
loom addressmapper list-mappings 
```
So that they return mapping addresses in hexadecimal.

enlight:
Hex should be the default, I don't think we need a `--base64` flag even, just output hex.
Also `list-mappings` should be displayed in a table similar to how the contract mapping are displayed by this command:
```
loom gateway list-contract-mappings -u http://test-z-us1.dappchains.com:80
```

PR #1309, opened 2019-07-08
Addressmapper command response address should shown as Hex

atchapcyp:
These command should show response as Hex
```
loom addressmapper get-mapping 
loom addressmapper list-mappings
```
ref : https://github.com/loomnetwork/loomchain/issues/1304


pathornteng:
Currently, if we just call `get-mapping` without any argument, it will `index out of range` error. Please add 
`Args:  cobra.ExactArgs(1)` to `get-mapping` command

atchapcyp:
`From` - `To` printing twice is my intension since the result has too many row 
till user can't see the table header and don't understand which one is `from` or `to`.

Should we print once at the bottom ?

enlight:
> Should we print once at the bottom ?

Yes, mappings are bi-directional so it doesn't really matter which order the columns appear in, and there are only two columns so they're not going to forget.

cmd/loom/addressmapper.go
```diff
@@ -128,11 +135,26 @@ func ListMappingCmd() *cobra.Command {
...
+			type maxLength struct {
+				From int
+				To   int
+			}
+			ml := maxLength{From: 50, To: 50}
+			toList := make(map[string]bool)
+			for _, value := range resp.Mappings {
+				fromAddr := loom.UnmarshalAddressPB(value.From).String()
+				toAddr := loom.UnmarshalAddressPB(value.To).String()
+				if !toList[fromAddr] {
```
pathornteng: This is kind of confusing. Address mapper is 1-1 mapping between 2 addresses, so you can just check if the address is found before. To simplify the algorithm, we can do
```
addressList := make(map[string]bool)
for _, value := range resp.Mappings {
  fromAddr := loom.UnmarshalAddressPB(value.From).String()
  toAddr := loom.UnmarshalAddressPB(value.To).String()
  if addressList[fromAddr] || addressList[toAddr] {
     continue
  }
  addressList[toAddr] = true
  addressList[fromAddr] = true
  fmt.Printf("%-*s | %-*s |\n",
      ml.From, fromAddr,
      ml.To, toAddr)
}

File: cmd/loom/addressmapper.go [hunk 2]
```diff
@@ -102,15 +103,21 @@ func GetMapping() *cobra.Command {
 			if err != nil {
 				return err
 			}
-			out, err := formatJSON(&resp)
-			if err != nil {
-				return err
+			type maxLength struct {
+				From   int
+				To     int
+				Status int
```
pathornteng: What is this status? if not used, please remove it.

cmd/loom/addressmapper.go
```diff
@@ -128,11 +134,25 @@ func ListMappingCmd() *cobra.Command {
...
+					continue
+				}
+				addressList[fromAddr] = true
+				addressList[toAddr] = true
+				fmt.Printf("%-*s | %-*s |\n",
+					ml.From, fromAddr,
+					ml.To, toAddr)
 			}
-			fmt.Println(out)
+			fmt.Printf("%-*s | %-*s |\n", ml.From, "From", ml.To, "To")
```
enlight: The header should go at the top. Also lets drop the final `|` (from the header & the body) - it looks awkward, so the format string should just be `"%-*s | %-*s\n"`

cmd/loom/addressmapper.go
```diff
@@ -102,15 +103,20 @@ func GetMapping() *cobra.Command {
...
+				To   int
 			}
-			fmt.Println(out)
+			ml := maxLength{From: 50, To: 50}
+
+			fmt.Printf("%-*s | %-*s |\n", ml.From, "From", ml.To, "To")
+
+			fmt.Printf("%-*s | %-*s |\n",
+				ml.From, loom.UnmarshalAddressPB(resp.From).String(),
+				ml.To, loom.UnmarshalAddressPB(resp.To).String())
```
enlight: Table is overkill since there's only one mapping, can just output a single line, e.g.
```
default:0xdeadbeef123 -> eth:0xdeadbeef321
```

cmd/loom/addressmapper.go
enlight: I actually wanted this filtering to happen in the contract, not in the command.

builtin/plugins/address_mapper/address_mapper.go
```diff
@@ -136,18 +136,22 @@ func (am *AddressMapper) ListMapping(ctx contract.StaticContext, req *ListMappin
 	listMappingResponse := ListMappingResponse{
 		Mappings: []*AddressMapping{},
 	}
-
+	addressList := make(map[string]bool)
```
enlight:
```golang
addressList := make(map[string]bool, len(mappingRange)/2)
```

PR #1313, opened 2019-07-08
Convert log evm root mismatch form error to log error

Sriep:

### DIFF
--- a/store/multi_writer_app_store.go
+++ b/store/multi_writer_app_store.go
@@ -84,8 +84,8 @@ func NewMultiWriterAppStore(appStore *IAVLStore, evmStore *EvmStore, saveEVMStat
 	}
 	evmStoreEvmRoot, version := store.evmStore.getLastSavedRoot(store.appStore.Version())
 	if !bytes.Equal(appStoreEvmRoot, evmStoreEvmRoot) {
-		return nil, fmt.Errorf("EVM roots mismatch, evm.db(%d): %X, app.db(%d): %X",
-			version, evmStoreEvmRoot, appStore.Version(), appStoreEvmRoot)
+		log.Error(fmt.Sprintf("EVM roots mismatch, evm.db(%d): %X, app.db(%d): %X",
+			version, evmStoreEvmRoot, appStore.Version(), appStoreEvmRoot))
 	}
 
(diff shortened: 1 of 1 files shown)


pathornteng:
Unfortunately, we cannot ignore the evm root mismatch error. It is a fatal error as the height of evm.db and app.db are not the same. This probably causes consensus failure later on.

Sriep:
So I stop a node then call `loom run` and get a 
````I[2019-07-08|13:33:01.240] Loading Multi-Writer App Store               module=loom 
Error: EVM roots mismatch, evm.db(32): 250E14A528FE0D4665DC309903B451D2614BFDE56B72C27814238F40D8A70A92, app.db(35): DF11774E9B0E71565D6A94F6B7228A7F18AC117F4C6A8842FF116F799DD7629E
Usage:
  loom run [root contract] [flags]```
response.
I now should run 
`loom debug set-app-height ./app.db --height ??`
what height should I use? Or maybe I am saying shouldn't the error message give the height plus the correct command to run (thinking of idiot user here)?

**Comment by pathornteng** [comment]
> So I stop a node then call `loom run` and get a
> 
> ```
> Error: EVM roots mismatch, evm.db(32): 250E14A528FE0D4665DC309903B451D2614BFDE56B72C27814238F40D8A70A92, app.db(35): DF11774E9B0E71565D6A94F6B7228A7F18AC117F4C6A8842FF116F799DD7629E
> Usage:
>   loom run [root contract] [flags]
> ```
> 
> response.
> I now should run
> `loom debug set-app-height ./app.db --height ??`
> what height should I use? Or maybe I am saying shouldn't the error message give the height plus the correct command to run (thinking of idiot user here = me atm)? Or maybe automatically run the correcting command, without erroring out.

you have to set the app.db height to match evm.db height. So in your case, you should do
```
loom debug set-app-height ./app.db --height 32 
```
Yes, we can add a flag like --evm-height or something so that our command will find the latest height of evm.db and then replay app.db from that height

enlight:
Going to close this now since this error should be fatal, and shouldn't happen under normal circumstances, further discussion can go in https://github.com/loomnetwork/loomchain/issues/1335 if needed.

PR #1319, opened 2019-07-09
[WIP] Cache Feature Flags and Cfg Settings

pathornteng:
Features and Cfg settings are going to be read from app.db every block by several managers (such as dpos manager and chainconfig manager). We should cache it in `MultiWriterAppStore` to reduce IOPS.


pathornteng:
This PR is no longer needed.

Issue #1307, opened 2019-07-08
Add ListReferrers method to DPOSv3 contract


pathornteng:
Currently, there is no way to get the registered referrer information on DPOS contract. We need method `ListReferrers` that returns a list of registered referrers. A command to get a list of referrers should be
```
loom dpos3 list-referrers 
```

enlight:
Resolved

PR #1320, opened 2019-07-09
Add list-referrers method to dpos3 contract

atchapcyp:
This PR relates to issue #1307 

- Change the way to Set/Get referrer name method from directly append key and value Now using `util.PrefixKey` instead.

- Add `loom dpos3 list-referrers` command

- Add some message type to `go-loom`


builtin/plugins/dposv3/storage.go
```diff
@@ -26,8 +27,13 @@ var (
 	referrersKey   = []byte("referrers")
```
pathornteng: Delete this key if it is not used anymore.

e2e/dposv3-delegation.toml
```diff
@@ -261,6 +261,11 @@
 [[TestCases]]
   RunCmd = "{{ $.LoomPath }} coin approve dposV3 40 -k {{index $.NodePrivKeyPathList 2}}"
 
+[[TestCases]]
+  RunCmd = "{{ $.LoomPath }} dpos3 list-referrers"
+  Condition = "contains"
+  Excluded = ["acct2","acct1"]
```
pathornteng: Should it be `Expected`?

pathornteng:
This PR assumes that function `SetReferrer` is never called in production so we don't add a feature flag to protect it.

enlight:
> This PR assumes that function `SetReferrer` has never called in production. We don't add a feature flag to protect it because the function can only be called by the Oracle. So we need to make sure that all the nodes are running the version that this change is in.

Lets add a feature flag so I have one less thing to worry about, `dpos:v3.5` I think.

builtin/plugins/dposv3/storage.go
```diff
@@ -448,15 +453,15 @@ func LoadCandidateList(ctx contract.StaticContext) (CandidateList, error) {
 
 func GetReferrer(ctx contract.StaticContext, name string) *types.Address {
 	var address types.Address
-	err := ctx.Get(append(referrersKey, name...), &address)
+	err := ctx.Get(referrerKey(name), &address)
```
enlight: This change needs to be behind a feature flag.

builtin/plugins/dposv3/storage.go
```diff
@@ -448,15 +453,15 @@ func LoadCandidateList(ctx contract.StaticContext) (CandidateList, error) {
...
 	if err != nil {
 		return nil
 	}
 	return &address
 }
 
 func SetReferrer(ctx contract.Context, name string, address *types.Address) error {
-	return ctx.Set(append(referrersKey, name...), address)
+	return ctx.Set(referrerKey(name), address)
```
enlight: And this change too needs to be behind a feature flag.

features.go
```diff
@@ -30,6 +30,10 @@ const (
 	// NOTE: The DPOS v3 contract must be loaded & deployed first!
 	DPOSVersion3_3 = "dpos:v3.3"
 
+	// Enables referrer functions
```
pathornteng: Enables ListReferrers method

builtin/plugins/dposv3/dpos.go
```diff
@@ -2001,7 +2025,10 @@ func (c *DPOS) RegisterReferrer(ctx contract.Context, req *RegisterReferrerReque
 		return logDposError(ctx, errOnlyOracle, req.String())
 	}
 
-	SetReferrer(ctx, req.Name, req.Address)
+	err = SetReferrer(ctx, req.Name, req.Address)
+	if err != nil {
+		return err
+	}
```
enlight:
```golang
if err := SetReferrer(ctx, req.Name, req.Address); err != nil {
```

builtin/plugins/dposv3/storage.go
```diff
@@ -456,6 +474,13 @@ func GetReferrer(ctx contract.StaticContext, name string) *types.Address {
 }
 
 func SetReferrer(ctx contract.Context, name string, address *types.Address) error {
+	if ctx.FeatureEnabled(loomchain.DPOSVersion3_5, false) {
+		return ctx.Set(referrerKey(name), address)
+	}
+	return DeprecatedSetReferrer(ctx, name, address)
```
enlight: Since the error from `SetReferrer` was ignored in `RegisterReferrer` we'll have to keep ignoring any error until the feature flag is enabled:
```golang
DeprecatedSetReferrer(ctx, name, address)
return nil
```

PR #1328, opened 2019-07-10
Add Metrics for Contract Tx Limiter Middleware

suhasagg:
1)Map Loading Times
2)Contract Tx Middleware Execution Latency

### DIFF
--- a/throttle/contract_tx_limiter_middleware.go
+++ b/throttle/contract_tx_limiter_middleware.go
@@ -22,10 +22,10 @@ var (
 )
 
 var (
-	tierMapLoadLatency                   metrics.Histogram
-	contractTierMapLoadLatency           metrics.Histogram
-	contractStatsMapLoadLatency          metrics.Histogram
-	contractTxMiddlewareExecutionLatency metrics.Histogram
+	tierMapLoadLatency                          metrics.Histogram
+	contractTierMapLoadLatency                  metrics.Histogram
+	contractStatsMapLoadLatency                 metrics.Histogram
+	contractTxLimiterMiddlewareExecutionLatency metrics.Histogram
 )
 
 func init() {
@@ -55,7 +55,7 @@ func init() {
 		Objectives: map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
 	}, fieldKeys)
 
-	contractTxMiddlewareExecutionLatency = kitprometheus.NewSummaryFrom(stdprometheus.SummaryOpts{
+	contractTxLimiterMiddlewareExecutionLatency = kitprometheus.NewSummaryFrom(stdprometheus.SummaryOpts{
 		Namespace:  "loomchain",
 		Subsystem:  "contract_tx_limiter_middleware",
 		Name:       "contract_tx_middleware_execution_latency_microseconds",
@@ -152,7 +152,7 @@ func NewContractTxLimiterMiddleware(cfg *ContractTxLimiterConfig,
 	) (res loomchain.TxHandlerResult, err error) {
 		defer func(begin time.Time) {
 			lvs := []string{"method", "NewContractTxLimiterMiddleware", "error", fmt.Sprint(err != nil)}
-			contractTxMiddlewareExecutionLatency.With(lvs...).Observe(time.Since(begin).Seconds())
+			contractTxLimiterMiddlewareExecutionLatency.With(lvs...).Observe(time.Since(begin).Seconds())
 		}(time.Now())
 
 		if !isCheckTx {
--- a/throttle/contract_tx_limiter_middleware.go
+++ b/throttle/contract_tx_limiter_middleware.go
@@ -59,7 +59,7 @@ func init() {
 		Namespace:  "loomchain",
 		Subsystem:  "contract_tx_limiter_middleware",
 		Name:       "contract_tx_middleware_execution_latency_microseconds",
-		Help:       "Total time taken for Contract Tx Middleware to Execute in microseconds.",
+		Help:       "Total time taken for Contract Tx Limiter Middleware to Execute in microseconds.",
 		Objectives: map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
 	}, fieldKeys)
 
--- a/throttle/contract_tx_limiter_middleware.go
+++ b/throttle/contract_tx_limiter_middleware.go
@@ -1,8 +1,11 @@
 package throttle
 
 import (
+	"fmt"
 	"time"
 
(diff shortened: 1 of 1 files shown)


suhasagg:
Screenshot from 2019-07-10 15-14-44
Screenshot from 2019-07-10 15-15-28
Screenshot from 2019-07-10 15-16-04
Screenshot from 2019-07-10 15-16-42

suhasagg:
4 metrics are recorded - 

1)tierMapLoadLatency                  
2)contractTierMapLoadLatency          
3)contractStatsMapLoadLatency          
4)contractTxMiddlewareExecutionLatency

throttle/contract_tx_limiter_middleware.go
```diff
@@ -11,12 +14,57 @@ import (
...
+	contractTxMiddlewareExecutionLatency metrics.Histogram
+)
+
+func init() {
+	fieldKeys := []string{"method", "error"}
+
+	tierMapLoadLatency = kitprometheus.NewSummaryFrom(stdprometheus.SummaryOpts{
+		Namespace:  "loomchain",
+		Subsystem:  "contract_tx_limiter_middleware",
+		Name:       "tier_map_load_latency_microseconds",
```
enlight: Remove `_microseconds` from all the names, it's actually seconds, but the units shouldn't be part of the name anyway.

throttle/contract_tx_limiter_middleware.go
```diff
@@ -179,12 +237,19 @@ func NewContractTxLimiterMiddleware(cfg *ContractTxLimiterConfig,
...
 
 		if txl.isAccountLimitReached(contractAddr, state.Block().Height) {
 			return loomchain.TxHandlerResult{}, ErrTxLimitReached
 		}
-
+		start = time.Now()
 		txl.updateState(contractAddr, state.Block().Height)
+		contractStatsMapLoadDuration := time.Since(start)
+		lvs = []string{"method", "contract Stats Map Load", "error", fmt.Sprint(err != nil)}
+		contractStatsMapLoadLatency.With(lvs...).Observe(contractStatsMapLoadDuration.Seconds())
```
enlight: This isn't going to be a very useful metric, it doesn't load anything, just accesses what's already in memory, remove it.

throttle/contract_tx_limiter_middleware.go
```diff
@@ -102,6 +150,11 @@ func NewContractTxLimiterMiddleware(cfg *ContractTxLimiterConfig,
 		next loomchain.TxHandlerFunc,
 		isCheckTx bool,
 	) (res loomchain.TxHandlerResult, err error) {
+		defer func(begin time.Time) {
+			lvs := []string{"method", "NewContractTxLimiterMiddleware", "error", fmt.Sprint(err != nil)}
+			contractTxMiddlewareExecutionLatency.With(lvs...).Observe(time.Since(begin).Seconds())
+		}(time.Now())
```
enlight: This is going to include the time spent running the tx handler and other middleware, so not going to be very useful or tell us anything we don't know already from the other metrics, might as well remove it.

throttle/contract_tx_limiter_middleware.go
```diff
@@ -144,6 +198,9 @@ func NewContractTxLimiterMiddleware(cfg *ContractTxLimiterConfig,
 			}
 			txl.contractToTierMap = contractToTierMap
 			txl.contractDataLastUpdated = time.Now().Unix()
+			contractTierMapLoadDuration := time.Since(start)
+			lvs := []string{"method", "contractTierMap Load", "error", fmt.Sprint(err != nil)}
+			contractTierMapLoadLatency.With(lvs...).Observe(contractTierMapLoadDuration.Seconds())
```
enlight: Move the code in this if block into a separate function and instrument that function, otherwise the error won't be captured.

throttle/contract_tx_limiter_middleware.go
```diff
@@ -144,6 +198,9 @@ func NewContractTxLimiterMiddleware(cfg *ContractTxLimiterConfig,
 			}
 			txl.contractToTierMap = contractToTierMap
 			txl.contractDataLastUpdated = time.Now().Unix()
+			contractTierMapLoadDuration := time.Since(start)
+			lvs := []string{"method", "contractTierMap Load", "error", fmt.Sprint(err != nil)}
```
enlight: `contractTierMap Load` -> `GetTierMap`

throttle/contract_tx_limiter_middleware.go
```diff
@@ -179,12 +237,19 @@ func NewContractTxLimiterMiddleware(cfg *ContractTxLimiterConfig,
 			}
 			txl.tierMap[contractTierID] = tierInfo
 		}
+		tierMapLoadDuration := time.Since(start)
+		lvs := []string{"method", "tierMap Load", "error", fmt.Sprint(err != nil)}
+		tierMapLoadLatency.With(lvs...).Observe(tierMapLoadDuration.Seconds())
```
enlight: Don't think there's any point in having this metric either, it only loads one tier each time.

suhasagg: Get Tier Map does a range query with tier key prefix and loads all the tierId --> TierInfo binding.
Later part of the code fetches a specific TierId data, so I think we should exclude this part and keep loading time measured from above

enlight: Instrument the code where `GetTierMap` is called, the `GetTierInfo` latency can be ignored.

suhasagg:
tierMaploadLatency
Tier Map Load Latency 

contract_tier_map_latency
Contract Tier Map Load Latency

throttle/contract_tx_limiter_middleware.go
suhasagg: Done.

throttle/contract_tx_limiter_middleware.go
suhasagg: Done.

throttle/contract_tx_limiter_middleware.go
suhasagg: Done.

throttle/contract_tx_limiter_middleware.go
suhasagg: Done.

throttle/contract_tx_limiter_middleware.go
suhasagg: Done

throttle/contract_tx_limiter_middleware.go
suhasagg: Done.

Issue #1325, opened 2019-07-10
Extend e2e test harness to support running clusters with a mix of builds


enlight:
We need to be able to run the e2e tests with nodes that are on different builds, this should help us detect consensus breaking changes between production release builds. It should work something like this:

1. Given build numbers X & Y, where X < Y.
2. In each multi-node e2e test spin up 1 node on build Y, and the rest on build X.

Might need a better way to detect consensus failure, right now I don't think it's obvious one has occurred during a test, gotta dig through logs.

These multi-build e2e tests will be run in a separate Jenkins job where build numbers X & Y will be parameters. The job will run once a day and will set Y to match the last successful build, and X to match the current production build.

Sriep:
Added new environmental variables to jenkins.sh
```bash
export LOOMEXE_PATH="../loom"
export LOOMEXE_ALTPATH="../loom2"
export MIN_VALIDATORS=2
export MIN_ALT_VALIDATORS=2
export ALWAYS_APPHASH_CHECK="yes"
make test
```
LOOMEXE_PATH and LOOMEXE_ALTPATH are paths to too loom builds you want to test.
MIN_VALIDATORS and MIN_ALT_VALIDATORS how many nodes you want using each of the two builds. Will override the settings already set.
ALWAYS_APPHASH_CHECK if non-empty will do a `chckeapphash` at the end of every test.

enlight:
Resolved by https://github.com/loomnetwork/loomchain/pull/1333

PR #1333, opened 2019-07-11
Extend e2e test harness to support two builds

Sriep:

Addresses https://github.com/loomnetwork/loomchain/issues/1325


e2e/common/common.go
```diff
@@ -16,9 +17,19 @@ import (
 	"github.com/loomnetwork/loomchain/e2e/node"
 )
 
+const (
+	loomExeEv = "LOOMEXE_PATH1"
+	loomExe2Ev = "LOOMEXE_ALTPATH"
+	minValidatorsEv = "MIN_VALIDATORS"
+	minAltValidatorsEv = "MIN_ALT_VALIDATORS"
```
enlight: Not entirely sure what these min validator vars are for.

e2e/lib/config.go
```diff
@@ -41,6 +41,8 @@ type Config struct {
 	NodePrivKeyPathList     []string
 	NodeRPCAddressList      []string
 	NodeProxyAppAddressList []string
+
+	AlwasyApphashCheck      bool
```
enlight: This is supposed to control whether or not the app hash check happens at the end of the test or not, right? Then `CheckAppHashOnExit` might be a better name.

e2e/common/common.go
Sriep: Currently, the number of validators are set in the `??_test.go` file and passed into common.NewConfig or otherwise hardcoded. Typically 4 and 0 for the two builds. 
These two environmental variables allow the number of nodes in using each of the two build to be set from the calling script rather than hardcoded.

e2e/common/common.go
```diff
@@ -68,19 +75,29 @@ func NewConfig(
 	conf := lib.Config{
 		Name:        name,
 		BaseDir:     basedirAbs,
-		LoomPath:    loompathAbs,
+		LoomPath:    "error",
```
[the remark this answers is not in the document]
Sriep: Its a debug thing, I'll remove it; The value set in NewConfig should be unused.
Basically, there are two loom paths, so the value actually used is going to be the one set in the node. We still need Config to have a loomPath member because of the way ["text/template" works with lib.Config](https://github.com/loomnetwork/loomchain/blob/issue1325-e2e/e2e/engine/cmd.go#L56).

e2e/common/common.go
```diff
@@ -35,28 +46,67 @@ func NewConfig(
...
+
+	LoomPath2 := os.Getenv(loomExe2Ev)
+	if len(LoomPath2) == 0 {
+		LoomPath2 = DefatulLoomPath2
+	}
+
+	var err error
+	validatorsEv := os.Getenv(ValidatorsEv)
+	if len(validatorsEv) > 0 {
+		validators, err = strconv.Atoi(validatorsEv)
```
pathornteng: I don't think we should override the number of validators like this because it breaks some of the tests that rely on the number of validators (like chainconfig voting test). We should check if the number of validators in the test and the number of validators from env match before overriding it. 

For example, suppose `ValidatorsEv` is 2, and `AltValidatorsEv` is 2
Case1: a test requires 4 validators, it should works fine.
Case2: a test requires 1 validator, it should keep the number of validator to be 1.
Case3: a test requires 6 validators, 4 of them will use loomExeEv and 2 of them will use loomExe2Ev

Then we should gradually fix our test not to run with a single validator.

e2e/engine/cmd.go
```diff
@@ -286,7 +254,60 @@ func (e *engineCmd) Run(ctx context.Context, eventC chan *node.Event) error {
 			}
 		}
 	}
+	if e.conf.CheckAppHashOnExit {
```
pathornteng: I think we should run checkapphash() every test command not just on exist, so that we can better detect the command that breaks consensus.

e2e/common/common.go
```diff
@@ -35,28 +46,90 @@ func NewConfig(
...
+
+	if validators == 1 {
+		return 1, 0, nil
+	}
+
+	if validators >= evValidators+evAltValidators {
+		return validators - evAltValidators, evAltValidators, nil
+	}
+
+	if validators <= evValidators+1 {
```
pathornteng: So with this logic if validators = 3, evValidator = 1, evAltValidator = 3, we will get 1, 1 which overrides the number of validators

e2e/cmd/new.go
```diff
@@ -1,134 +1,51 @@
...
-	var n, k int
-	var basedir, contractdir, loompath, name string
+	var validators, altValidators, k, numEthAccounts int
+	var basedir, contractdir, loompath, loompath2, name string
 	var logLevel, logDest string
 	var genesisFile, configFile string
 	var logAppDb bool
 	var force bool
-	var useFnConsensus bool
+	var useFnConsensus, checkAppHashOnExit bool
```
pathornteng: Should be just `checkAppHash`

e2e/cmd/new.go
```diff
@@ -1,134 +1,51 @@
...
-			conf.Accounts = accounts
-			if err := lib.WriteConfig(conf, "runner.toml"); err != nil {
-				return err
-			}
-
-			return nil
+			_, err := common.GenerateConfig(
+				name, "", genesisFile,configFile, basedir, contractdir, loompath, loompath2,
+				validators, altValidators, k, numEthAccounts,
+				useFnConsensus, force, checkAppHashOnExit,
```
pathornteng: ditto

e2e/cmd/new.go
```diff
@@ -1,134 +1,51 @@
...
+	flags.IntVarP(&k, "account", "k", 1, "Number of accounts to be created")
+	flags.IntVarP(&numEthAccounts, "numEthAccounts", "e", 1, "Number of ethereum accounts to be created")
 	flags.BoolVarP(&logAppDb, "log-app-db", "a", false, "Log the app state database usage")
 	flags.BoolVarP(&useFnConsensus, "fnconsensus", "", false, "Enable fnconsensus via the reactor")
 	flags.BoolVarP(&force, "force", "f", false, "Force to create new cluster")
 	flags.StringVar(&logLevel, "log-level", "debug", "Log level")
 	flags.StringVar(&logDest, "log-destination", "file://loom.log", "Log Destination")
 	flags.StringVarP(&genesisFile, "genesis-template", "g", "", "Path to genesis.json")
 	flags.StringVarP(&configFile, "config-template", "c", "", "Path to loom.yml")
+	flags.BoolVarP(&checkAppHashOnExit, "check-apphash", "p", false, "Check apphash on exit from test")
```
pathornteng: ditto

e2e/common/common.go
```diff
@@ -16,11 +17,21 @@ import (
 	"github.com/loomnetwork/loomchain/e2e/node"
 )
 
+const (
+	loomExeEv             = "LOOMEXE_PATH1"
+	loomExe2Ev            = "LOOMEXE_ALTPATH"
+	ValidatorsEv          = "VALIDATORS"
+	AltValidatorsEv       = "ALT_VALIDATORS"
+	checkAppHashAfterCmds = "CHECK_APPHASH_AFTERCMDS"
```
pathornteng: we should just change it to `checkAppHash`

e2e/common/common.go
```diff
@@ -35,28 +46,90 @@ func NewConfig(
 	validators, account, numEthAccounts int,
 	useFnConsensus bool,
 ) (*lib.Config, error) {
-	basedirAbs, err := filepath.Abs(path.Join(BaseDir, name))
+	checkAppHashAfterEV := os.Getenv(checkAppHashAfterCmds)
+	checkAppHashAfterCmds := len(checkAppHashAfterEV) > 0
```
pathornteng: ditto

jenkins.sh
```diff
@@ -51,6 +51,14 @@ export GORACE="log_path=`pwd`/racelog"
 #make loom-race
 #make test-race
 
+# LOOMEXE_PATH and LOOMEXE_ALTPATH are paths to the loom builds you want to test.
+# VALIDATORS and ALT_VALIDATORS how many nodes you want using each of the two builds. Will override the settings already set.
+# CHECK_APPHASH_ONEXIT if non-empty will do a chckeapphash at the end of every test.
```
pathornteng: Remove this if not used

e2e/common/common.go
```diff
@@ -66,21 +139,25 @@ func NewConfig(
...
-		LoomPath:    loompathAbs,
-		ContractDir: contractdirAbs,
-		TestFile:    testFileAbs,
-		Nodes:       make(map[string]*node.Node),
+		Name:                 name,
+		BaseDir:              basedirAbs,
+		ContractDir:          contractdirAbs,
+		TestFile:             testFileAbs,
+		Nodes:                make(map[string]*node.Node),
+		CheckAppHashAfterCmd: checkAppHashOnExit,
```
pathornteng: ditto

e2e/engine/cmd.go
```diff
@@ -285,8 +253,62 @@ func (e *engineCmd) Run(ctx context.Context, eventC chan *node.Event) error {
...
+			return err
+		}
+		newLastBlockHeight, err := strconv.ParseInt(info.Result.Response.LastBlockHeight, 10, 64)
+		if err != nil {
+			return err
+		}
+		if lastBlockHeight == 0 {
+			lastBlockHeight = newLastBlockHeight
+		}
+		if lastBlockHeight == newLastBlockHeight {
```
pathornteng: If some nodes are one block behind, app hash won't be included?

e2e/engine/cmd.go
```diff
@@ -285,8 +253,62 @@ func (e *engineCmd) Run(ctx context.Context, eventC chan *node.Event) error {
...
+			fmt.Printf("--> GET: %s, AppHash: %0xX\n", u, info.Result.Response.LastBlockAppHash)
+		}
 	}
 
+	// apphash should have only 1 entry
+	// this might not be true if network latency is hight
+	if len(apphash) != 1 {
+		return fmt.Errorf("Wrong Block.Header.AppHash")
+	}
 	return nil
```
pathornteng: Should this function check that all app hashes are the same before returning nil?

e2e/lib/config.go
```diff
@@ -41,6 +41,8 @@ type Config struct {
 	NodePrivKeyPathList     []string
 	NodeRPCAddressList      []string
 	NodeProxyAppAddressList []string
+
+	CheckAppHashAfterCmd bool
```
pathornteng: ditto

jenkins.sh
```diff
@@ -51,6 +51,14 @@ export GORACE="log_path=`pwd`/racelog"
...
+# export LOOMEXE_PATH="../loom"
+# export LOOMEXE_ALTPATH="../loom2"
+# export VALIDATORS=3
+# export ALT_VALIDATORS=3
+# export CHECK_APPHASH_AFTERCMDS="true"
```
pathornteng: We should change  it to `CHECK_APP_HASH`?

e2e/common/common.go
Sriep: if validators = 3, evValidator = 1, evAltValidator = 3 this check fails as !(3<= 1+1), but if it did not fail the return would be (3-1, 1) = (2,1). The actual values returned is from the last line`evValidators, validators - evValidators` ie (1, 3-1) = (1,2).
Adding the returned values should always produce validators e.g. `validators - 1 + 1 = validators`

e2e/engine/cmd.go
Sriep: apphash is a map, so `len(apphash) == 1` confirms there is at least one apphash and that they are all the same.

e2e/engine/cmd.go
Sriep: I guess when the code was originally written this was not considered a problem. I'll try adding a wait-loop here?

e2e/engine/cmd.go
```diff
@@ -47,96 +49,64 @@ type abciResponseInfo2 struct {
...
 			// check all  the nodes
 			if n.All {
 				for j, v := range e.conf.Nodes {
-					cmd, err := makeCmd(base, dir, *v)
+					cmd, err := getCommand(e.conf, *v, n)
 					if err != nil {
 						return err
 					}
-
+					//if strings.Contains(cmd.Path, "loom2") {
```
pathornteng: Remove the comment

e2e/common/common.go
```diff
@@ -107,7 +183,20 @@ func NewConfig(
...
+		nodes = append(nodes, n)
+	}
+
+	loompathAbs2, err := filepath.Abs(loompath2)
+	if err != nil {
+		return nil, err
+	}
+
+	for i := validators; i < validators+altValidators; i++ {
+		n := node.NewNode(int64(i), conf.BaseDir, loompathAbs2, conf.ContractDir, genesisTmpl, yamlFile)
```
pathornteng: We should print out what nodes are running `loom` and what nodes are running `loom2`

e2e/engine/cmd.go
```diff
@@ -285,8 +255,72 @@ func (e *engineCmd) Run(ctx context.Context, eventC chan *node.Event) error {
...
+		}
+		newLastBlockHeight, err := strconv.ParseInt(info.Result.Response.LastBlockHeight, 10, 64)
+		if err != nil {
+			return err
+		}
+		if lastBlockHeight == 0 {
+			lastBlockHeight = newLastBlockHeight
+		}
+
+		for i := 0; i < waitIntervals; i++ {
```
pathornteng: I don't think this is going to work. It is going to wait 100 second and nothing change. The easiest way is to only compare the app hashes of nodes that are on the same height only.

e2e/engine/cmd.go
```diff
@@ -285,8 +255,72 @@ func (e *engineCmd) Run(ctx context.Context, eventC chan *node.Event) error {
...
+			if lastBlockHeight == newLastBlockHeight {
+				break
+			}
+			time.Sleep(sleepInterval)
+		}
+		if lastBlockHeight == newLastBlockHeight {
+			apphash[string(info.Result.Response.LastBlockAppHash)] = struct{}{}
+			fmt.Printf("--> GET: %s, AppHash: %0xX\n", u, info.Result.Response.LastBlockAppHash)
```
pathornteng: Please also print height here.

e2e/common/common.go
```diff
@@ -8,19 +8,31 @@ import (
...
 
+	"github.com/pkg/errors"
+
 	"github.com/loomnetwork/loomchain/e2e/engine"
 	"github.com/loomnetwork/loomchain/e2e/lib"
 	"github.com/loomnetwork/loomchain/e2e/node"
 )
 
+const (
+	loomExeEv       = "LOOMEXE_PATH1"
```
enlight: Is it really meant to be `LOOMEXE_PATH1`? Or is it supposed to be `LOOMEXE_PATH`?

e2e/common/common.go
```diff
@@ -8,19 +8,31 @@ import (
...
+	loomExe2Ev      = "LOOMEXE_ALTPATH"
+	ValidatorsEv    = "VALIDATORS"
+	AltValidatorsEv = "ALT_VALIDATORS"
```
enlight: Both of these should start with a lower case letter since they don't need to be exported from this package.

e2e/common/common.go
```diff
@@ -8,19 +8,31 @@ import (
...
+	checkAppHash    = "CHECK_APP_HASH"
+)
+
 var (
 	// assume that this test runs in e2e directory
-	LoomPath    = "../loom"
-	ContractDir = "../contracts"
-	BaseDir     = "test-data"
+	DefaultLoomPath  = "../loom"
+	DefatulLoomPath2 = "../loom2"
```
enlight: These too should start with a lower-case letter.

e2e/common/common.go
```diff
@@ -35,28 +47,90 @@ func NewConfig(
 	validators, account, numEthAccounts int,
 	useFnConsensus bool,
 ) (*lib.Config, error) {
-	basedirAbs, err := filepath.Abs(path.Join(BaseDir, name))
+	checkAppHashEV := os.Getenv(checkAppHash)
+	checkAppHash := len(checkAppHashEV) > 0
+
+	LoomPath := os.Getenv(loomExeEv)
```
enlight: `LoomPath` -> `loomPath`

e2e/common/common.go
```diff
@@ -35,28 +47,90 @@ func NewConfig(
...
+	if len(LoomPath) == 0 {
+		LoomPath = DefaultLoomPath
+	}
+
+	LoomPath2 := os.Getenv(loomExe2Ev)
```
enlight: `LoomPath2` -> `loomParth2`

Sriep:
I am confused. Are you saying that we require Validators >= 3*AltValidators so that block height can increase in the face of failed apphash check. In which case we should add this as a constraint in [GenerateConfig](https://github.com/loomnetwork/loomchain/blob/issue1325-e2e/e2e/common/common.go#L91) when building the node array.

pathornteng:
> Are you saying that we require `Validators >= 3*AltValidators` so that block height can increase in the face of a failed apphash check. In which case we should add this as a constraint in [GenerateConfig](https://github.com/loomnetwork/loomchain/blob/issue1325-e2e/e2e/common/common.go#L91) when building the node array.

Currently, if there is consensus failure and the cluster doesn't have enough validator power to produce the block, `checkapphash` keeps querying old block so to fix this we have two options here.

1. Only run `checkapphash` on tests that have `Validators >= 3*AltValidators` 
2. Make `checkapphash` to be able to detect consensus failure and block height does not increase.

Either way, I think it is useless to run `checkapphash` on tests that run with single validator.

e2e/common/common.go
```diff
@@ -11,52 +11,106 @@ import (
...
+
+	return validators - 1, 1
+}
+
+func doCheckAppHash(checkAppHash bool, validators, altValidators uint64) bool {
+	if !checkAppHash || validators == 0 || altValidators == 0 {
+		return false
+	}
+	return float32(validators)/float32(altValidators) >= minRatioForAppHashCheck ||
+		float32(altValidators)/float32(validators) >= minRatioForAppHashCheck
```
enlight: Why convert to float32 here? Just use integers.

PR #1337, opened 2019-07-12
Add Binance Singed Tx To Auth Middleware

phoorichet:

This PR adds Binance signer to auth middleware so the client can use send txn using Binance private key (in secp256k1 format) with Bitcoin's style address.

### DIFF
--- a/cmd/loom/addressmapper.go
+++ b/cmd/loom/addressmapper.go
@@ -1,3 +1,5 @@
+// +build evm
+
 package main
 
 import (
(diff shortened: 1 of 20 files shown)


cmd/loom/addressmapper.go
```diff
@@ -56,14 +61,30 @@ func AddIdentityMappingCmd() *cobra.Command {
...
 			foreignAddr := loom.Address{ChainID: chainId, Local: foreignLocalAddr}
-			mapping.To = foreignAddr.MarshalPB()
-			mapping.Signature, err = address_mapper.SignIdentityMapping(user, foreignAddr, privkey)
+			sig, err := address_mapper.SignIdentityMapping(user, foreignAddr, privkey)
 			if err != nil {
 				return errors.Wrapf(err, "sigining mapping with %s key", chainId)
 			}
+			typedSig := append(make([]byte, 0, 66), byte(sigType))
+			// sig is already prefixed with default EIP721 signature type
+			sig = append(typedSig, sig[1:]...)
```
enlight: Instead of hacking the generated sig get rid of the call to `SignIdentityMapping`, and create the right hash here, then sign it with `GenerateTypedSig`.

phoorichet: Fixed

PR #1340, opened 2019-07-13
Sort output of List-features chainconfig go contract

suhasagg:

Sort output of List-features chainconfig go contract

### DIFF
--- a/builtin/plugins/chainconfig/chainconfig.go
+++ b/builtin/plugins/chainconfig/chainconfig.go
@@ -1,6 +1,8 @@
 package chainconfig
 
 import (
+	"sort"
+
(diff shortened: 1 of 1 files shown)


builtin/plugins/chainconfig/chainconfig.go
```diff
@@ -13,6 +13,7 @@ import (
 	"github.com/loomnetwork/loomchain/builtin/plugins/dposv3"
 	"github.com/loomnetwork/loomchain/registry"
 	"github.com/pkg/errors"
+	"sort"
```
enlight: You should be using `goimports` for formatting so it moves imports of built-in modules up top.

PR #1345, opened 2019-07-16
upgrade evm

ParthDesai:

In progress:
- [ ] ~Upgrade to newer version of geth~ (No new commits pertaining to Constantinople) [Comparison](https://github.com/loomnetwork/go-ethereum/compare/master...ethereum:master)
- [ ] Write tests for evm upgradation

### DIFF
--- a/evm/evm.go
+++ b/evm/evm.go
@@ -147,11 +147,7 @@ func NewEvm(sdb vm.StateDB, lstate loomchain.State, abm *evmAccountBalanceManage
 	p := new(Evm)
 	p.sdb = sdb
 
-	if lstate.FeatureEnabled(loomchain.EvmConstantinopleFeature, false) {
-		p.chainConfig = constantinopleChainConfig()
-	} else {
-		p.chainConfig = defaultChainConfig()
-	}
+	p.chainConfig = defaultChainConfig(lstate.FeatureEnabled(loomchain.EvmConstantinopleFeature, false))
 
 	p.vmConfig = defaultVmConfig(debug)
 	p.context = vm.Context{
@@ -276,11 +272,17 @@ func constantinopleChainConfig() params.ChainConfig {
 	}
 }
 
-func defaultChainConfig() params.ChainConfig {
+func defaultChainConfig(shouldConstantinopleFlagEnabled bool) params.ChainConfig {
 	cliqueCfg := params.CliqueConfig{
 		Period: 10,   // Number of seconds between blocks to enforce
 		Epoch:  1000, // Epoch length to reset votes and checkpoint
 	}
+
+	var constantinopleBlock *big.Int = nil
+	if shouldConstantinopleFlagEnabled {
+		constantinopleBlock = big.NewInt(0)
+	}
+
 	return params.ChainConfig{
 		ChainID:        big.NewInt(0), // Chain id identifies the current chain and is used for replay protection
 		HomesteadBlock: nil,           // Homestead switch block (nil = no fork, 0 = already homestead)
@@ -292,7 +294,7 @@ func defaultChainConfig() params.ChainConfig {
 		EIP155Block:         big.NewInt(0),                        // EIP155 HF block
 		EIP158Block:         big.NewInt(0),                        // EIP158 HF block
 		ByzantiumBlock:      big.NewInt(0),                        // Byzantium switch block (nil = no fork, 0 = already on byzantium)
-		ConstantinopleBlock: nil,                                  // Constantinople switch block (nil = no fork, 0 = already activated)
+		ConstantinopleBlock: constantinopleBlock,                  // Constantinople switch block (nil = no fork, 0 = already activated)
 		// Various consensus engines
 		Ethash: new(params.EthashConfig),
 		Clique: &cliqueCfg,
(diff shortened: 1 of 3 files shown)


evm/evm.go
```diff
@@ -247,6 +253,29 @@ func (e Evm) NewEnv(origin common.Address) *vm.EVM {
 	return vm.NewEVM(e.context, e.sdb, &e.chainConfig, e.vmConfig)
 }
 
+func constantinopleChainConfig() params.ChainConfig {
```
enlight: There's a one line difference between `defaultChainConfig` and `constantinopleChainConfig`, add a new parameter to `defaultChainConfig` to indicate which version of the EVM should be used so it's immediately obvious what the difference between the two versions is.

features.go
```diff
@@ -74,4 +74,7 @@ const (
 
 	// Enable option to allow checking the registry error
 	DeployTxVersion1_1Feature = "deploytx:v1.1"
+
+	// Enables constantinopole hard fork in evm interpreter
+	EvmConstantinopleFeature = "constantinopole:evm"
```
enlight: `constantinopole:evm` -> `evm:constantinople`

suhasagg:
Brief Summary upgrade to release.1.9 -->
In crypto == > sha3 module has been replaced.
It is breaking few modules in go-loom and loomchain.
evm compat in go-loom
Functions getting affected -->
soliditysha3
rlpencodewithSha3
and plasma cash module in loomchain.
Specific files -->
client/plasma_cash/transaction.go
ALso mamamerkle/sparse_merkle_tree.go:12 is getting affected.
In constantinople also sha3 module has been replaced.  

I am figuring out more modules which are getting affected

../go-loom/crypto/yubihsm.go:313:18: undefined: secp256k1.NormalizeLaxDERSignature
../go-loom/client/gateway/mainnet_gateway_ext.go:19:2: not enough arguments to return
../go-loom/client/gateway/mainnet_gateway_ext.go:19:19: t.contract.UnsignedTransact undefined (type *bind.BoundContract has no field or method UnsignedTransact)
# github.com/loomnetwork/loomchain/evm
evm/loomethdb.go:66:3: cannot use newBatch (type *batch) as type ethdb.Batch in return argument:
    *batch does not implement ethdb.Batch (missing Replay method)
evm/loomethdb.go:196:11: cannot use &b.batch (type *batch) as type ethdb.Batch in return argument:
    *batch does not implement ethdb.Batch (missing Replay method)
evm/loomethdb.go:207:2: cannot use b (type *LogBatch) as type ethdb.Batch in return argument:
    *LogBatch does not implement ethdb.Batch (missing Replay method)
evm/loomevm.go:55:7: cannot use NewLoomEthdb(loomState, logContext) (type *LoomEthdb) as type ethdb.Database in assignment:
    *LoomEthdb does not implement ethdb.Database (missing Ancient method)
evm/loomevm.go:91:15: levm.sdb.RawDump undefined (type StateDB has no field or method RawDump) 


There will be 2 changes which will be covered by loom fork of go-ethereum
These include -
1)go-loom/crypto/yubihsm.go:313:18: undefined: secp256k1.NormalizeLaxDERSignature
../go-loom/client/gateway/mainnet_gateway_ext.go:19:2: not enough arguments to return
2)/go-loom/client/gateway/mainnet_gateway_ext.go:19:19: t.contract.UnsignedTransact undefined (type *bind.BoundContract has no field or method UnsignedTransact)
3)sdb.RawDump
Rest of the changes will need to be done because go-ethereum release/1.9 has updated its modules


Some discussion on  sha3 removal from go-ethereum/crypto module - https://github.com/immutability-io/vault-ethereum/issues/51
Upstream go-ethereum removed github.com/ethereum/go-ethereum/crypto/sha3 which contained NewKeccak256. This function is now in golang.org/x/crypto/sha3 as NewLegacyKeccak256.



```
// AncientStore contains all the methods required to allow handling different
// ancient data stores backing immutable chain data store.
type AncientStore interface {
    AncientReader
    AncientWriter
    io.Closer
}
This is a new store that is added that has AncientReader/AncientWriter Interface

type BatchReplay interface {
   Put(key, value []byte)
   Delete(key []byte)
}
```
New interface dealing with batch operations

Here are few relevant PRs - https://github.com/ethereum/go-ethereum/pull/17814

Release notes - https://blog.ethereum.org/2019/07/10/geth-v1-9-0/

suhasagg:
Adding Ancient Store Compatibility and using geth release/1.9
For ancient store compatibility we can just add dummy interfaces of AncientStore in loomethdb.go.
This will do optimisation for state stored in (trie stored as key value pair in level db ) in geth and should reduce the strain on Go's GC, reduce the amount of memory used by Geth using ancient store
Relevant code in ==> (loomevm)
// Commit writes the state to the underlying in-memory trie database in geth.
root, err := levm.sdb.Commit(true)

if err := levm.sdb.Database().TrieDB().Commit(root, false); err != nil {
        return root, err
    }
Recent blocks, all state and accelerations structures are kept in a fast key-value store (LevelDB) as until now. This is meant to be run on top of an SSD as both disk IO performance is crucial.
Blocks and receipts that are older than a cutoff threshold (3 epochs) are moved out of LevelDB into a custom freezer database, that is backed by a handful of append-only flat files. Since the node rarely needs to read these data, and only ever appends to them, an HDD should be more than suitable to cover it.
Later for evmstore which stores evm state and for other stores, if we want to do Ancinet Store optimisation using immutable append-only flat files, we can do this optimisation.
2)I was also geting error for Raw Dump functionality during upgradation
    d := levm.sdb.RawDump()
    output, err := json.MarshalIndent(d, "", "  ")
    if err != nil {
        panic(err)
    }
fuctionality which takes evm dump, it is being utilised in evm DB CLI which is present in db/evm.go. For example in - newDumpEVMStateCommand()

enlight:
go-ethereum upgrade can go in another PR.

Issue #1347, opened 2019-07-16
JSON RPC eth_getTransactionReceipt improvement


eduardonunesp:
The command [eth_getTransactionReceipt](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_gettransactionreceipt) when called with a specific hash that doesn't exist, should return `null` on the result instead of an error, it should return errors only when really there's an error.

Current behavior 

```json
{
  "jsonrpc": "2.0",
  "id": 0,
  "error": {
    "code": -32000,
    "message": "Server error",
    "data": "loom error: Tx (D93C44702F5BDBDF9652FE82F6E94B7684EEA4227161F9C11F5139786C19ED1A) not found"
  }
}
```

The expected result of not found the receipt
```json
{
  "jsonrpc": "2.0",
  "id": 0,
  "result": null
}
```

enlight:
Resolved by https://github.com/loomnetwork/loomchain/pull/1349

PR #1349, opened 2019-07-16
Make eth_getTransactionReceipt to return null result when receipt isn't found

Sriep:

implements https://github.com/loomnetwork/loomchain/issues/1347

### DIFF
--- a/e2e/eth-1-test.toml
+++ b/e2e/eth-1-test.toml
@@ -27,3 +27,9 @@
         '"message": "Parse params"',
         '"data": "excess input arguments, expected 0 got 1"',
   ]
+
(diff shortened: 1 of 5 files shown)


Sriep:
The problem with e2e tests for `eth_getTransactionReceipt` and similar is that the transaction hashes are not known beforehand and our e2e testing tool cannot use return values from previous commands as input to future commands.

E2e test for `eth_getTransactionReceipt` were added to the [truffle tests ](https://github.com/loomnetwork/loomchain/blob/master/e2e/tests/truffle/test/EthFunctions.js). However, the truffle tests seem to have been disabled.

So the two solutions here would be either to activate the truffle tests or update e2e to remember previous commands results to be used as input to later commands.

PR #1350, opened 2019-07-17
Prevent negative value field in transaction object

Sriep:

Fix for https://github.com/loomnetwork/loomchain/issues/1361


pathornteng:
The value of `CallTx` cannot be negative as it is `BigUInt`. The checking is unnecessary. The problem is the conversion. We should fix these two lines https://github.com/loomnetwork/loomchain/blob/b582584f0b2360fda0c9c199fba4e3c7c795f576/eth/query/block.go#L184 
https://github.com/loomnetwork/loomchain/blob/b582584f0b2360fda0c9c199fba4e3c7c795f576/eth/query/block.go#L168 It should covert `BigUint` to `Uint64` instead of `Int64`

Sriep:
Fixed lines in block.go. BigUInt seems to be a wrapper around *big.Int, which can be negative so left the non-negatie check in for the moment.

vm/handler.go
```diff
@@ -60,6 +61,9 @@ func (h *DeployTxHandler) ProcessTx(
 	if tx.Value == nil {
 		value = loom.NewBigUIntFromInt(0)
 	} else {
+		if !common.IsPositive(tx.Value.Value) && !common.IsZero(tx.Value.Value) {
```
pathornteng: If we are going to check the negative value, a feature flag is needed to prevent consensus failure

e2e/eth-3-test.toml
```diff
@@ -4,7 +4,7 @@
 
 # Perform a transaction so can complete the first block
 [[TestCases]]
-  RunCmd = "{{ $.LoomPath }} deploy -b SimpleStore.bin -n SimpleStore -k {{index $.AccountPrivKeyPathList 0}}"
+  RunCmd = "{{ $.LoomPath }} deploy -b SimpleStore.bin -n SimpleStore  -v 1844674407370955161500000 -k {{index $.AccountPrivKeyPathList 0}}"
```
pathornteng: This test command fails with insufficient fund error. Should we fix it?

Sriep: No reason it needs to work.

pathornteng: ok

eduardonunesp:
LGTM

Issue #902, opened 2019-03-27
Add request validation to Coin Go contract


enlight:
Most of the methods in the coin contract don't validate the fields in the request before attempting to unmarshal them https://github.com/loomnetwork/loomchain/blob/master/builtin/plugins/coin/coin.go

So when some of those request fields are `nil` the node panics, and then we see an alert:
```
[...]

PR #1352, opened 2019-07-18
Coin contract need to validate some field in the request

atchapcyp:
Since some method in coin contract did't check the value before trying to unmarshal.
So It is possible to unmarshal `nil` variable.
To prevent this we should validate variable before unmarshal.

In order to do that we need ...
- Check the variable before unmarshal. 
- This change needs to be enabled by feature flag `coin:v1.2`
- Add coin contract unit test for flag `coin:v1.2` is enabled.

ref : #902 

### DIFF
--- a/builtin/plugins/coin/coin.go
+++ b/builtin/plugins/coin/coin.go
@@ -234,7 +234,6 @@ func (c *Coin) BalanceOf(
 	ctx contract.StaticContext,
 	req *BalanceOfRequest,
 ) (*BalanceOfResponse, error) {
-
 	if req.Owner == nil {
 		return nil, ErrInvalidRequest
 	}
@@ -346,7 +345,6 @@ func (c *Coin) Allowance(
 }
 
 func (c *Coin) TransferFrom(ctx contract.Context, req *TransferFromRequest) error {
-
 	if ctx.FeatureEnabled(loomchain.CoinVersion1_2Feature, false) {
 		if req.Amount == nil || req.From == nil || req.To == nil {
 			return ErrInvalidRequest
@@ -434,9 +432,6 @@ func loadAccount(
 }
 
 func saveAccount(ctx contract.Context, acct *Account) error {
-	if ctx.FeatureEnabled(loomchain.CoinVersion1_2Feature, false) && (acct.Owner == nil || acct.Balance == nil) {
-		return ErrInvalidRequest
-	}
 	owner := loom.UnmarshalAddressPB(acct.Owner)
 	return ctx.Set(accountKey(owner), acct)
 }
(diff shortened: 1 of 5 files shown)


builtin/plugins/coin/coin.go
```diff
@@ -318,6 +330,9 @@ func (c *Coin) Allowance(
 	ctx contract.StaticContext,
 	req *AllowanceRequest,
 ) (*AllowanceResponse, error) {
+	if ctx.FeatureEnabled(loomchain.CoinVersion1_2Feature, false) && (req.Spender == nil || req.Owner == nil) {
```
pathornteng: Like BalanceOf, Allowance does not change state of the chain, so the feature flag check is unnecessary. All methods that use `contract.StaticContext` are quite safe to be changed without a feature flag.

builtin/plugins/coin/coin.go
```diff
@@ -233,6 +234,10 @@ func (c *Coin) BalanceOf(
 	ctx contract.StaticContext,
 	req *BalanceOfRequest,
 ) (*BalanceOfResponse, error) {
+
```
enlight: Get rid of the blank line

builtin/plugins/coin/coin.go
```diff
@@ -332,6 +346,12 @@ func (c *Coin) Allowance(
 }
 
 func (c *Coin) TransferFrom(ctx contract.Context, req *TransferFromRequest) error {
+
```
enlight: Get rid of the blank line

builtin/plugins/coin/coin.go
```diff
@@ -417,6 +434,9 @@ func loadAccount(
 }
 
 func saveAccount(ctx contract.Context, acct *Account) error {
+	if ctx.FeatureEnabled(loomchain.CoinVersion1_2Feature, false) && (acct.Owner == nil || acct.Balance == nil) {
+		return ErrInvalidRequest
```
enlight: This isn't a contract method, so returning `ErrInvalidRequest` doesn't make sense in this context, you should move the check to the caller of `saveAccount`

builtin/plugins/coin/coin.go
```diff
@@ -441,6 +461,11 @@ func loadAllowance(
 }
 
 func saveAllowance(ctx contract.Context, allow *Allowance) error {
+	if ctx.FeatureEnabled(loomchain.CoinVersion1_2Feature, false) {
+		if allow.Owner == nil || allow.Spender == nil || allow.Amount == nil {
+			return ErrInvalidRequest
```
enlight: See comment on `saveAccount`, same thing applies here.

Issue #1341, opened 2019-07-15
Init DPOSv3 properly when initializing a new node


enlight:
Currently if you do:
```bash
mkdir new-node
cd new-node
loom init
loom run
```

The node will keep printing this every block:
```
I[2019-07-15|12:53:34.745] Event emitted                                module=loom index=1 length=240 msg="{\"topics\":[\"dposv3:election\"],\"caller\":{\"chain_id\":\"default\",\"local\":\"AAAAAAAAAAAAAAAAAAAAAAAAAAA=\"},\"address\":{\"chain_id\":\"default\",\"local\":\"AdEAKcJT+gLXYYi4S1hGqz0ZUQ0=\"},\"block_height\":\"1\",\"encoded_body\":\"CAE=\",\"block_time\":\"1563170009\"}"
ts=2019-07-15T05:53:35.7954838Z module=loom level=info _msg="Attempted to reward validator no longer on candidates list." validator="pub_key:\"\\315K\\350\\336\\252 }\\27729J\\231Y\\035\\340Y\\361\\254\\336\\300\\003\\230\\202\\035t\\375\\320d\\337\\363\\226j\" power:10 "
```

I'm guessing the proper way to fix this is to update DPOS.Init to register the validator address specified in the DPOS genesis as a candidate.

pathornteng:
Registering a candidate requires a certain amount of delegation. I think the better way to fix this is to change the log level from `Info` to `Debug`

enlight:
It should be possible to fully initialize a new cluster via the genesis file, changing the log level just hides the fact the cluster isn't configured properly. So I'll merge the log level change for now to get rid of the annoying log spam, but we still should fix this properly at some point, the DPOS genesis should allow you to specify the whitelist amount for each validator and then the genesis validators should be registered as candidates in DPOS.Init.

enlight:
On second thought the thing that was annoying me was that it was getting printed on every block, it's only supposed to be printed during an election, which means an election was running on every block, probably because `ElectionCycleLength` is zero - this should be fixed in the code that generates the default genesis.

enlight:
Resolved I guess.

PR #1357, opened 2019-07-19
Init DPOS candidates

pathornteng:
This PR adds logic to register all validators for candidates with minimum whitelist amount. 

Ref: https://github.com/loomnetwork/loomchain/issues/1341


builtin/plugins/dposv3/dpos.go
```diff
@@ -197,6 +198,40 @@ func (c *DPOS) Init(ctx contract.Context, req *InitRequest) error {
 		params.DowntimePeriod = defaultDowntimePeriod
 	}
 
+	candidates, err := LoadCandidateList(ctx)
```
enlight: Should just be `candidates := &CandidateList{}`

builtin/plugins/dposv3/dpos.go
```diff
@@ -197,6 +198,40 @@ func (c *DPOS) Init(ctx contract.Context, req *InitRequest) error {
...
+	if err != nil {
+		return err
+	}
+
+	// if InitCandidates is true, whitelist validators and register them for candidates
+	if params.InitCandidates {
```
enlight: `params` are persisted in the `State`, there's no reason for this to be in there though.

builtin/plugins/dposv3/dpos.go
```diff
@@ -197,6 +198,40 @@ func (c *DPOS) Init(ctx contract.Context, req *InitRequest) error {
...
+		for i, validator := range req.Validators {
+			candidateAddr := loom.Address{ChainID: "default", Local: loom.LocalAddressFromPublicKey(validator.PubKey)}
```
enlight: Hardcoded chain ID...

builtin/plugins/dposv3/dpos.go
pathornteng: Done

builtin/plugins/dposv3/dpos.go
pathornteng: Moved it to `DPOSInitRequest` instead

builtin/plugins/dposv3/dpos.go
pathornteng: Allow configurable chain ID

builtin/plugins/dposv3/dpos.go
```diff
@@ -197,6 +198,40 @@ func (c *DPOS) Init(ctx contract.Context, req *InitRequest) error {
 		params.DowntimePeriod = defaultDowntimePeriod
 	}
 
+	chainID := "default"
+	if req.ChainId != "" {
+		chainID = req.ChainId
+	}
```
enlight: Should be able to use `ctx.Block().ChainID` here, no?

pathornteng: Done, ChainID was empty in init block but not it is fixed.

Issue #1359, opened 2019-07-22
Investigate app/evm DB divergence after power failure


enlight:
One of the external PlasmaChain validators encountered this error after a power failure:
```
Jul 18 07:47:27 X[5806]: Error: EVM roots mismatch, evm.db(8076865): 6F753C94CA47298CE34B055F9A98217FA690618E4BC6161B1E65E6005F3E5254, app.db(8076973): 2494B5DD066018D3AAD48D0C8C966622D18F9093A387DBB5CDBDA1D805B2D7
```

We should investigate how it's possible to end up in this state, and what can be done to either prevent this from happening, or to auto-recover when it does happen.

pathornteng:
According to MultiWriterAppStore logic, we save EVM root to evm.db before app.db. So suppose the power failure happens right after evm.db is committed with changed EVM root at height 5, and app.db is still at height 4. When a node starts, MultiWriterAppStore will load EVM root at height 4 from evm.db root and the roots should match.

Is it possible that GolevelDB or Tendermint GolevelDB loses some field?

We can implement auto-recovery mechanism by making the app try to reload EVM root from evm.db at the exact height of app.db that has EVM root saved.

enlight:
From that error `evm.db` was at version `8076865`, and `app.db` was at version `8076973`, so `app.db` was 108 versions ahead.

pathornteng:
> From that error `evm.db` was at version `8076865`, and `app.db` was at version `8076973`, so `app.db` was 108 versions ahead.

Right, I misread the number. So it makes sense that the app.db height is higher than evm.db but it does not make sense that app.db has a EVM root saved at higher height since evm.db must be committed before app.db. Not sure how this can happen.

To fix this, we can have to check EVM roots before we load app.db and evm.db. Similar solution to https://github.com/loomnetwork/loomchain/issues/1335

PR #1362, opened 2019-07-24
Checking app/evm.db root on startup

atchapcyp:
Since the app.db evm root and evm.db root sometimes different on startup.
We should add the logic that check and change app.db version to evm.db version.

This PR will..

- Change `getEvmRoot` function for use outside it's package
- Check iavlstore and evmstore version before load `MultiWriterAppstore`

ref : https://github.com/loomnetwork/loomchain/issues/1359

### DIFF
--- a/cmd/loom/loom.go
+++ b/cmd/loom/loom.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bytes"
 	"encoding/base64"
 	"encoding/binary"
 	"encoding/json"
@@ -76,6 +77,7 @@ import (
 
 var (
 	appHeightKey = []byte("appheight")
+	rootKey      = []byte("vmroot")
 )
 
 var RootCmd = &cobra.Command{
@@ -598,14 +600,32 @@ func loadAppStore(cfg *config.Config, logger *loom.Logger, targetVersion int64) 
 		if err != nil {
 			return nil, err
 		}
+		iavlStoreRoot := iavlStore.Get(rootKey)
+
 		evmStore, err := loadEvmStore(cfg, iavlStore.Version())
 		if err != nil {
 			return nil, err
 		}
+		evmStoreRoot, evmVersion := evmStore.GetLastSavedRoot(iavlStore.Version())
+		fmt.Println("BEFORE EVM version : %d", evmVersion)
+		fmt.Println("BEFORE IAVL version : %d", iavlStore.Version())
+		if !bytes.Equal(iavlStoreRoot, evmStoreRoot) && evmVersion < iavlStore.Version() {
+			// downgrade iavlStore version
+			fmt.Println("EVM version : %d", evmVersion)
+			fmt.Println("IAVL version : %d", iavlStore.Version())
+			iavlStore, err = store.NewIAVLStore(db, cfg.AppStore.MaxVersions, evmVersion, cfg.AppStore.IAVLFlushInterval)
+			if err != nil {
+				return nil, err
+			}
+		}
+		fmt.Println("NEW EVM version : %d", evmVersion)
+		fmt.Println("NEW IAVL version : %d", iavlStore.Version())
+
 		appStore, err = store.NewMultiWriterAppStore(iavlStore, evmStore, cfg.AppStore.SaveEVMStateToIAVL)
 		if err != nil {
 			return nil, err
 		}
+
 	} else {
 		return nil, errors.New("Invalid AppStore.Version config setting")
 	}
--- a/store/multi_writer_app_store.go
+++ b/store/multi_writer_app_store.go
@@ -82,7 +82,7 @@ func NewMultiWriterAppStore(appStore *IAVLStore, evmStore *EvmStore, saveEVMStat
 			appStoreEvmRoot = defaultRoot
 		}
 	}
-	evmStoreEvmRoot, version := store.evmStore.getLastSavedRoot(store.appStore.Version())
+	evmStoreEvmRoot, version := store.evmStore.GetLastSavedRoot(store.appStore.Version())
 	if !bytes.Equal(appStoreEvmRoot, evmStoreEvmRoot) {
 		return nil, fmt.Errorf("EVM roots mismatch, evm.db(%d): %X, app.db(%d): %X",
 			version, evmStoreEvmRoot, appStore.Version(), appStoreEvmRoot)
--- a/store/evmstore.go
+++ b/store/evmstore.go
@@ -219,7 +219,7 @@ func (s *EvmStore) Commit(version int64) []byte {
 func (s *EvmStore) LoadVersion(targetVersion int64) error {
 	s.cache = make(map[string]cacheItem)
 	// find the last saved root
-	root, version := s.getLastSavedRoot(targetVersion)
+	root, version := s.GetLastSavedRoot(targetVersion)
 	if bytes.Equal(root, defaultRoot) {
 		root = []byte{}
 	}
@@ -240,7 +240,7 @@ func (s *EvmStore) Version() ([]byte, int64) {
 	return s.rootHash, s.version
 }
 
-func (s *EvmStore) getLastSavedRoot(targetVersion int64) ([]byte, int64) {
+func (s *EvmStore) GetLastSavedRoot(targetVersion int64) ([]byte, int64) {
 	start := util.PrefixKey(vmPrefix, evmRootPrefix)
 	end := prefixRangeEnd(evmRootKey(targetVersion))
 	iter := s.evmDB.ReverseIterator(start, end)
@@ -264,7 +264,7 @@ func (s *EvmStore) GetSnapshot(version int64) db.Snapshot {
 	if exist {
 		targetRoot = val.([]byte)
 	} else {
-		targetRoot, _ = s.getLastSavedRoot(version)
+		targetRoot, _ = s.GetLastSavedRoot(version)
 	}
 	return NewEvmStoreSnapshot(s.evmDB.GetSnapshot(), targetRoot)
 }


cmd/loom/loom.go
```diff
@@ -598,14 +600,32 @@ func loadAppStore(cfg *config.Config, logger *loom.Logger, targetVersion int64)
...
 		evmStore, err := loadEvmStore(cfg, iavlStore.Version())
 		if err != nil {
 			return nil, err
 		}
+		evmStoreRoot, evmVersion := evmStore.GetLastSavedRoot(iavlStore.Version())
+		fmt.Println("BEFORE EVM version : %d", evmVersion)
+		fmt.Println("BEFORE IAVL version : %d", iavlStore.Version())
+		if !bytes.Equal(iavlStoreRoot, evmStoreRoot) && evmVersion < iavlStore.Version() {
+			// downgrade iavlStore version
+			fmt.Println("EVM version : %d", evmVersion)
```
pathornteng: We should print out some log before trying to load lower version of app.db.
Change this to
```
log.Printf("EVM roots mismatch, evm.db(%d): %X, app.db(%d): %X", evmVersion, evmStoreEvmRoot, appStore.Version(), appStoreEvmRoot)
log.Printf("Try to load app.db at height %d", evmVersion)
```
```

File: cmd/loom/loom.go [hunk 2]
```diff
@@ -598,14 +600,32 @@ func loadAppStore(cfg *config.Config, logger *loom.Logger, targetVersion int64)
...
+		iavlStoreRoot := iavlStore.Get(rootKey)
+
 		evmStore, err := loadEvmStore(cfg, iavlStore.Version())
 		if err != nil {
 			return nil, err
 		}
+		evmStoreRoot, evmVersion := evmStore.GetLastSavedRoot(iavlStore.Version())
+		fmt.Println("BEFORE EVM version : %d", evmVersion)
+		fmt.Println("BEFORE IAVL version : %d", iavlStore.Version())
+		if !bytes.Equal(iavlStoreRoot, evmStoreRoot) && evmVersion < iavlStore.Version() {
```
pathornteng: `iavlStoreRoot` -> `appStoreEvmRoot`
`evmStoreRoot` -> `evmStoreEvmRoot`

cmd/loom/loom.go
```diff
@@ -598,14 +600,32 @@ func loadAppStore(cfg *config.Config, logger *loom.Logger, targetVersion int64)
...
 			return nil, err
 		}
+		iavlStoreRoot := iavlStore.Get(rootKey)
+
 		evmStore, err := loadEvmStore(cfg, iavlStore.Version())
 		if err != nil {
 			return nil, err
 		}
+		evmStoreRoot, evmVersion := evmStore.GetLastSavedRoot(iavlStore.Version())
+		fmt.Println("BEFORE EVM version : %d", evmVersion)
```
pathornteng: Remove all `fmt.Println`

cmd/loom/loom.go
```diff
@@ -598,14 +600,32 @@ func loadAppStore(cfg *config.Config, logger *loom.Logger, targetVersion int64)
...
+		if !bytes.Equal(iavlStoreRoot, evmStoreRoot) && evmVersion < iavlStore.Version() {
+			// downgrade iavlStore version
+			fmt.Println("EVM version : %d", evmVersion)
+			fmt.Println("IAVL version : %d", iavlStore.Version())
+			iavlStore, err = store.NewIAVLStore(db, cfg.AppStore.MaxVersions, evmVersion, cfg.AppStore.IAVLFlushInterval)
+			if err != nil {
+				return nil, err
+			}
+		}
+		fmt.Println("NEW EVM version : %d", evmVersion)
```
pathornteng: Remove these debug logs

cmd/loom/loom.go
```diff
@@ -598,14 +600,29 @@ func loadAppStore(cfg *config.Config, logger *loom.Logger, targetVersion int64)
...
 		evmStore, err := loadEvmStore(cfg, iavlStore.Version())
 		if err != nil {
 			return nil, err
 		}
+
+		evmStoreEvmRoot, evmVersion := evmStore.GetLastSavedRoot(iavlStore.Version())
+		if !bytes.Equal(appStoreEvmRoot, evmStoreEvmRoot) && evmVersion < iavlStore.Version() {
+			// downgrade iavlStore version
+			log.Info(fmt.Sprintf("EVM roots mismatch, evm.db(%d): %X, app.db(%d): %X", evmVersion, evmStoreEvmRoot, iavlStore.Version(), appStoreEvmRoot))
+			log.Info(fmt.Sprintf("Try to load app.db at height %d", evmVersion))
```
enlight: `log.Info` accepts extra key/values, no need to `fmt.Sprintf`:
```golang
log.Info(
  "EVM root mismatch, resyncing roots...",
  "evm.db-ver", evmVersion, "app.db-ver", iavlStore.Version(),
  "evm.db-root", hex.encodeToString(evmStoreEvmRoot), "app.db-evm-root", hex.encodeToString(appStoreEvmRoot),
)
```

cmd/loom/loom.go
```diff
@@ -598,14 +600,29 @@ func loadAppStore(cfg *config.Config, logger *loom.Logger, targetVersion int64)
...
+			iavlStore, err = store.NewIAVLStore(db, cfg.AppStore.MaxVersions, evmVersion, cfg.AppStore.IAVLFlushInterval)
```
enlight: This isn't going to work reliably if `IAVLFlushInterval > 0` because the `evmVersion` may not have a corresponding IAVL tree version on disk.

atchapcyp:
If we're trying to load version 0 while version > 0 exists, It will automatically load the most recent version of `app.db`. 
So my solution is empty the `app.db` first then reload db again and It worked.

cmd/loom/loom.go
```diff
@@ -644,14 +647,38 @@ func loadAppStore(cfg *config.Config, logger *loom.Logger, targetVersion int64)
...
 		}
+		evmStoreEvmRoot, evmVersion := evmStore.GetLastSavedRoot(iavlStore.Version())
+		if !bytes.Equal(appStoreEvmRoot, evmStoreEvmRoot) {
+			// downgrade iavlStore version
+			log.Info(
+				"EVM root mismatch, resyncing roots...",
+				"evm.db-ver", evmVersion, "app.db-ver", iavlStore.Version(),
+				"evm.db-root", hex.EncodeToString(evmStoreEvmRoot), "app.db-evm-root", hex.EncodeToString(appStoreEvmRoot),
+			)
+			lastestVersion, _ := store.LastestIAVLStoreVersion(db, evmVersion)
```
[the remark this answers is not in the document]
atchapcyp: Normally, function `MutableTree.LoadVersion` itself will return the latest version available that's less than the target version if the target version is not available along with the error.
```golang
	if !(targetVersion == 0 || latestVersion == targetVersion) {
		return latestVersion, fmt.Errorf("wanted to load target %v but only found up to %v",
			targetVersion, latestVersion)
	}
```
for example : `latestVersion,errror = 10, "wanted to load target 25 but only found up to 10"` 

And my intension for these line of code 
```golang
if lastestVersion == 0 {
 lastestVersion = -1
 }
```
Is if the function `MutableTree.LoadVersion` find out that the latest version available is version 0, And we continue call `store.NewIAVLStore` function with target version = 0
it will automatically return `IAVLStore` with Newest version not the version 0 that we want.
This is what I do to solve the problem.
```golang
func NewIAVLStore(db dbm.DB, maxVersions, targetVersion, flushInterval int64) (*IAVLStore, error) {
	tree := iavl.NewMutableTree(db, 10000)
	if targetVersion != -1 {
		_, err := tree.LoadVersion(targetVersion)
		if err != nil {
			return nil, err
		}
	}
	// always keep at least 2 of the last versions
	if (maxVersions != 0) && (maxVersions < 2) {
		maxVersions = 2
	}

	return &IAVLStore{
		tree:          tree,
		maxVersions:   maxVersions,
		flushInterval: flushInterval,
	}, nil
}
```
I just skip `tree.LoadVersion` from loading latest version again.

Now I will try to move to `store/multi_writer_app_store.go` as you comment.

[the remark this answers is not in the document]
enlight: > Normally, function `MutableTree.LoadVersion` itself will return the latest version available that's less than the target version if the target version is not available along with the error.

If it returns an error you can't rely on the return value, otherwise you're relying on internal implementation details that could change in the next version of the package.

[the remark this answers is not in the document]
atchapcyp: got it.

cmd/loom/loom.go
```diff
@@ -648,10 +649,24 @@ func loadAppStore(cfg *config.Config, logger *loom.Logger, targetVersion int64)
 		if err != nil {
 			return nil, err
 		}
+		iavlStore, err = store.LoadMultiWriterAppStore(iavlStore, evmStore, db, cfg.AppStore.MaxVersions, cfg.AppStore.IAVLFlushInterval)
+		if err != nil {
+			return nil, err
+		}
+		_, evmVersion := evmStore.Version()
```
pathornteng: Move all loading logic to `LoadMultiWriterAppStore`. `LoadMultiWriterAppStore` should return `MultiWriterAppStore` not `IAVLStore`

cmd/loom/loom.go
```diff
@@ -648,10 +649,24 @@ func loadAppStore(cfg *config.Config, logger *loom.Logger, targetVersion int64)
...
+		if evmVersion != iavlStore.Version() { // reload EVMStore
+			evmStore.CloseDB()
+			evmStore, err = loadEvmStore(cfg, iavlStore.Version())
```
pathornteng: Move `loadEvmStore` to `store/evmstore.go`

cmd/loom/loom.go
```diff
@@ -648,10 +649,24 @@ func loadAppStore(cfg *config.Config, logger *loom.Logger, targetVersion int64)
...
+			if err != nil {
+				return nil, err
+			}
+		}
+
 		appStore, err = store.NewMultiWriterAppStore(iavlStore, evmStore, cfg.AppStore.SaveEVMStateToIAVL)
```
pathornteng: This line should be just
```
appStore, err = store.LoadMultiWriterAppStore(db, cfg.EvmStore, cfg.AppStore)
```

store/multi_writer_app_store.go
```diff
@@ -97,6 +98,80 @@ func NewMultiWriterAppStore(appStore *IAVLStore, evmStore *EvmStore, saveEVMStat
...
+		)
+		if err != nil {
+			return nil, err
+		}
+		evmStore, err = loadEvmStore(evmDB, iavlStore.Version(), evmstoreCfg.NumCachedRoots)
+		if err != nil {
+			return nil, err
+		}
+	}
+	store := &MultiWriterAppStore{
```
pathornteng: We should check EVM root one more time to make sure that EVM roots are the same. So instead of copy the logic from `NewMultiWriterAppStore`, we can do
```
return NewMultiWriterAppStore(iavlStore, evmStore, appstoreCfg.SaveEVMStateToIAVL)
```
so we can delete the logic below this line.

cmd/loom/loom.go
```diff
@@ -639,18 +640,12 @@ func loadAppStore(cfg *config.Config, logger *loom.Logger, targetVersion int64)
...
-		if err != nil {
-			return nil, err
-		}
-		evmStore, err := loadEvmStore(cfg, iavlStore.Version())
-		if err != nil {
-			return nil, err
-		}
-		appStore, err = store.NewMultiWriterAppStore(iavlStore, evmStore, cfg.AppStore.SaveEVMStateToIAVL)
+		appStore, err = store.LoadMultiWriterAppStore(cfg.AppStore, cfg.EvmStore, db, cfg.AppStore.MaxVersions,
+			cfg.Metrics.Database, cfg.RootPath())
```
enlight: Format this like so:
```golang
appStore, err = store.LoadMultiWriterAppStore(
    cfg.AppStore, cfg.EvmStore, db, cfg.AppStore.MaxVersions,
    cfg.Metrics.Database, cfg.RootPath(),
)
```
Also the fourth argument is supposed to be the `targetVersion`, doesn't make sense to pass in `cfg.AppStore.MaxVersions` there.

store/multi_writer_app_store.go
```diff
@@ -97,6 +98,67 @@ func NewMultiWriterAppStore(appStore *IAVLStore, evmStore *EvmStore, saveEVMStat
...
+		}
+		evmStore.CloseDB()
+		evmDB, err = db.LoadDB(
+			evmstoreCfg.DBBackend,
+			evmstoreCfg.DBName,
+			rootPath,
+			evmstoreCfg.CacheSizeMegs,
+			evmstoreCfg.WriteBufferMegs,
+			metricsDatabase,
+		)
```
enlight: Why do we need to close and reload the DB here?

atchapcyp: Previously, When I trying to reload `evmStore` with function `loadEvmStore` errors resource in used will occurs  so I need to close the `db` before reloading it again. 

```golang
func loadEvmStore(cfg *config.Config, targetVersion int64) (*store.EvmStore, error) {
	evmStoreCfg := cfg.EvmStore
	db, err := cdb.LoadDB(
		evmStoreCfg.DBBackend,
		evmStoreCfg.DBName,
		cfg.RootPath(),
		evmStoreCfg.CacheSizeMegs,
		evmStoreCfg.WriteBufferMegs,
		cfg.Metrics.Database,
	)
}
```

But right now when I move to `store/multi_writer_app_store.go` we only load `evmdb` once so it should be fine. I will remove it.

cmd/loom/loom.go
```diff
@@ -76,6 +76,7 @@ import (
 
 var (
 	appHeightKey = []byte("appheight")
+	rootKey      = []byte("vmroot")
```
enlight: I think you can remove this change now.

db/goleveldb.go
```diff
@@ -44,6 +44,7 @@ func LoadGoLevelDB(name, dir string, cacheSizeMeg int, bufferSizeMeg int, collec
 	}
 
 	if collectMetrics {
+		prometheus.Unregister(metrics.NewStatsCollector(fmt.Sprintf("goleveldb_%s", name), log.Default, db)) // unregister previous registered
```
enlight: If this was added because `Register` returned an error then we should probably fix the error handling to check for `AlreadyRegisteredError` being returned from `Register` and ignore it.

store/multi_writer_app_store.go
```diff
@@ -97,6 +98,55 @@ func NewMultiWriterAppStore(appStore *IAVLStore, evmStore *EvmStore, saveEVMStat
 	return store, nil
 }
 
+func LoadMultiWriterAppStore(appStoreCfg *AppStoreConfig, evmstoreCfg *EvmStoreConfig, appStoreDB db.DBWrapper,
+	targetVersion int64, metricsDatabase bool, rootPath string) (*MultiWriterAppStore, error) {
+	iavlStore, err := NewIAVLStore(appStoreDB, appStoreCfg.MaxVersions, targetVersion, appStoreCfg.IAVLFlushInterval)
```
enlight: `evmstoreCfg` -> `evmStoreCfg`

db/goleveldb.go
atchapcyp: Since we don't need to reload evm.db again this line is not necessary now.

PR #1370, opened 2019-07-25
Fix registry/registry.pb.go path

Gilg4mesh:
The path should be fix if follow the Building in README


pathornteng:
Please use `make proto` if you want to generate go Protobuf messages as we are using plugin `protoc-gen-gogo` so this change is not necessary. You can take a look at `Makefile` to understand how to generate Protobuf message. 

Feel free to let me know if you have any question.

Issue #1365, opened 2019-07-24
Investigate issue with JSON RPC transaction receipt on extdev network


eduardonunesp:
The command `get_transactionReceipt` on testnet `https://extdev-plasma-us1.dappchains.com/eth` for a particular transaction hash `0x1ae0c9dcf722c4414b27425490fa8a53875dff4125ec572d58dc08f8bd71dea6` is returning an error.

```bash
$ curl -X POST --data '{"id": 0, "jsonrpc":"2.0", "method": "eth_getTransactionReceipt", "params": ["0x1ae0c9dcf722c4414b27425490fa8a53875dff4125ec572d58dc08f8bd71dea6"]}' https://extdev-plasma-us1.dappchains.com/eth
{
  "jsonrpc": "2.0",
  "id": 0,
  "error": {
    "code": -32000,
    "message": "Server error",
    "data": "loom error: Transaction index 1 out of bounds for transactions in block 1"
  }
}
``` 

It expects to return the transaction receipt correctly like the transaction receipt below:

```bash
$ curl -X POST --data '{"id": 0, "jsonrpc":"2.0", "method": "eth_getTransactionReceipt", "params": ["0xfc9d1fa0bfd5940344563333ccad28c8a4751a5cced385b13b365a20fec45923"]}' https://extdev-plasma-us1.dappchains.com/eth
{
  "result": {
    "transactionHash": "0xfc9d1fa0bfd5940344563333ccad28c8a4751a5cced385b13b365a20fec45923",
    "transactionIndex": "0x0",
    "blockHash": "0xf69f2ab3b3d62e175d54c4fe527bb96e2c1b45a86a464690bf43487db8cc2929",
    "blockNumber": "0x70fd06",
    "from": "0x848e5e35cccaa5ddb1cf707ed195a391bdc95b71",
    "cumulativeGasUsed": "0x0",
    "gasUsed": "0x0",
    "contractAddress": "0x95c594d2a38c762e015e40a4a74b4ccd4383ebc4",
    "to": null,
    "logs": [],
    "logsBloom": "0x0",
    "status": "0x1"
  },
  "jsonrpc": "2.0",
  "id": 0
}
```

Sriep:
https://github.com/loomnetwork/loomchain/pull/1375 fixes issues

eduardonunesp:
Causing https://github.com/loomnetwork/loomchain/issues/1466

enlight:
[...]

PR #1375, opened 2019-07-26
TxReceipt: get tx index from blockstore not receipt.

Sriep:

Addresses https://github.com/loomnetwork/loomchain/issues/1365
This gets the transaction index from the TM-BlockStore rather than the receipt database. 

The tx Index should be the same in both, however, for some historic reason it a difference might occur. In this case, the value in the TM-blockstore is the correct one. This should take slightly longer but guaranteed always returning the correct tx index value.

### DIFF
--- a/rpc/query_server.go
+++ b/rpc/query_server.go
@@ -790,11 +790,8 @@ func (s *QueryServer) EthGetTransactionReceipt(hash eth.Data) (*eth.JsonTxReceip
 		return nil, err
 	}
 	txReceipt.TransactionIndex, err = s.getTransactionIndex(&txReceipt)
-	if int32(len(blockResult.Block.Data.Txs)) <= txReceipt.TransactionIndex {
-		return nil, errors.Errorf(
-			"Transaction index %v out of bounds for transactions in block %v",
-			txReceipt.TransactionIndex, len(blockResult.Block.Data.Txs),
-		)
+	if err != nil {
+		return nil, err
 	}
 	txResults, err := s.BlockStore.GetTxResult(blockResult.Block.Data.Txs[txReceipt.TransactionIndex].Hash())
 	if err != nil {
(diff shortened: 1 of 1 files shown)


rpc/query_server.go
```diff
@@ -788,11 +789,9 @@ func (s *QueryServer) EthGetTransactionReceipt(hash eth.Data) (*eth.JsonTxReceip
 	if err != nil {
 		return nil, err
 	}
-	if int32(len(blockResult.Block.Data.Txs)) <= txReceipt.TransactionIndex {
-		return nil, errors.Errorf(
-			"Transaction index %v out of bounds for transactions in block %v",
-			txReceipt.TransactionIndex, len(blockResult.Block.Data.Txs),
-		)
+	txReceipt.TransactionIndex, err = s.getTransactionIndex(&txReceipt)
```
enlight: What do you think about doing this only if `int32(len(blockResult.Block.Data.Txs)) <= txReceipt.TransactionIndex`?

Sriep: My worry is that the issue's due to the tx index being 1 based rather than zero-based. In which case, while you only get an error if the tx is the last in the block, your never going to return the right tx details.

enlight: The tx index is 1-based?

enlight:
Still not clear why the tx index on the receipts is wrong, but we've fixed a few related issues recently (`receipts:v3`, `receipts:v3.1`, `receipts:v3.2`) that will hopefully stop this kind of thing from happening again. As for the existing messed up receipts we've got https://github.com/loomnetwork/loomchain/issues/1445 that we'll be running on the testnets to fix the receipts with the bogus tx indices.

enlight:
There's a clusterkit command to fix the indices, so this is no longer needed.

PR #1381, opened 2019-07-30
[fnConsensus] Add some docs and minor refactoring for clarity

enlight:


ParthDesai:
Note: Partial reason for taking below approach, was that at first reactor was going to be incorporated in tendermint repo, so we decide to don't use protobuf, and don't trust the fn plug-in that much.

1. Q: Why are the hash & message copied here?
See note. Reactor doesn't trust fn, so to prevent fn mistakenly modify byte array, it was safer to pass a copy.

2. Q: What relation does currentNonce have to numberOfAgreeVotes?
This is to decide which node is going to submit a message to contract. Idea is, validators who has voted for the majority message, are ordered in the same order as validator set, and the `currentNonce % numberOfAgreeVotes` index validator is responsible to call the contract.

3. Q: If remote nonce is greater or equal to ours then we end up sending the remote voteset back
to the peer that sent it to us, why? Shouldn't we exclude that peer from the broadcast?
It is a bug. We should totally do this.

4. Why this particular byte/byte seq?
Content of byte could be anything. Reason behind adding separator is so that contents cannot be fabricated to create same serialization. 

5. Why isn't this a protobuf?
See Note. Also, I didn't implement amino codec, because, I only need to calculate this to create the hash. It's never going to be de-serialized. So, It seems more performant to just create a byte array and calculate signature bytes.

ParthDesai:
@enlight So, 

> // NOTE: f.state is still locked at this point, so until the broadcast is complete we won't be able
	// to receive any votesets from anyone else because both handleVoteSetChannelMessage and
	// handleMaj23VoteSetChannel must acquire the f.state lock before they can do anything of substance.

This is kind of trade off between optimising for performance and have a safeguard in place, so we don't have any locked mutexes, by end of the vote unit execution. 

I think on hindsight, what we could do is, do broadcast message in async fashion, which I think would bring best of both.

Issue #1342, opened 2019-07-15
Add more tests for app store with in-memory IAVL versions


enlight:
We need some tests of the multi-writer app store with `IAVLFlushInterval > 0`, multiple readers, and a single writer to simulate what will happen on non-validator nodes when they respond to queries and retain snapshots that keep multiple in-memory-only versions of the IAVL tree alive.

Sriep:
https://loom-network.slack.com/archives/G9TCVAV16/p1563869326016700

Sriep:
fix https://github.com/loomnetwork/loomchain/pull/1382

PR #1382, opened 2019-07-30
Unit test multi-writer app store

Sriep:

Addresses https://github.com/loomnetwork/loomchain/issues/1342

### DIFF
--- a/store/iavlstore_test.go
+++ b/store/iavlstore_test.go
@@ -114,7 +114,7 @@ func TestIavl(t *testing.T) {
 	require.True(t, os.IsNotExist(err))
 
 	blocks = nil
-	blocks = iavl.GenerateBlocksHashKeys(numBlocks, blockSize)
+	blocks = iavl.GenerateBlocksHashKeys(numBlocks, blockSize, nil)
 	tree = iavl.NewMutableTree(db.NewMemDB(), 0)
 	for _, program := range blocks {
 		require.NoError(t, program.Execute(tree))
@@ -135,13 +135,13 @@ func testNormal(t *testing.T) {
 	_, err = diskTree.Load()
 	require.NoError(t, err)
 	/*
-	for _, entry := range store.Range(nil) {
-		_, value := tree.Get(entry.Key)
-		require.Zero(t, bytes.Compare(value, entry.Value))
-		_, diskValue := diskTree.Get(entry.Key)
-		require.Zero(t, bytes.Compare(value, diskValue))
+		for _, entry := range store.Range(nil) {
+			_, value := tree.Get(entry.Key)
+			require.Zero(t, bytes.Compare(value, entry.Value))
+			_, diskValue := diskTree.Get(entry.Key)
+			require.Zero(t, bytes.Compare(value, diskValue))
 
-	}*/
+		}*/
 	tree.Iterate(func(key []byte, value []byte) bool {
 		require.Zero(t, bytes.Compare(value, store.Get(key)))
 		_, diskValue := diskTree.Get(key)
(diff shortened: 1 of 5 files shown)


Sriep:
Warning running test with `-race` produces lots of errors, due to
```go
					for _, data := range snapshot.Range(prefix) {
						temp = data.Key
						temp = data.Value
					}
```
which could potentally read data at the same time its being changed.

store/multi_writer_app_store_test.go
```diff
@@ -269,3 +278,82 @@ func mockMultiWriterStore(flushInterval int64) (*MultiWriterAppStore, error) {
...
+			}
+		}
+
+	}(testStore)
+
+	wg.Wait()
+
+	for _, data := range testStore.Range(prefix) {
+		value := controlStore.Get(util.PrefixKey(prefix, data.Key))
+		require.Zero(t, bytes.Compare(value, data.Value))
```
enlight: Use `bytes.Equal`

store/multi_writer_app_store_test.go
```diff
@@ -269,3 +278,82 @@ func mockMultiWriterStore(flushInterval int64) (*MultiWriterAppStore, error) {
...
+	}
+	for _, data := range controlStore.Range(prefix) {
+		value := testStore.Get(util.PrefixKey(prefix, data.Key))
+		require.Zero(t, bytes.Compare(value, data.Value))
```
enlight: Ditto

store/multi_writer_app_store_test.go
```diff
@@ -269,3 +278,82 @@ func mockMultiWriterStore(flushInterval int64) (*MultiWriterAppStore, error) {
...
+		for {
+			select {
+			case <-quit:
+				return
+			default:
+				{
+					snapshot := s.GetSnapshot()
+					for _, data := range snapshot.Range(prefix) {
+						temp = data.Key
+						temp = data.Value
```
enlight: What's the point of assigning to `temp` here?

store/multi_writer_app_store_test.go
```diff
@@ -269,3 +278,82 @@ func mockMultiWriterStore(flushInterval int64) (*MultiWriterAppStore, error) {
...
+		controlHashes = append(controlHashes, hash)
+		require.NoError(t, err)
+	}
+
+	testStore, err := mockMultiWriterStore(flushInterval)
+	require.NoError(t, err)
+
+	quit := make(chan bool)
+	var wg sync.WaitGroup
+	wg.Add(2)
```
enlight: Need one writer and multiple readers, like say 100 readers.

store/iavlstore_test.go
```diff
@@ -135,13 +135,13 @@ func testNormal(t *testing.T) {
...
-		_, diskValue := diskTree.Get(entry.Key)
-		require.Zero(t, bytes.Compare(value, diskValue))
+		for _, entry := range store.Range(nil) {
+			_, value := tree.Get(entry.Key)
+			require.Zero(t, bytes.Compare(value, entry.Value))
+			_, diskValue := diskTree.Get(entry.Key)
+			require.Zero(t, bytes.Compare(value, diskValue))
 
-	}*/
+		}*/
```
enlight: Can we get rid of this commented out code?

store/multi_writer_app_store_test.go
Sriep: In case we want to address `-race` issues. Until these are fixed setting these values will generate errors when `go test` is run with the `-race` option.
`temp` is global to make sure something actually happens here, and the compiler does not optimise everything away.

Issue #1372, opened 2019-07-26
Unexpected error messages in rpc websocket connections.


Sriep:
When running the unit tests `Multi Websocket JSON-RPC` and `Single Websocket JSON-RPC` 
```bash unnecessary error messages appear when closing WebSocket.
E[2019-07-26|09:41:24.992] websocket client read message error          module=query-server err="websocket: close 1006 (abnormal closure): unexpected EOF"
E[2019-07-26|09:41:24.992] error writing message to websocket           module=query-server err="io: read/write on closed pipe"
E[2019-07-26|09:41:24.992] websocket client read message error          module=query-server err="websocket: close 1006 (abnormal closure): unexpected EOF"
E[2019-07-26|09:41:24.992] error writing message to websocket           module=query-server err="io: read/write on closed pipe"
E[2019-07-26|09:41:24.993] websocket client read message error          module=query-server err="websocket: close 1006 (abnormal closure): unexpected EOF"
```

enlight:
Should be resolved by https://github.com/loomnetwork/loomchain/pull/1386

PR #1386, opened 2019-07-31
Clean up websocket logs

Sriep:

Remove unnecessary logging of errors when closing WebSockets.
Add mutex to mock query server to remove `-race` errors.
https://github.com/loomnetwork/loomchain/issues/1372


rpc/client.go
```diff
@@ -53,18 +53,28 @@ type Client struct {
 // reads from this goroutine.
 func (c *Client) readPump(funcMap map[string]eth.RPCFunc, logger log.TMLogger) {
 	defer func() {
+		if r := recover(); r != nil {
+			logger.Error("WebSocket read panicked", "err", r)
+		}
 		c.hub.unregister <- c
-		_ = c.conn.Close()
+		if err := c.conn.Close(); err != nil {
+			logger.Info("Error closing WebSocket read pump", "err", err)
```
enlight: Why is this `Info` instead of `Error`?

rpc/client.go
```diff
@@ -114,15 +127,19 @@ func (c *Client) writePump(logger log.TMLogger) {
 			}
 
 			if err := c.conn.WriteMessage(websocket.TextMessage, message); err != nil {
-				logger.Error("error writing message to websocket", "err", err)
+				if err.Error() != "io: read/write on closed pipe" {
```
enlight: Err, nope, not gonna do that. Why does it matter what kind of error it is? Just log it regardless.

rpc/client.go
```diff
@@ -114,15 +127,19 @@ func (c *Client) writePump(logger log.TMLogger) {
...
+				}
 				return
 			}
 
 			n := len(c.send)
 			for i := 0; i < n; i++ {
 				msg := <-c.send
 				if err := c.conn.WriteMessage(websocket.TextMessage, msg); err != nil {
-					logger.Error("error writing message to websocket", "err", err)
+					if err.Error() != "io: read/write on closed pipe" {
```
enlight: Same here.

rpc/client.go
Sriep: `io: read/write on closed pipe` error occurs whenever the client closes the WebSocket connection. So it's not really an error so much as being informed that the dialogue is over.
Perhaps give an Info message, rather than an error, to record when the WebSocket connect closed?

rpc/client.go
Sriep: `io: read/write on closed pipe` error occurs whenever the client closes the WebSocket connection. So it's not really an error so much as being informed that the dialogue is over.
Perhaps give an Info message, rather than an error, to record when the WebSocket connect closed?

rpc/client.go
enlight: You can check for `err != io.ErrClosedPipe` then. It doesn't seem like logging any of these socket errors as `logger.Error` is particularly useful, maybe `logger.Debug` instead... in case anyone ever needs to debug something. Keep using `logger.Error` for the panics though, those shouldn't be happening often I presume.

Sriep: Two types of error.
1)  Errors that occur repeatedly during normal operation. (e.g. closed socket and EOF)
2)  Errors that normally never happen.
So report 1) as debug logs and 2) as error logs.

rpc/client.go
```diff
@@ -53,18 +54,30 @@ type Client struct {
...
-			websocket.IsUnexpectedCloseError(err, websocket.CloseGoingAway, websocket.CloseAbnormalClosure)
-			break
+			if websocket.IsUnexpectedCloseError(
+				err,
+				websocket.CloseGoingAway,
+				websocket.CloseAbnormalClosure,
+			) {
+				logger.Error("WebSocket unexpected close error", "err", err)
+			} else {
+				logger.Debug("WebSocket close request", "message", err)
```
enlight: Which one are you trying to log here, `message` or `err`?

rpc/client.go
```diff
@@ -94,41 +107,52 @@ func (c *Client) writePump(logger log.TMLogger) {
...
 			if err := c.conn.SetWriteDeadline(time.Now().Add(writeWait)); err != nil {
 				logger.Error("error setting write deadline", "err", err)
 			}
 
 			if err := c.conn.WriteMessage(websocket.TextMessage, message); err != nil {
-				logger.Error("error writing message to websocket", "err", err)
+				if err != io.ErrClosedPipe {
+					logger.Error("error writing message to WebSocket", "err", err)
+				} else {
+					logger.Debug("WebSocket connection closed", "message", err)
```
enlight: Ditto

rpc/client.go
```diff
@@ -94,41 +107,52 @@ func (c *Client) writePump(logger log.TMLogger) {
...
 
 			n := len(c.send)
 			for i := 0; i < n; i++ {
 				msg := <-c.send
 				if err := c.conn.WriteMessage(websocket.TextMessage, msg); err != nil {
-					logger.Error("error writing message to websocket", "err", err)
+					if err != io.ErrClosedPipe {
+						logger.Error("error writing message to WebSocket", "err", err)
+					} else {
+						logger.Debug("WebSocket connection closed", "message", err)
```
enlight: Ditto

rpc/client.go
```diff
@@ -53,18 +54,30 @@ type Client struct {
...
+			if websocket.IsUnexpectedCloseError(
+				err,
+				websocket.CloseGoingAway,
+				websocket.CloseAbnormalClosure,
+			) {
+				logger.Error("WebSocket unexpected close error", "err", err)
+			} else {
+				logger.Debug("WebSocket close request", "err", err)
+			}
+			return
```
enlight: There was a `break` here before, now there's a `return`, was this change intentional?

Sriep: No reason that I remember, both should do the same thing. I guess `return` is more explicit.

Issue #1240, opened 2019-06-18
Remove legacy receipt handlers & refactor receipt handling


enlight:
The legacy receipt handlers can be removed now that PlasmaChain has been migrated to the LevelDB based receipts handler. The receipts code is way too complicated right now, so removing the legacy handlers should allow us to simplify that code a fair bit.

enlight:
Noticed we're still hitting `app.db` whenever a tx receipt isn't found in `receipts_db`, need to get rid of that fall back. Migrate the old receipts over to `receipts_db`, we didn't do it initially for "reasons", but I want this code eliminated as part of the cleanup & refactor so we'll have to move this data (only need to do it on the PlasmaChain non-validator nodes).

enlight:
TODO:
- [ ] Remove legacy code that handled receipts before they were moved to `receipts_db` (https://github.com/loomnetwork/loomchain/pull/1387)
- [ ] Fix EVM tx hash generation so it's no longer the hash of the receipt.

PR #1387, opened 2019-08-01
Remove and refactor legacy receipt code 

pathornteng:
This PR refactors receipt code to make it cleaner. 

Ref: https://github.com/loomnetwork/loomchain/issues/1240


receipts/handler/handler.go
```diff
@@ -18,86 +17,45 @@ import (
...
-			return nil, errors.Wrap(err, "new leved db receipt handler")
-		}
-		rh.leveldbReceipts = leveldbHandler
-	}
-	return rh, nil
 }
 
 func (r *ReceiptHandler) Version() ReceiptHandlerVersion {
-	return r.v
+	return ReceiptHandlerLevelDb
```
enlight: Is this used anywhere?

pathornteng: Umm, look like it is an unused function, removed.

receipts/handler/handler.go
```diff
@@ -18,86 +17,41 @@ import (
...
-			receipt, chainErr = r.chainReceipts.GetReceipt(state, txHash)
-			if chainErr != nil {
-				return receipt, errors.Wrapf(common.ErrTxReceiptNotFound, "%v: %v", chainErr, err)
-			}
-		}
-		return receipt, nil
+func (r *ReceiptHandler) GetReceipt(txHash []byte) (types.EvmTxReceipt, error) {
+	receipt, err := r.leveldbReceipts.GetReceipt(txHash)
+	if err != nil {
+		return receipt, errors.Wrapf(common.ErrTxReceiptNotFound, "%v: %v", err, err)
```
enlight: Huh? What's with the `err, err`?

receipts/handler/handler.go
```diff
@@ -120,113 +74,71 @@ func (r *ReceiptHandler) GetCurrentReceipt() *types.EvmTxReceipt {
...
-		r.mutex.RUnlock()
-	default:
-		err = loomchain.ErrInvalidVersion
-	}
-
 	r.mutex.Lock()
+	defer r.mutex.Unlock()
+	if err := r.leveldbReceipts.CommitBlock(state, r.receiptsCache, uint64(height)); err != nil {
+		return err
+	}
```
enlight: This is different to how it worked before, before the hash list & receipts cache were reset regardless of whether or not `err == nil`. You think the old behavior was wrong?

pathornteng: No, the behavior should be the same. Fixed

receipts/handler/handler.go
pathornteng: Fixed

PR #1394, opened 2019-08-05
Timing Issue When Running "loom run" with Gateway Reactors Enabled

### LINKED ISSUE #1331: Timing Issue When Running "loom run" with Gateway Reactors Enabled
When running `loom run`, it calls `initQueryService` and `startGatewayReactors` respectively. The `initQueryService` function is non-blocking. When `startGatewayReactors` is run, it tries to connect to QueryServer. 

The issue happens when `startGatewayReactors` call `QueryServer` before `initQueryService`  finishes open the port and finalizes the service. Thus `loom run` returns error `connect: connection refused`. 

This issue is likely to happen when we're tring to run e2e test with multinode validators on a single host because sharing resources might causing the delay on `initQueryService`.

https://github.com/loomnetwork/loomchain/blob/ce05f2501eaa2caef910f3feafd45edb971b03a5/cmd/loom/loom.go#L412
https://github.com/loomnetwork/loomchain/blob/ce05f2501eaa2caef910f3feafd45edb971b03a5/cmd/loom/loom.go#L416

mukuls9971:

resolve #1331

### DIFF
--- a/cmd/loom/gateway_reactors.go
+++ b/cmd/loom/gateway_reactors.go
@@ -4,6 +4,8 @@ package main
 
 import (
 	"fmt"
+	"net/http"
+	"time"
 
 	glAuth "github.com/loomnetwork/go-loom/auth"
 	"github.com/loomnetwork/loomchain/config"
@@ -17,19 +19,49 @@ func startGatewayReactors(
 	cfg *config.Config,
 	nodeSigner glAuth.Signer,
 ) error {
+	if err := checkQueryService(cfg.TransferGateway.DAppChainReadURI, cfg.CheckQueryServiceTimeout); err != nil {
+		return err
+	}
 	if err := startGatewayFn(chainID, fnRegistry, cfg.TransferGateway, nodeSigner); err != nil {
 		return err
 	}
 
+	if err := checkQueryService(cfg.LoomCoinTransferGateway.DAppChainReadURI, cfg.CheckQueryServiceTimeout); err != nil {
+		return err
+	}
 	if err := startLoomCoinGatewayFn(chainID, fnRegistry, cfg.LoomCoinTransferGateway, nodeSigner); err != nil {
 		return err
 	}
 
+	if err := checkQueryService(cfg.TronTransferGateway.DAppChainReadURI, cfg.CheckQueryServiceTimeout); err != nil {
+		return err
+	}
 	if err := startTronGatewayFn(chainID, fnRegistry, cfg.TronTransferGateway, nodeSigner); err != nil {
 		return err
 	}
 
 	return nil
+}
+
+// checkQueryService is a helper function to check if corresponding QueryService has started
+// timeout is configurable with variable CheckQueryServiceTimeout
+func checkQueryService(url string, timeout int) error {
+	timeoutTick := time.After(time.Duration(timeout) * time.Millisecond)
+	tick := time.Tick(1000 * time.Millisecond)
+	for {
+		select {
+		case <-tick:
+			var netClient = &http.Client{
+				Timeout: time.Second * 1,
+			}
+			resp, _ := netClient.Head(url)
+			if resp != nil && resp.StatusCode == 200 {
+				return nil
+			}
+		case <-timeoutTick:
+			return fmt.Errorf("CheckQueryServer timeout occured")
+		}
+	}
 }
 
 func startGatewayFn(
--- a/cmd/loom/gateway_reactors.go
+++ b/cmd/loom/gateway_reactors.go
@@ -35,14 +35,15 @@ func startGatewayReactors(
 	return nil
 }
 
-// checkQueryService is a helper function to check if corresponding QueryService has started
-// timeout is configurable with variable QueryServicePollTimeout
+// Checks if the query server at the given URL is responding.
 func checkQueryService(url string) {
-	ticker := time.NewTicker(5000 * time.Millisecond)
+	ticker := time.NewTicker(5 * time.Second)
+	defer ticker.Stop()
+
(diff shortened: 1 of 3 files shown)


config/config.go
```diff
@@ -89,10 +89,11 @@ type Config struct {
...
 	// Transfer gateway
-	TransferGateway         *TransferGatewayConfig
-	LoomCoinTransferGateway *TransferGatewayConfig
-	TronTransferGateway     *TransferGatewayConfig
-	BinanceTransferGateway  *TransferGatewayConfig
+	TransferGateway          *TransferGatewayConfig
+	LoomCoinTransferGateway  *TransferGatewayConfig
+	TronTransferGateway      *TransferGatewayConfig
+	BinanceTransferGateway   *TransferGatewayConfig
+	CheckQueryServiceTimeout int
```
ParthDesai: Make it `QueryServicePollTimeout`

cmd/loom/gateway_reactors.go
```diff
@@ -17,21 +19,51 @@ func startGatewayReactors(
...
+		case <-tick:
+			var netClient = &http.Client{
+				Timeout: time.Second * 1,
+			}
+			resp, _ := netClient.Head(url)
+			if resp != nil && resp.StatusCode == 200 {
+				return nil
+			}
+		case <-timeoutTick:
+			return fmt.Errorf("CheckQueryServer timeout occured")
```
ParthDesai: Write something along the line of: 
```
Unable to connect to queryserver at <url>
```

Replace url with the url parameter passed to the function.

ParthDesai:
Also, Have you tested this? @mukuls9971

mukuls9971:
Gateway readURI is hardcoded(by default config) but proxyAppPort while starting up cluster with validators on local node(in e2e tests) is not hardcoded:
1. In current case, cfg.BatchSignFnConfig.Enabled = false, so nothing is detected
2. In my case, gateway is not able to resolve readURI and checkQueryService returns timeout error everytime 
Need to check further about other issues, please comment. Lets discuss this tomorrow. (spent a lot of time in drilling down the issue) @ParthDesai @enlight

cmd/loom/gateway_reactors.go
```diff
@@ -17,21 +19,51 @@ func startGatewayReactors(
 	cfg *config.Config,
 	nodeSigner glAuth.Signer,
 ) error {
+	if err := checkQueryService(cfg.TransferGateway.DAppChainReadURI, cfg.QueryServicePollTimeout); err != nil {
+		return err
+	}
```
enlight: You need to move this into `startGatewayFn`, and only ping the query server if the batch signing is actually enabled in the config.

phoorichet:
Immediate returns when getting 200 doesn't seem to solve the issue though. 

https://github.com/loomnetwork/loomchain/blob/a785bffd67433fd004dbffaf01ac80c17d3628af/cmd/loom/gateway_reactors.go#L53-L55

My guess is the query service might take a bit longer to be ready for service a request. I tried adding 2 seconds delay before the return, it run just fine on multi-node.

We might have to actually connect to the TransferGateway to make sure that the service is ready.

enlight:
> We might have to actually connect to the TransferGateway to make sure that the service is ready.

Yeah, resolving the gateway address by name is probably going to be the most reliable option.

cmd/loom/gateway_reactors.go
```diff
@@ -4,13 +4,25 @@ package main
...
 
 	glAuth "github.com/loomnetwork/go-loom/auth"
+	"github.com/loomnetwork/go-loom/client"
 	"github.com/loomnetwork/loomchain/config"
 	"github.com/loomnetwork/loomchain/fnConsensus"
+	"github.com/loomnetwork/loomchain/log"
 	tgateway "github.com/loomnetwork/transfer-gateway/gateway"
 )
 
+type ServiceName string
```
phoorichet: Maybe we don't actually need a new type here. Using just const without type should be enough.

cmd/loom/gateway_reactors.go
```diff
@@ -32,6 +44,23 @@ func startGatewayReactors(
...
 }
 
+// Checks if the query server at the given DAppChainReadURI is responding.
+func checkQueryService(name ServiceName, chainID string, DAppChainReadURI string, DAppChainWriteURI string) {
+	ticker := time.NewTicker(5 * time.Second)
+	defer ticker.Stop()
+
+	for ; true; <-ticker.C {
+		rpcClient := client.NewDAppChainRPCClient(chainID, DAppChainWriteURI, DAppChainReadURI)
+		gatewayAddr, err := rpcClient.Resolve(fmt.Sprintf("%s", name))
```
enlight: The `fmt.Sprintf` here is pointless, just pass in the `name`.

cmd/loom/gateway_reactors.go
```diff
@@ -32,6 +44,23 @@ func startGatewayReactors(
...
+		if err != nil {
+			log.Error("Error while resolving service", DAppChainReadURI, err)
```
enlight: -> `log.Error("Failed to resolve gateway contract", "gateway", name, "err", err)`

cmd/loom/gateway_reactors.go
```diff
@@ -32,6 +44,23 @@ func startGatewayReactors(
...
+		} else {
+			log.Info(fmt.Sprintf("Resolved %s : %#v", name, gatewayAddr))
```
enlight: Make this `log.Debug("Resolved gateway contract", "gateway", name, "addr", gatewayAddr)`

Issue #1396, opened 2019-08-06
Fix receipt not found error


pathornteng:
Currently, receipt not found error still persists in a few function calls. Our junior team is having a problem with EVM event scanner due to receipt not found error like

```
  August 6th 2019, 04:57:02.710    templeraid-scan    error: failed to get logs for TempleTimeUpdated: Response error: RPC error -32603 - Internal error: getting receipt: Tx receipt not found: get receipt for 1�-g�j*�J.�xQD��A�P@���Q�|: leveldb: not found: Tx receipt not found
```
and 
```
{ code: -32603,
  message: 'Internal error',
  data:
   'get receipt: Tx receipt not found: get receipt for ��-,�)u\t J&զ/��\u0004�}�!�\u0017�\'�F�Bnl]: leveldb: not found: Tx receipt not found' }
```
So when receipt is not found, query server should return nil receipt not error. We need to fix this problem in `GetBlockLogs` and `EvmTxReceipt` functions.

Note: currently, we don't store failed txs receipts in db so we probably need to review this issue once https://github.com/loomnetwork/loomchain/issues/1242 has been resolved.

enlight:
Which cluster is this happening on?

pathornteng:
Asia1

enlight:
Hacked around this in https://github.com/loomnetwork/loomchain/pull/1397

PR #1397, opened 2019-08-06
Fix receipt not found error

pathornteng:
Issue: https://github.com/loomnetwork/loomchain/issues/1396

### DIFF
--- a/eth/query/eth.go
+++ b/eth/query/eth.go
@@ -12,7 +12,6 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/gogo/protobuf/proto"
-	"github.com/loomnetwork/go-loom/plugin/types"
 	ptypes "github.com/loomnetwork/go-loom/plugin/types"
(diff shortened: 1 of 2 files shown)


pathornteng:
We still need to merge this PR. Even though `GetEvmTxReceipt` is deprecated but `GetBlockLogs` is still used in /eth

eth/query/eth.go
```diff
@@ -102,9 +104,13 @@ func GetBlockLogs(
 			var logsBlock []*ptypes.EthFilterLog
 			for _, txHash := range txHashList {
 				txReceipt, err := readReceipts.GetReceipt(state, txHash)
-				if err != nil {
+				receiptNotFound := errors.Cause(err) == common.ErrTxReceiptNotFound
```
enlight: Should just do
```golang
if errors.Cause(err) == common.ErrTxReceiptNotFound {
  continue
}
```

pathornteng: Done

Issue #1242, opened 2019-06-18
Store receipts for EVM txs that failed


enlight:
Currently we only store a receipt for successful EVM txs, but we should also store receipts for EVM txs that fail in DeliverTx. Now that we've migrated EVM receipts out of `app.db` to `receipts_db` we should change the receipt handler to also write receipts for failed txs.

enlight:
Ideally we clean out the old cruft first https://github.com/loomnetwork/loomchain/issues/1240

enlight:
Resolved by https://github.com/loomnetwork/loomchain/pull/1399

PR #1399, opened 2019-08-07
Store EVM receipt for txs that fail

pathornteng:
Ref issues: 
1. store failed EVM receipts https://github.com/loomnetwork/loomchain/issues/1242


app.go
```diff
@@ -614,9 +615,14 @@ func (a *Application) processTx(txBytes []byte, isCheckTx bool) (TxHandlerResult
 
 	r, err := a.TxHandler.ProcessTx(state, txBytes, isCheckTx)
 	if err != nil {
-		storeTx.Rollback()
-		// TODO: save receipt & hash of failed EVM tx to node-local persistent cache (not app state)
-		receiptHandler.DiscardCurrentReceipt()
+		evmRevertedError := strings.Contains(err.Error(), "evm: execution reverted")
```
enlight: Nah, this is too hacky to be relied on. Need a better way.

app.go
```diff
@@ -614,9 +615,14 @@ func (a *Application) processTx(txBytes []byte, isCheckTx bool) (TxHandlerResult
...
+		if evmRevertedError && state.FeatureEnabled(EvmTxReceiptsVersion2_1Feature, false) {
+			receiptHandler.CommitCurrentReceipt()
+			storeTx.Commit()
```
enlight: This is going to persist all state changes on error...

app.go
pathornteng: The error returned is https://github.com/ethereum/go-ethereum/blob/aa6005b469fdd1aa7a95f501ce87908011f43159/core/vm/instructions.go#L35. We can only compare string. Maybe to make it better I can do
```
err.Error() == "evm: execution reverted"
```

app.go
pathornteng: I have checked, only field that it changes is account nonce

enlight: Did you test an EVM call? What about a call to a Go contract that calls the EVM? Anyway, there's an expectation that all state changes are discarded when a tx errors out, need to be very surgical about violating that expectation. Also the nonce changes shouldn't be persisted in `CheckTx`.

app.go
enlight: Can't rely on error strings matching, what if it errors out for a different reason?

pathornteng:
If EVM txs fails, the application does not return txhash back to Tendermint, as a result, Tendermint creates its own tx hash. As a result, tx hashes on receipts and transaction are different. To fix this, the application has to return tx hash to Tendermint even if transactions fail.

app.go
```diff
@@ -607,9 +618,12 @@ func (a *Application) processTx(txBytes []byte, isCheckTx bool) (TxHandlerResult
 	receiptHandler := a.ReceiptHandlerProvider.Store()
 	r, err := a.TxHandler.ProcessTx(state, txBytes, isCheckTx)
 	if err != nil {
+		if !isCheckTx && state.FeatureEnabled(EvmTxReceiptsVersion2_1Feature, false) {
```
enlight: The receipts are saved to receipts_db and don't affect the app hash anymore, so the feature flag shouldn't be necessary right?

app.go
```diff
@@ -579,9 +579,20 @@ func (a *Application) DeliverTx(txBytes []byte) abci.ResponseDeliverTx {
...
+		a.curBlockHash,
+		a.GetValidatorSet,
+	)
+
 	r, err := a.processTx(txBytes, false)
 	if err != nil {
 		log.Error(fmt.Sprintf("DeliverTx: %s", err.Error()))
+		if state.FeatureEnabled(EvmTxReceiptsVersion2_1Feature, false) {
+			return abci.ResponseDeliverTx{Code: 1, Data: r.Data, Log: err.Error()}
+		}
```
enlight: Can you explain why this changed?

app.go
pathornteng: Right, we can just change it. Feature flag removed.

app.go
pathornteng: For success tx, the application returns receipt tx hash back to Tendermint, so the receipt tx hash is used as a transaction hash but for failed tx, the application does not return receipt tx hash back to Tendermint, so it creates its own tx hash. So the tx hash on the receipt and transaction tx hash are different. As a result, we cannot query receipt using Tendermint tx hash.

pathornteng: Receipt Tx hash is created here and returned to Tendermint through ResponseDeliverTx https://github.com/loomnetwork/loomchain/blob/445a4053cf53ac1582c7cb5dd5dbd688b2a048a0/receipts/leveldb/leveldb.go#L55

Sriep: Actually, you can query for a receipt using `Tendermint transaction hash` `(= types.Tx(txBytes).Hash())`. 
The receipts code is doubled up so you can get the receipt from both `tendermint` and `loom` tx hash. This was done so that you could get receipts from failed txs and the rtv from `eth_sendRawTransaction`. 

What we would probably want to do is replace the loom-tx hash with one of
* Use the tenderment tx hash throughout which would allow us to greatly simplify the tx-receipt code.
* Use  ethereum tx hash. Which would mean having to continue using two tx hashes and could keep the code much as it is.

pathornteng: @Sriep are you referencing to this PR https://github.com/loomnetwork/loomchain/pull/1401? Right now it seems that we can only query receipt using loom tx hash not tm tx hash.

app.go
```diff
@@ -680,8 +692,10 @@ func (a *Application) processTx(txBytes []byte, isCheckTx bool) (TxHandlerResult
 
 	r, err := a.TxHandler.ProcessTx(state, txBytes, isCheckTx)
 	if err != nil {
+		if !isCheckTx {
+			receiptHandler.CommitCurrentReceipt()
```
enlight: What did we just learn about changing the number of receipts that end up in the receipts cache?

evm/loomevm.go
```diff
@@ -164,8 +174,12 @@ func (lvm LoomVm) Create(caller loom.Address, code []byte, value *loom.BigUInt)
 			)
 		}
 
+		ethereumTxHash := types.NewContractCreation(
+			uint64(auth.Nonce(lvm.state, caller)), nil, lvm.gasLimit, big.NewInt(0), code,
```
enlight: The gas limit for the tx is supposed to be specified by the user that sends it, eventually the `/eth` interface will allow submitting Eth txs and we can take the gas limit from there... but for now we can just set it to `math.MaxUint64`.

app.go
pathornteng: Oops, not much apparently, fixed

evm/loomevm.go
pathornteng: Done

receipts/handler/handler.go
```diff
@@ -164,8 +166,14 @@ func (r *ReceiptHandler) CacheReceipt(
 		r.eventHandler, int32(len(r.receiptsCache)), int64(auth.Nonce(state, caller)),
 	)
 	if err != nil {
-		return []byte{}, errors.Wrap(err, "receipt not written, returning empty hash")
+		return nil, errors.Wrap(err, "receipt not written, returning empty hash")
+	}
+
+	// use go-ethereum tx hash instead of loom tx hash
```
enlight: `loom tx hash` -> `non-deterministic tx hash`
Pass the tx hash to `WriteReceipt()` and skip the old hash generation code in there if the `txHash != nil`

evm/loomevm.go
```diff
@@ -209,12 +216,18 @@ func (lvm LoomVm) Call(caller, addr loom.Address, input []byte, value *loom.BigU
 			)
 		}
 
+		ethereumTxHash := types.NewTransaction(
+			uint64(auth.Nonce(lvm.state, caller)), common.BytesToAddress(addr.Local),
+			nil, math.MaxUint64, big.NewInt(0), input,
```
enlight: Here too the `value` shouldn't be `nil`.

receipts/handler/handler.go
enlight: And move the feature check out to the caller of `CacheReceipt`

rpc/query_server.go
```diff
@@ -758,7 +759,6 @@ func (s *QueryServer) EthGetTransactionReceipt(hash eth.Data) (*eth.JsonTxReceip
 		}
 		return resp, nil
 	}
-	snapshot.Release()
```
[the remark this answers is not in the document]
pathornteng: It is redundant. `snapshot.Release()` has been called here https://github.com/loomnetwork/loomchain/blob/3e71b475020089761b7f07f9a5fac15c39fbea6f/rpc/query_server.go#L746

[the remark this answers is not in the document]
enlight: It's not redundant, it's explicitly released before we go rummaging through the block store so the snapshot isn't retained longer than necessary.

app.go
```diff
@@ -680,8 +692,10 @@ func (a *Application) processTx(txBytes []byte, isCheckTx bool) (TxHandlerResult
 
 	r, err := a.TxHandler.ProcessTx(state, txBytes, isCheckTx)
 	if err != nil {
+		if !isCheckTx && state.FeatureEnabled(features.EvmTxReceiptsVersion3_1, false) {
+			receiptHandler.CommitCurrentReceipt()
+		}
```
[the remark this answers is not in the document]
pathornteng: The current receipt is discarded after `processTx` is finished. If you think we should move this receipt code to `DeliverTx`, including all the logic in `if !isCheckTx {}` scope, I will move them all. Do you want it to be done in this PR or a separate one?

app.go
```diff
@@ -651,9 +651,21 @@ func (a *Application) DeliverTx(txBytes []byte) abci.ResponseDeliverTx {
 		deliverTxLatency.With(lvs...).Observe(time.Since(begin).Seconds())
 	}(time.Now())
 
+	state := NewStoreState(
+		context.Background(),
+		a.Store,
+		a.curBlockHeader,
+		a.curBlockHash,
+		a.GetValidatorSet,
+	)
```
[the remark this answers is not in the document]
pathornteng: Done

app.go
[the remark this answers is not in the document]
enlight: Just do it this PR.

receipts/handler/handler.go
pathornteng: Done

evm/loomevm.go
pathornteng: Done

rpc/query_server.go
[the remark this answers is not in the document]
pathornteng: In that case `defer snapshot.Release()` should be removed as it is pointless to call it as the snapshot will always be released before.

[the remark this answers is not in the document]
enlight: No, the `defer snapshot.Release()` is not pointless, it ensures the snapshot is released in case an error or panic causes the current function to return before the non-deferred `snapshot.Release()` is called.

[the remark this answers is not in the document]
pathornteng: True, fixed

app.go
```diff
@@ -678,43 +677,50 @@ func (a *Application) processTx(txBytes []byte, isCheckTx bool) (TxHandlerResult
...
-	r, err := a.TxHandler.ProcessTx(state, txBytes, isCheckTx)
+	r, err := a.TxHandler.ProcessTx(state, txBytes, false)
 	if err != nil {
+		log.Error(fmt.Sprintf("DeliverTx: %s", err.Error()))
 		storeTx.Rollback()
-		// TODO: save receipt & hash of failed EVM tx to node-local persistent cache (not app state)
-		return r, err
+		// Pass tx hash generated from go-etheruem back to Tendermint on failed EVM txs
+		if state.FeatureEnabled(features.EvmTxReceiptsVersion3_1, false) {
+			//receiptHandler.CommitCurrentReceipt()
```
enlight: Remove

app.go
```diff
@@ -678,43 +677,50 @@ func (a *Application) processTx(txBytes []byte, isCheckTx bool) (TxHandlerResult
...
 	defer receiptHandler.DiscardCurrentReceipt()
 
-	r, err := a.TxHandler.ProcessTx(state, txBytes, isCheckTx)
+	r, err := a.TxHandler.ProcessTx(state, txBytes, false)
 	if err != nil {
+		log.Error(fmt.Sprintf("DeliverTx: %s", err.Error()))
 		storeTx.Rollback()
-		// TODO: save receipt & hash of failed EVM tx to node-local persistent cache (not app state)
-		return r, err
+		// Pass tx hash generated from go-etheruem back to Tendermint on failed EVM txs
```
enlight: -> `Pass the EVM tx hash (if any) back to Tendermint so it stores it in block results`

app.go
```diff
@@ -678,43 +677,50 @@ func (a *Application) processTx(txBytes []byte, isCheckTx bool) (TxHandlerResult
...
+			receiptHandler.CommitCurrentReceipt()
 		}
-		storeTx.Commit()
 	}
-	return r, nil
+
+	storeTx.Commit()
+	if r.Info == utils.CallEVM || r.Info == utils.DeployEvm {
+		isEvmTx = true
+	}
```
enlight:
```golang
isEvmTx = r.Info == utils.CallEVM || r.Info == utils.DeployEvm
```

app.go
pathornteng: It should be uncommented. Done

app.go
pathornteng: Done

app.go
pathornteng: Done

pathornteng:
LGTM

app.go
```diff
@@ -720,6 +742,63 @@ func (a *Application) processTx(txBytes []byte, isCheckTx bool) (TxHandlerResult
...
+	r, txErr := a.TxHandler.ProcessTx(state, txBytes, false)
+
+	// Store the receipt even if the tx itself failed
+	var receiptTxHash []byte
+	if a.ReceiptHandlerProvider.Reader().GetCurrentReceipt() != nil {
+		receiptTxHash = a.ReceiptHandlerProvider.Reader().GetCurrentReceipt().TxHash
+		txHash := ttypes.Tx(txBytes).Hash()
+		// If a receipt was generated for an EVM tx add a link between the TM tx hash and the EVM tx hash
+		// so that we can use it to lookup relevant events using the TM tx hash.
+		if !bytes.Equal(txHash, receiptTxHash) {
```
pathornteng: We are creating an unnecessary link between TM Tx Hash and Receipt Tx Hash for EVM txs.
Receipt Tx hash is used as tx hash for EVM tx. So there is no need to create a link between TM tx has and TM Tx hash of a EVM transaction.

pathornteng: For GO tx calling EVM, this is still needed.

enlight: Yes the link is somewhat redundant for `TM(EVM)` txs, only really need it for `TM(GO->EVM)` txs, but this is meant to be an index. And for `TM(GO->EVM)` txs it means we don't have to hit `state.db` or `tx_index.db` to lookup the tx result to figure out what the EVM tx hash is. Ultimately we should probably just fix the Call/DeployTxHandler so they always store the tx type & EVM tx hash (if any EVM calls were made) in the `TxHandlerResult`.

pathornteng: Got it

Issue #1395, opened 2019-08-05
Logs lost from Solidity contract called from a Go-transaction


Sriep:
Emit an event from a Solidity contract called from a Go-contract. The events are lost.

Sriep:
Transaction of Go transaction that calls three EVM tx that emit three events.
1) Go -> Solidity -> emit(event)
2) Go -> Solidity -> Solidity -> emit(event)
3) Go -> Solidity -> emit(event)

```bash
curl -X POST --data '{"jsonrpc":"2.0","method":"eth_getBlockByNumber","params":["0x161",true],"id":83}' -X POST http://127.0.0.1:46658/eth
{
  "result": {
    "number": "0x161",
    "hash": "0xdd007004a52d1e58339c6f34e341f27a3ba92880cb04e6974418a81cbcd8ac79",
    "parentHash": "0x9b40d0f9c2082dc6ccfd1b3307e14be87bb15b6124b98c7e0f39a686ab0076d4",
    "nonce": "0x0000000000000000",
    "sha3Uncles": "0x0000000000000000000000000000000000000000000000000000000000000000",
    "logsBloom": "0x889a8293824424428205",
    "transactionsRoot": "0x0000000000000000000000000000000000000000000000000000000000000000",
    "stateRoot": "0x0000000000000000000000000000000000000000000000000000000000000000",
    "receiptsRoot": "0x0000000000000000000000000000000000000000000000000000000000000000",
    "miner": "0xee4964a38bb41959d1aa71a8fab90edd1aeb3569",
    "difficulty": "0x0",
    "totalDifficulty": "0x0",
    "extraData": "0x0",
    "size": "0x0",
    "gasLimit": "0x0",
    "gasUsed": "0x0",
    "timestamp": "0x5d4c47f7",
    "transactions": [
      {
        "hash": "0x48855495684591315caad2d3f4eb1242099236f1b534b10cc06cfebe77d1db22",
        "nonce": "0x3",
        "blockHash": "0xdd007004a52d1e58339c6f34e341f27a3ba92880cb04e6974418a81cbcd8ac79",
        "blockNumber": "0x161",
        "transactionIndex": "0x0",
        "from": "0x9a1ac42a17aad6dbc6d21c162989d0f701074044",
        "to": "0xde28fb974f31dfbe759cfcb3d1d44c2eefdfadd1",
        "value": "0x0",
        "gasPrice": "0x0",
        "gas": "0x0",
[...]

PR #1401, opened 2019-08-07
Logs for nested transactions

Sriep:

For https://github.com/loomnetwork/loomchain/issues/1395
Needs https://github.com/loomnetwork/go-loom/pull/440


app.go
```diff
@@ -605,6 +608,7 @@ func (a *Application) processTx(txBytes []byte, isCheckTx bool) (TxHandlerResult
 	)
 
 	receiptHandler := a.ReceiptHandlerProvider.Store()
+	receiptHandler.DiscardCurrentReceipt()
```
pathornteng: Why do we discard current receipt here?

config/loom.example.yaml
```diff
@@ -13,6 +13,7 @@ UseCheckTx: true
 RPCBindAddress: "tcp://0.0.0.0:46658"
 # PlasmaCashEnabled: false
 Oracle: "default:0xAfaA41C81A316111fB0A4644B7a276c26bEA2C9F"
+Testing: true
```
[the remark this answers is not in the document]
Sriep: changed to `SampleGoContract`. Added SampleGoContract to config.go for truffle test, so makes sense to check it works in the unit test.

cmd/loom/sample_go_contract.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/loomnetwork/go-loom/cli"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+const (
+	SampleGoContractName = "sample-go-contract"
+)
+
+func TestingNestedEvm() *cobra.Command {
```
enlight: Is this still needed?

e2e/tests/truffle/test/GoContracts.js
```diff
@@ -0,0 +1,88 @@
...
+
+        const goResult = await sampleGoContract.testNestedEvmCallsAsync(
+            new Address(client.chainId, LocalAddress.fromHexString(innerEmitter.address)),
+            new Address(client.chainId, LocalAddress.fromHexString(outerEmitter.address)),
+            innerEmitterValue,
+            outerEmitterValue,
+        );
+        const goContractLogs = await web3js.eth.getPastLogs({
+            address: innerEmitter.address,
+        });
```
enlight: So `innerEmitter.address` gets us logs for `outerEmitter` too? That seems a bit odd.
What do I get if I call `getPastLogs({ address: outerEmitter.address })`?
And what happens with `getPastLogs({ address: sampleGoContract.address })`?

Sriep: The addresses used in[ the logs are provided by the EVM](https://github.com/loomnetwork/loomchain/blob/nestedlogs/evm/loomevm.go#L163),  and not under our control. EVM knows nothing about Go-Contracts so will only return the account number for EVM-contracts.
eth_getLogs and similar search for logs using the data in them (EVM-account#, Height, Topics) so knows nothing about tx-hashes or G0-transactions.
I did not use `getPastLogs` to get the logs. I used `getTransactionReceipt
` using the transaction hash. It did not seem important where I got the logs as I was not testing the getPastLogs functionality, just making sure the right logs where genearted.
I did use `getPastLogs` to get the transaction hash of the go-contract call so that I could get the receipt with the logs. There is no easy way to get the transaction hash of go-contracts. It is not returned as the transaction hash, and the web3 functionality does not understand go-contracts. (Although a go-transaction's tx-hash is the tendermint tx-hash so can in principle be calculated - maybe make a js utility function for this?)

PR #1404, opened 2019-08-08
Add ability to auto-prune blockstore.db on startup

suhasagg:

### DIFF
--- a/cmd/loom/loom.go
+++ b/cmd/loom/loom.go
@@ -351,6 +351,18 @@ func newRunCommand() *cobra.Command {
 			if cfg.FnConsensus.Enabled {
 				fnRegistry = fnConsensus.NewInMemoryFnRegistry()
 			}
+			if cfg.BlockStore.PruneOnStartup {
+				fmt.Println("In prune on startup")
+				bs := store.NewPruningBlockStore("./chaindata", false, cfg)
+				if strings.EqualFold(cfg.BlockStore.PruningAlgorithm, "Copy") {
+					err = bs.PruneviaCopying("./chaindata", cfg.BlockStore.NumBlocksToRetain, cfg.BlockStore.PruneGraceFactor, cfg.BlockStore.SkipMissing)
+				} else {
+					err = bs.PruneviaDeletion("./chaindata", cfg.BlockStore.NumBlocksToRetain, cfg.BlockStore.PruneGraceFactor, 1, 10, cfg.BlockStore.SkipMissing, cfg.BlockStore.SkipCompaction)
+				}
+				if err != nil {
+					return err
+				}
+			}
 			var loaders []plugin.Loader
 			for _, loader := range cfg.ContractLoaders {
 				if strings.EqualFold("static", loader) {
@@ -364,6 +376,7 @@ func newRunCommand() *cobra.Command {
 				}
 			}
 			backend := initBackend(cfg, abciServerAddr, fnRegistry)
+			fmt.Println("backend")
 			loader := plugin.NewMultiLoader(loaders...)
 			termChan := make(chan os.Signal)
 			go func(c <-chan os.Signal, l plugin.Loader) {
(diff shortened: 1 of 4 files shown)


store/pruning_block_store.go
```diff
@@ -0,0 +1,160 @@
...
+type PruningBlockStore struct {
+	blockStoreDB       dbm.DB
+	prunedBlockStoreDB dbm.DB
+	*blockchain.BlockStore
+	cfg       *BlockStoreConfig
+	batchSize uint64
+}
+
+func NewPruningBlockStore(blockStoreConfig *BlockStoreConfig) *PruningBlockStore {
+	blockStoreDB := dbm.NewDB("blockstore", "leveldb", path.Join(blockStoreConfig.ChainDataPath, "data"))
```
enlight: The original `blockstore.db` should be opened in read-only mode.

store/pruning_block_store.go
```diff
@@ -0,0 +1,160 @@
...
+	// If minimum height is greater than target height, there are no blocks to prune, skip pruning
+	if oldestHeight >= targetHeight {
+		bs.blockStoreDB.Close()
+		bs.prunedBlockStoreDB.Close()
+		return nil
+	}
+	batch := bs.prunedBlockStoreDB.NewBatch()
+	numHeight := int64(0)
+	for height := targetHeight; height <= latestHeight; height++ {
+		log.Info("Copying block at height", "height", height)
```
enlight: This is going to be too spammy, should log the block range that will processed before the loop, and then the number of blocks actually processed at the end of the loop.

store/pruning_block_store.go
```diff
@@ -0,0 +1,160 @@
...
+	bs.blockStoreDB.Close()
+	bs.prunedBlockStoreDB.Close()
+}
+
+func (bs *PruningBlockStore) Height() int64 {
+	return blockchain.LoadBlockStoreStateJSON(bs.blockStoreDB).Height
+}
+
+func (bs *PruningBlockStore) Prune() error {
+	latestHeight := bs.Height()
```
enlight: We also need to ensure that `latestHeight` is less than the current `app.db` height, otherwise the node may not be able to replay all the blocks from the current app height up to `latestHeight`.

adyromantika:

enlight:
rebuild sesame

enlight:
rebuild sesame

cmd/loom/loom.go
```diff
@@ -406,6 +411,15 @@ func newRunCommand() *cobra.Command {
 			if err != nil {
 				return err
 			}
+
+			if cfg.BlockStore.PruneOnStartup {
+				if err = store.PruneBlockStore(
```
pathornteng: `NumBlocksToRetain` is not working properly since we always use `app.Store.Version()` as `minHeight`. With the current implementation, only the lastest block will be copied.

enlight: Ah, condition was inverted, fixed now.

store/block_store_pruner.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	srcDB.Close()
+	destDB.Close()
+
+	log.Info("[Block Store Pruner] Finished copying blocks", "count", numBlocksWritten)
+
+	if numBlocksWritten > 0 {
+		// Rename original blockstore to blockstore.db.bak{N}
+		backupPath := getBackupDBPath(srcDBPath)
```
pathornteng: I don't think we should keep the old blockstores. It's going to be useless and requires us to manually remove it later. We should just remove it here I think.

store/block_store_pruner.go
```diff
@@ -0,0 +1,168 @@
...
+	// If minimum height is greater than target height, there are no blocks to prune, skip pruning
+	if oldestHeight >= targetHeight {
+		return nil
+	}
+
+	log.Info("[Block Store Pruner] Copying blocks", "fromHeight", targetHeight, "toHeight", latestHeight)
+
+	batch := destDB.NewBatch()
+	numBlocksWritten := int64(0)
+	for height := targetHeight; height <= latestHeight; height++ {
```
pathornteng: FYI, This going to retain `NumBlocksToRetain` + 1 block but I think it is fine.

PR #1408, opened 2019-08-13
Remove mapped message from BatchSignWithdrawalFn

atchapcyp:
From the reference issue below, We need to deprecate `MapMessage` in `Fn` interface and move the way to collect message to `ReactorState` instead.
To do that I will 

- remove `MapMessage` from `Fn` interface
- add new `Message` variable type `map[string][]byte` to `ReactorState`

ref : https://github.com/loomnetwork/transfer-gateway/issues/70


fnConsensus/reactor.go
```diff
@@ -475,7 +455,7 @@ func (f *FnConsensusReactor) vote(fnID string, fn Fn, currentValidators *types.V
 	if aggregateExecutionResponse != nil {
 		f.safeSubmitMultiSignedMessage(
 			fn,
-			safeCopyBytes(aggregateExecutionResponse.Hash),
+			safeCopyBytes(f.state.Message[fnID]),
```
pathornteng: `stateMtx` is required to access hash map in state to prevent race condition. Also need to make sure that acquiring the lock will not cause deadlock

enlight: Mutex is already acquired at this point, so this looks fine.

fnConsensus/reactor.go
```diff
@@ -412,6 +401,8 @@ func (f *FnConsensusReactor) vote(fnID string, fn Fn, currentValidators *types.V
 		return
 	}
 
+	f.state.Message[fnID] = message
+
```
enlight: This needs to be moved down, after `f.stateMtx.Lock()` since you need to hold the lock before updating the state.

pathornteng: yeah, wrong line

fnConsensus/types.go
```diff
@@ -51,6 +51,7 @@ type ReactorState struct {
 	PreviousTimedOutVoteSets map[string]*FnVoteSet // TODO: unused, consider removing
 	PreviousMajVoteSets      map[string]*FnVoteSet
 	PreviousValidatorSet     *types.ValidatorSet
+	Message                  map[string][]byte
```
enlight: `Message` -> `Messages`
And lets store the hash with each message too, that way we can double-check the hash matches before calling `safeSubmitMultiSignedMessage`.

fnConsensus/types.go
```diff
@@ -51,7 +51,8 @@ type ReactorState struct {
 	PreviousTimedOutVoteSets map[string]*FnVoteSet // TODO: unused, consider removing
 	PreviousMajVoteSets      map[string]*FnVoteSet
 	PreviousValidatorSet     *types.ValidatorSet
-	Message                  map[string][]byte
+	Messages                 map[string][]byte
+	Hashes                   map[string][]byte
```
enlight: Like this:
```golang
type Message struct {
  Payload []byte // message bytes
  Hash []byte // hash of the payload
}

Messages map[string]Message
```

fnConsensus/reactor.go
```diff
@@ -452,9 +455,12 @@ func (f *FnConsensusReactor) vote(fnID string, fn Fn, currentValidators *types.V
 	// Have we achieved Maj23 already?
 	aggregateExecutionResponse := voteSet.MajResponse(f.cfg.FnVoteSigningThreshold, currentValidators)
 	if aggregateExecutionResponse != nil {
+		if !f.isMessageHashEqual(fnID, voteMethodID) {
```
enlight: Don't need to re-hash the message, just check that `bytes.Equal(f.state.Messages[fnID].Hash, response.Hash)` (need to make the change below first)

fnConsensus/reactor.go
```diff
@@ -473,9 +455,16 @@ func (f *FnConsensusReactor) vote(fnID string, fn Fn, currentValidators *types.V
 	// Have we achieved Maj23 already?
 	aggregateExecutionResponse := voteSet.MajResponse(f.cfg.FnVoteSigningThreshold, currentValidators)
 	if aggregateExecutionResponse != nil {
+		if !bytes.Equal(f.state.Messages[fnID].Hash, aggregateExecutionResponse.Hash) {
+			f.Logger.Error(
+				"FnConsensusReactor: message hash and response hash value are not the same",
+				"fnID", fnID, "method", voteMethodID,
```
enlight: Lets also log the `currentNonce`, and the `validatorIndex`

fnConsensus/reactor.go
```diff
@@ -598,10 +587,17 @@ func (f *FnConsensusReactor) commit(fnID string) {
 				// The consensus result only needs to be sent to the cluster by a single validator,
 				// that validator is chosen in a round-robin fashion every voting round.
 				if agreeVoteIndex != -1 && (currentNonce%int64(numberOfAgreeVotes)) == int64(agreeVoteIndex) {
+					if !bytes.Equal(f.state.Messages[fnID].Hash, majExecutionResponse.Hash) {
+						f.Logger.Error(
+							"FnConsensusReactor: message hash and hash value are not the same",
+							"fnID", fnID, "method", commitMethodID,
```
enlight: Lets also log the `currentNonce`, and the `validatorIndex`

Issue #1367, opened 2019-07-25
Add support for stricter output matching in e2e test harness


enlight:
Right now the `contains` and `excludes` conditions use basic string matching on the output of each command executed during an e2e test to check if the condition is satisfied. The current matching is too lenient, the order of the strings doesn't matter, the structure of the response doesn't matter. We need better & stricter pattern matching for command output so the tests can be less handwavy.

atchapcyp:
My suggested solution is
We can check JSON response value along with structure. 
In .toml file add the `ExpectedKey` and `ExpectedValue` key. 

```ExpectedKey = [“statistics[0].address.chainId”,“statistics[0].address.local”,“statistics[1].address.local”]```

```ExpectedValue = [“default”,"aY5Zh9a2YegP19SbjBZzVj+8/jo=,"I3jqGcbGtV6Qq4Jk7gUVhaTFd6s="]```
>Array index is only use to indicate that the result must be on the same group not the order of the result.
```json
{
  "statistics": [
    {
      "address": {
        "chainId": "default",
        "local": "aY5Zh9a2YegP19SbjBZzVj+8/jo="
      },
      "whitelistAmount": null,
      "locktimeTier": "TIER_ZERO",
      "delegationTotal": null,
      "slashPercentage": null,
      "recentlyMissedBlocks": "0",
      "updateWhitelistAmount": null,
      "updateLocktimeTier": "TIER_ZERO",
      "jailed": false
    },
    {
      "address": {
        "chainId": "default",
        "local": "I3jqGcbGtV6Qq4Jk7gUVhaTFd6s="
      },
      "whitelistAmount": null,
      "locktimeTier": "TIER_ZERO",
      "delegationTotal": null,
      "slashPercentage": null,
      "recentlyMissedBlocks": "0",
      "updateWhitelistAmount": null,
      "updateLocktimeTier": "TIER_ZERO",
      "jailed": false
    }
  ]
}
```

enlight:
Lets combine the keys & values if possible, so instead of having `ExpectedKey` & `ExpectedValue` we can just have:
```
Expected = [
  { “statistics[0].address.chainId”: “default” },
  { “statistics[0].address.local”: "aY5Zh9a2YegP19SbjBZzVj+8/jo=" },
  { “statistics[1].address”: "default:deadbeef" }
]
```

PR #1411, opened 2019-08-14
[WIP] Improve E2E test by using JSON searching

atchapcyp:
The end to end test is using basic string matching to check the result which unable to detect the structure of JSON response.
To check the JSON structure we can use `gjson` to search into the response by `JSON key` 
For example 
```json
{
  "jsonrpc": "2.0",
  "id": "",
  "result": {
    "block_height": "22",
    "validators": [
      {
        "address": "0EF70147374265B7D3B3C081EAAE18B365D885FA",
        "pub_key": {
          "type": "tendermint/PubKeyEd25519",
          "value": "6BmI6XQkIc4jYxWCfFTLYLgoWH5mmoVRjw+b8yAVREQ="
        },
        "voting_power": "1250000027746",
        "proposer_priority": "-1328140117919"
      },
      {
        "address": "8FE281BC26B3985479F6B6E78E9CC282D9F7BB69",
        "pub_key": {
          "type": "tendermint/PubKeyEd25519",
          "value": "st/xGw7dk4DdPfUySXqJtuyxgNwgGwEB64kbvssaFvY="
        },
        "voting_power": "1250010043600",
        "proposer_priority": "-1171869953427"
      }
    ]
  }
}
``` 
And then we use 
```golang
 gjson.Get(JsonOutput,result.validators.#.pub_key.value )
//[ "6BmI6XQkIc4jYxWCfFTLYLgoWH5mmoVRjw+b8yAVREQ=",
//"st/xGw7dk4DdPfUySXqJtuyxgNwgGwEB64kbvssaFvY=" ]
```
To get the value of `pub_key` and The result will be in term of array.

This is just my idea about how to improve the e2e test in case we what to check the structure of the response.

Ref : https://github.com/loomnetwork/loomchain/issues/1367
https://github.com/tidwall/gjson


e2e/engine/cmd.go
```diff
@@ -337,6 +340,43 @@ func checkConditions(e *engineCmd, n lib.TestCase, out []byte) error {
...
+					return err
+				}
+				buf := new(bytes.Buffer)
+				err = t.Execute(buf, e.conf)
+				if err != nil {
+					return err
+				}
+				if i%2 == 0 {
+					jsonKeys = append(jsonKeys, buf.String())
+				} else if i%2 == 1 {
```
enlight: Seeing as there are only two possibilities (either `i%2 == 0` or it's `1`) this should just be an `else`

e2e/engine/cmd.go
```diff
@@ -364,6 +404,63 @@ func checkConditions(e *engineCmd, n lib.TestCase, out []byte) error {
 	return nil
 }
 
+type testGroup struct {
+	Key        []string
+	Value      []string
+	BrakePoint string
+}
+
+func iter(isArray bool, i int, keys []string, v interface{}) (interface{}, error) {
```
enlight: What's this for?

e2e/dpos-2-validators.toml
```diff
@@ -4,6 +4,13 @@
   All = true
   Condition = "contains"
   Expected = ["{{index $.NodeBase64AddressList 0}}", "{{index $.NodeBase64AddressList 1}}"]
+  Json = true
+  ExpectedJson = [
+  "statistics.#.address.chainId", "default",
+  "statistics.#.address.local", "{{index $.NodeBase64AddressList 0}}",
+  "statistics.#.address.chainId", "default",
+  "statistics.#.address.local", "{{index $.NodeBase64AddressList 1}}"
```
enlight: Is there a way to match the first element & second element specifically? So we can make sure the first statistic matches node 0, and the second statistic matches node 1?

e2e/dpos-2-validators.toml
```diff
@@ -4,6 +4,13 @@
   All = true
   Condition = "contains"
   Expected = ["{{index $.NodeBase64AddressList 0}}", "{{index $.NodeBase64AddressList 1}}"]
+  Json = true
+  ExpectedJson = [
```
enlight: The `Json` field is redundant, if `ExpectedJson` is not empty match it, otherwise don't.

Makefile
```diff
@@ -218,6 +218,7 @@ deps: $(PLUGIN_DIR) $(GO_ETHEREUM_DIR) $(SSHA3_DIR)
 		github.com/prometheus/client_golang/prometheus \
 		github.com/go-kit/kit/log \
 		github.com/BurntSushi/toml \
+		github.com/tidwall/gjson \
```
enlight: This should go into `Gopkg.toml`, we only `go get` packages shared with `go-loom`.
You can use `dep ensure -add github.com/tidwall/gjson` to update `Gopkg.toml` & `Gopkg.lock.

e2e/dpos-2-validators.toml
atchapcyp: Yes, to do that we need to add some query to the key. 
I already add some example of usage.

e2e/dpos-2-validators.toml
```diff
@@ -28,11 +43,24 @@
...
+  "result.validators.#.pub_key.value","{{index $.NodePubKeyList 0}}"
+]
+
+[[TestCases]]
+  RunCmd = "check_validators"
+  Condition = "arithmetic"
+  KeysValues = [
+    "result.validators.#.voting_power",">","1250000000000",
+    "result.validators.#(pub_key.value=={{index $.NodePubKeyList 0}}).voting_power","<=","1260000000000"
+]
```
enlight: Can you combine `KeyValues` & `ExpectedJSON` into one thing, so it would look like this?
```
ExpectedJSON = [
    "statistics.#.address.local", "==", "{{index $.NodeBase64AddressList 1}}"
    "result.validators.#.voting_power",">","1250000000000"
]
```
Basically run the template evaluation on the right-hand operand, then evaluate the whole expression using `gval.Evaluate`.

PR #1414, opened 2019-08-14
Make Web3 JSON-RPC endpoint accept string IDs in requests

enlight:
The JSON-RPC can be a number, string, or could be missing entirely,
instead of attempting to decode it into an integer pass it through as is
in the response/error sent back to the client.


Sriep:
Tested with an integer, string and null ids; looks fine.

PR #1417, opened 2019-08-15
Print out more info to investigate "Karma coin test fails sporadically"

pathornteng:
Karma coin unit test fails sporadically. This PR prints out more info on the contract and request in order to find the root cause.


pathornteng:
The problem seems to be not easy to reproduce on one PR. I think we should merge this PR so that more useful info can be printed out when the problem recurs.

PR #1418, opened 2019-08-15
add TGGatewayVersion1_1 feature flag

marktrs:
a part of loomnetwork/transfer-gateway#72

### DIFF
--- a/features.go
+++ b/features.go
@@ -12,6 +12,8 @@ const (
 	TGFixERC721Feature = "tg:fix-erc721"
 	// Enables support for Binance contract mappings in the Binance Gateway contract
 	TGBinanceContractMappingFeature = "tg:binance-cm"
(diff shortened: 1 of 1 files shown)


features.go
```diff
@@ -12,6 +12,8 @@ const (
 	TGFixERC721Feature = "tg:fix-erc721"
 	// Enables support for Binance contract mappings in the Binance Gateway contract
 	TGBinanceContractMappingFeature = "tg:binance-cm"
+	// Store Mainnet Gateway address in Gateway Go contract
+	TGGatewayVersion1_1 = "tg:v1.1"
```
enlight: `TG` stands for Transfer Gateway, so this reads "Transfer Gateway Gateway Version 1.1", which is a bit redundant, better to call this `TGVersion1_1`

PR #1426, opened 2019-08-19
Add mainnet gateway updating function

marktrs:


cmd/loom/gateway/general_cmds.go
```diff
@@ -646,3 +650,52 @@ func newSetWithdrawFeeCommand() *cobra.Command {
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			loomKeyPath := gatewayCmdFlags.PrivKeyPath
+			hsmPath := gatewayCmdFlags.HSMConfigPath
+			algo := gatewayCmdFlags.Algo
+			signer, err := cli.GetSigner(loomKeyPath, hsmPath, algo)
+			if err != nil {
+				return err
+			}
+
+			mainnetAddress, err := hexToLoomAddress(args[0])
```
enlight: It's not a hex address for Binance is it?

marktrs: it can be converted to hex, but for convenient purpose we can check and convert if it was `bech32`

enlight: It should be in whatever encoding we use in the Binance Oracle config file, if the wallet address is hex then keep this hex, if it's bech32 then convert it here.

PR #1430, opened 2019-08-21
Configurable gas limit

pathornteng:
According to Ethereum, gas limit is 8,000,000.

Ref: https://ethstats.net/

### DIFF
--- a/evm/evm.go
+++ b/evm/evm.go
@@ -150,8 +150,8 @@ type Evm struct {
 func NewEvm(sdb vm.StateDB, lstate loomchain.State, abm *evmAccountBalanceManager, debug bool) *Evm {
 	p := new(Evm)
 	p.sdb = sdb
-	if lstate.Config().GetEvm() != nil {
-		p.gasLimit = lstate.Config().GetEvm().GasLimit
+	if lstate.FeatureEnabled(features.EvmGasLimit, false) {
+		p.gasLimit = lstate.Config().GetEvm().GasLimit()
 	} else {
 		p.gasLimit = defaultGasLimit
 	}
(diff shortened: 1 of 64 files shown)


evm/evm.go
```diff
@@ -144,12 +140,14 @@ type Evm struct {
...
 	validateTxValue bool
+	gasLimit        uint64
 }
 
 func NewEvm(sdb vm.StateDB, lstate loomchain.State, abm *evmAccountBalanceManager, debug bool) *Evm {
 	p := new(Evm)
 	p.sdb = sdb
-
+	onChainCfg := lstate.Config()
+	p.gasLimit = onChainCfg.Evm.GasLimit
```
enlight: Probably need to do `onChainCfg.GetEvm().GasLimit` because `onChainCfg.Evm` is likely to be `nil`.
And need a feature flag.

evm/evm.go
```diff
@@ -144,12 +140,13 @@ type Evm struct {
...
 	vmConfig        vm.Config
 	validateTxValue bool
+	gasLimit        uint64
 }
 
 func NewEvm(sdb vm.StateDB, lstate loomchain.State, abm *evmAccountBalanceManager, debug bool) *Evm {
 	p := new(Evm)
 	p.sdb = sdb
-
+	p.gasLimit = lstate.Config().GetEvm().GasLimit
```
enlight: Suppose you set `Config.Evm.GasLimit` to zero, then the protobuf unmarshaller will unmarshal `Config.Evm` to `nil`, and then I suspect `lstate.Config().GetEvm().GasLimit` will cause a panic due to `nil` being de-referenced... might be a good idea to write a small test.

PR #1431, opened 2019-08-22
Add tx handler

marktrs:
extract migration tx handler from #1224
go-loom : loomnetwork/go-loom#446

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ BINANCE_TGORACLE_DIR=$(GOPATH)/src/$(PKG_BINANCE_TGORACLE)
 # NOTE: To build on Jenkins using a custom go-loom branch update the `deps` target below to checkout
 #       that branch, you only need to update GO_LOOM_GIT_REV if you wish to lock the build to a
 #       specific commit.
-GO_LOOM_GIT_REV = HEAD
+GO_LOOM_GIT_REV = add-tx-handler
 # Specifies the loomnetwork/transfer-gateway branch/revision to use.
 TG_GIT_REV = HEAD
 # loomnetwork/go-ethereum loomchain branch
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ BINANCE_TGORACLE_DIR=$(GOPATH)/src/$(PKG_BINANCE_TGORACLE)
 # NOTE: To build on Jenkins using a custom go-loom branch update the `deps` target below to checkout
 #       that branch, you only need to update GO_LOOM_GIT_REV if you wish to lock the build to a
 #       specific commit.
-GO_LOOM_GIT_REV = add-tx-handler
+GO_LOOM_GIT_REV = HEAD
 # Specifies the loomnetwork/transfer-gateway branch/revision to use.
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -28,7 +28,7 @@ BINANCE_TGORACLE_DIR=$(GOPATH)/src/$(PKG_BINANCE_TGORACLE)
 # NOTE: To build on Jenkins using a custom go-loom branch update the `deps` target below to checkout
 #       that branch, you only need to update GO_LOOM_GIT_REV if you wish to lock the build to a
 #       specific commit.
-GO_LOOM_GIT_REV = HEAD
+GO_LOOM_GIT_REV = add-tx-handler
```
enlight: I've merged the go-loom PR, can set this back to HEAD

tx_handler/migration_tx_handler_test.go
```diff
@@ -29,9 +30,9 @@ func TestMigrationTxHandler(t *testing.T) {
 	migrationTx1 := mockMessageTx(t, uint32(1), origin, origin)
 
 	migrationFuncs := map[int32]MigrationFunc{
-		1: func(ctx *migrations.MigrationContext) error { return nil },
-		2: func(ctx *migrations.MigrationContext) error { return nil },
-		3: func(ctx *migrations.MigrationContext) error { return nil },
+		1: func(ctx *migrations.MigrationContext, parameters []byte) error { return nil },
+		2: func(ctx *migrations.MigrationContext, parameters []byte) error { return nil },
+		3: func(ctx *migrations.MigrationContext, parameters []byte) error { return nil },
```
enlight: Update one of these functions to expect the parameters to match a particular value, and update the tests below to pass that value.

tx_handler/migration_tx_handler.go
```diff
@@ -86,13 +87,12 @@ func (h *MigrationTxHandler) ProcessTx(
...
 	}
-
 	migrationCtx := migrations.NewMigrationContext(h.Manager, h.CreateRegistry, state, origin)
-	if err := migrationFn(migrationCtx); err != nil {
+	if err := migrationFn(migrationCtx, tx.Input); err != nil {
 		return r, errors.Wrapf(err, "migration %d failed", int32(tx.ID))
 	}
 
-	state.Set(migrationKey(tx.ID), msg.Data)
+	state.Set(migrationKey(tx.ID), []byte{1})
```
pathornteng: Why is this changed? If you want to change this to save some space, a feature flag is needed or @enlight approve this.

marktrs: yep @enlight approved 
https://github.com/loomnetwork/loomchain/pull/1431#pullrequestreview-278234687

pathornteng: Approved with a feature flag lol, so you need to wrap this with a feature flag to prevent consensus failure.

enlight: Since migration txs can have input parameters now it doesn't make much sense to store all that in the app store, we only use this key to check if a migration has already ran, the value we store here is irrelevant.

This change does need a new feature flag.

Issue #1436, opened 2019-08-23
`loom reset` command should delete evm.db


enlight:
Currently it deletes `app.db` without deleting `evm.db`.

enlight:
Resolved by https://github.com/loomnetwork/loomchain/pull/1437

PR #1437, opened 2019-08-23
loom reset command now remove evm.db

atchapcyp:
ref : https://github.com/loomnetwork/loomchain/issues/1436


cmd/loom/loom.go
```diff
@@ -547,6 +552,10 @@ func resetApp(cfg *config.Config) error {
 	return destroyDB(cfg.DBName, cfg.RootPath())
 }
 
+func resetEvm(cfg *config.Config) error {
```
enlight: Not much point in having a one line function, just inline this code in the command implementation.

PR #1439, opened 2019-08-23
Init default config

pathornteng:
Currently, if we change config protobuf to add a new setting and the old version of config exists in store state, the new config protobuf will be over-written by the old one, making the new setting value `nil`. As a result, we cannot set its value as it is uninitialised (nil). To solve this problem, this PR initialises all the config settings with 0. The default values are read from genesis file and stored in chain state in `InitChain` .

### DIFF
--- a/app.go
+++ b/app.go
@@ -179,13 +179,7 @@ func (s *StoreState) SetFeature(name string, val bool) {
 // If an error occurs while trying to update the config the change is rolled back, if the rollback
 // itself fails this function will panic.
 func (s *StoreState) ChangeConfigSetting(name, value string) error {
-	chainCfgVersion1_4 := s.FeatureEnabled(ChainCfgVersion1_4, false)
-	var cfg *cctypes.Config
-	if chainCfgVersion1_4 {
-		cfg = loadOnChainConfig(s.store, chainCfgVersion1_4)
-	} else {
-		cfg = s.Config()
-	}
+	cfg := loadOnChainConfig(s.store)
 	backupConfigBytes, err := proto.Marshal(cfg)
 	if err != nil {
 		return err
@@ -208,7 +202,7 @@ func (s *StoreState) ChangeConfigSetting(name, value string) error {
 // Config returns the current on-chain config.
 func (s *StoreState) Config() *cctypes.Config {
 	if s.config == nil {
-		s.config = loadOnChainConfig(s.store, s.FeatureEnabled(ChainCfgVersion1_4, false))
+		s.config = loadOnChainConfig(s.store)
 	}
 	return s.config
 }
@@ -457,7 +451,7 @@ func (a *Application) InitChain(req abci.RequestInitChain) abci.ResponseInitChai
 		}
 	}
 	// load on-chain config
-	a.config = loadOnChainConfig(state, state.FeatureEnabled(ChainCfgVersion1_4, false))
+	a.config = loadOnChainConfig(state)
 
 	return abci.ResponseInitChain{}
 }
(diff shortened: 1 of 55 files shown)


app.go
```diff
@@ -179,7 +179,13 @@ func (s *StoreState) SetFeature(name string, val bool) {
 // If an error occurs while trying to update the config the change is rolled back, if the rollback
 // itself fails this function will panic.
 func (s *StoreState) ChangeConfigSetting(name, value string) error {
-	cfg := s.Config()
+	chainCfgVersion1_4 := s.FeatureEnabled(ChainCfgVersion1_4, false)
```
enlight: I haven't enabled it anywhere yet, don't need a separate feature flag for this fix

app.go
```diff
@@ -450,6 +450,9 @@ func (a *Application) InitChain(req abci.RequestInitChain) abci.ResponseInitChai
 			panic(err)
 		}
 	}
+	// load on-chain config
+	a.config = loadOnChainConfig(state)
```
enlight: Don't need to load it here, it gets lazy loaded in `StoreState.Config()`

plugin/chainconfig_manager.go
```diff
@@ -66,14 +66,14 @@ func (c *ChainConfigManager) EnableFeatures(blockHeight int64) error {
 }
 
 // UpdateConfig applies pending config changes to the on-chain config.
```
enlight: Document what it returns too

app.go
```diff
@@ -524,10 +526,14 @@ func (a *Application) BeginBlock(req abci.RequestBeginBlock) abci.ResponseBeginB
 			panic(err)
 		}
 
-		if err := chainConfigManager.UpdateConfig(); err != nil {
+		updatedConfig, err := chainConfigManager.UpdateConfig()
```
enlight: `updatedConfig` -> `numConfigChanges`

app.go
```diff
@@ -524,10 +526,14 @@ func (a *Application) BeginBlock(req abci.RequestBeginBlock) abci.ResponseBeginB
...
+		if err != nil {
 			panic(err)
 		}
 
+		if updatedConfig > 0 {
+			a.config = loadOnChainConfig(state)
+		}
```
enlight: I'd just do:
```golang
if numConfigChanges > 0 {
  // invalidate cached config so it's reloaded next time it's accessed
  a.config = nil
}
```

app.go
pathornteng: Done

plugin/chainconfig_manager.go
pathornteng: Done

app.go
pathornteng: Done

app.go
pathornteng: Yes, that would be better in term of performance I think

app.go
```diff
@@ -179,7 +179,7 @@ func (s *StoreState) SetFeature(name string, val bool) {
 // If an error occurs while trying to update the config the change is rolled back, if the rollback
 // itself fails this function will panic.
 func (s *StoreState) ChangeConfigSetting(name, value string) error {
-	cfg := s.Config()
+	cfg := loadOnChainConfig(s.store)
```
enlight: If you're reloading the config from the store before changing it then we don't need to bother with backup & rollback since the current config instance isn't modified. But we will need to update the cached config instance after `s.store.Set([]byte(configKey), configBytes)` (or at the very least reset the cached instance to `nil`)

pathornteng: Yeah, we do clear config cache after ChangeConfigSetting is called. https://github.com/loomnetwork/loomchain/pull/1439#discussion_r317565193

enlight: That's in another function, might as well be in another galaxy.

pathornteng: No it is in the same galaxy. `ChangeConfigSetting` is only called in `UpdateConfig`

app.go
```diff
@@ -188,6 +188,8 @@ func (s *StoreState) ChangeConfigSetting(name, value string) error {
 		return err
 	}
 	s.store.Set([]byte(configKey), configBytes)
+	// invalidate cached config so it's reloaded next time it's accessed
+	s.config = nil
```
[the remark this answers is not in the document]
enlight: Yep, restored the code that cleared `Application.config`

PR #1410, opened 2019-08-14
Increment nonce for failed txs

pathornteng:
Currently, account nonces do not get incremented if txs fail, this behavior differs from Ethereum and it also cause failed EVM txs to have duplicate tx hash. This PR fixes this by incrementing nonce for all failed txs. 

Ref: https://github.com/loomnetwork/loomchain/issues/1360
Might also fixed: https://github.com/loomnetwork/loomchain/issues/1409
Also fix https://github.com/loomnetwork/loomchain/issues/1424

- [x] I added unit tests for any code that added
- [ ] I updated the CHANGELOG.md 
- [x] All IP is original and not copied from another source
- [x] I assign all copyright to Loom Network for the code in the pull request

Issue #1360, opened 2019-07-22
Review nonce handling strategy


enlight:
Failed evm txs can be included in blocks, but they cannot alter chain state in any way, only successful txs can change chain state. The per-account tx nonce is stored in chain state so it doesn't get incremented when an EVM tx fails. We should investigate the possibility of incrementing nonces for failed txs to better emulate Ethereum,

pathornteng:
Saving nonce in chain state makes us unable to update nonce when transactions revert. One possible solution is to move all account nonces to receipts_db so it does not change app hash when we update nonce.

enlight:
OK, then we should probably increment the nonce if a tx fails in DeliverTx, but we need to be super careful that only the nonce is updated, and all other state changes are discarded.

enlight:
We can modify `NonceHandler` so it can be given a `KVStore` (reference to `Application.Store`) when the nonce middleware are constructed. That way the `NonceHandler` can increment the nonce unconditionally in `DeliverTx` directly via the `KVStore`.

pathornteng:
Expected NonceMiddleware Behavior

1. If a tx has been included in `BlockStore`, account nonce should increment. This means EVM reverted tx will increment account nonce.
2. An account can submit multiple txs in one blocks
3. Should nonce client speculate nonce or always query from the server? A tx with speculative nonce can get `Invallid nonce` if the previous tx does not pass `CheckTx`

Example
Block 1
Successful Tx -> nonce = 1
EVM reverted Tx -> nonce = 2
Successful Tx -> nonce = 3

Block 2
Successful Tx -> nonce = 4
Error Tx (Tx does not pass CheckTx) -> nonce  = 4 
Successful Tx -> nonce = 5

enlight:
Should be resolved by https://github.com/loomnetwork/loomchain/pull/1410

PR #1442, opened 2019-08-27
Increment cached nonce for failed tx in DeliverTx

pathornteng:
In this PR  https://github.com/loomnetwork/loomchain/pull/1410,  we decided to increment nonce for failed EVM but we did not increment cached nonce in DeliverTx. This is causing a problem when an account sends multiple txs in one block and some of them are failed txs since the cached nonce in DeliverTx is not incremented to match nonce in store state. This PR fixes this by incrementing nonce for failed txs.

Ref: https://github.com/loomnetwork/loomchain/issues/1360

### DIFF
--- a/middleware.go
+++ b/middleware.go
@@ -7,6 +7,8 @@ import (
 	"runtime/debug"
 	"time"
 
+	"github.com/loomnetwork/loomchain/features"
+
 	"github.com/go-kit/kit/metrics"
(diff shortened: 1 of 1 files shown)


pathornteng:
Close this as the problem is fixed in https://github.com/loomnetwork/loomchain/pull/1447

Issue #1443, opened 2019-08-27
Avoid pulling data out of tx_index.db in EthGetBlockByNumber / EthGetBlockByHash


enlight:
`EthGetBlockByNumber` and `EthGetBlockByHash` call `GetBlockByNumber` which attempts to load data from `tx_index.db`
https://github.com/loomnetwork/loomchain/blob/141380a5e0c157f3492774f74e960c51dcfaa0e5/eth/query/block.go#L84

On some production clusters that `tx_index.db` may be missing data so `EthGetBlockByNumber` & `EthGetBlockByNumber` fail with an error. We don't actually need to rely on `tx_index.db` in this case because the tx result data we need can be obtained from `TendermintBlockStore.GetBlockResults` (which pulls data from `state.db`). Furthermore the data in question is only needed for EVM txs, so we should avoid retrieving it.

In short in `GetBlockByNumber` for an EVM tx try to get data from `blockStore.GetBlockResults` first, then fall back to `blockStore.GetTxResult` if necessary.

enlight:
This issue can be reproduced using:
```
curl -X POST --data '{"jsonrpc":"2.0","method":"eth_getBlockByNumber","id":1,"params":["0x5", true]}' https://plasma.dappchains.com/eth
```

enlight:
Resolved by https://github.com/loomnetwork/loomchain/pull/1446

PR #1446, opened 2019-08-27
GetBlockByNumber should use tx result data from GetBlockResults

atchapcyp:
ref : https://github.com/loomnetwork/loomchain/issues/1443

### DIFF
--- a/eth/query/block.go
+++ b/eth/query/block.go
@@ -79,21 +79,24 @@ func GetBlockByNumber(
 	blockInfo.Number = eth.EncInt(height)
 	bloomFilter := evmAuxStore.GetBloomFilter(uint64(height))
 	blockInfo.LogsBloom = eth.EncBytes(bloomFilter)
+	blockResults, err := blockStore.GetBlockResults(&height)
+	if err != nil {
+		return resp, err
+	}
 	for index, tx := range blockResult.Block.Data.Txs {
 		if full {
-			var txResultBytes []byte
-			txResult, err := blockStore.GetBlockResults(&height)
-			if txResult == nil {
-				txResult, err := blockStore.GetTxResult(tx.Hash())
+			var blockResultBytes []byte
+			if blockResults == nil {
+				blockResults, err := blockStore.GetTxResult(tx.Hash())
 				if err != nil {
 					return resp, errors.Wrapf(err, "cant find tx details, hash %X", tx.Hash())
 				}
-				txResultBytes = txResult.TxResult.Data
+				blockResultBytes = blockResults.TxResult.Data
 			} else {
-				txResultBytes = txResult.Results.DeliverTx[index].Data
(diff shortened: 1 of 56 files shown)


eth/query/block.go
```diff
@@ -81,12 +81,19 @@ func GetBlockByNumber(
 	blockInfo.LogsBloom = eth.EncBytes(bloomFilter)
 	for index, tx := range blockResult.Block.Data.Txs {
 		if full {
-			txResult, err := blockStore.GetTxResult(tx.Hash())
-			if err != nil {
-				return resp, errors.Wrapf(err, "cant find tx details, hash %X", tx.Hash())
+			var txResultBytes []byte
+			txResult, err := blockStore.GetBlockResults(&height)
```
pathornteng: `height` does not change in the loop so `GetBlockResult` will return the same data, move this one out of `range blockResult.Block.Data.Txs` and change `txResult` to `blockResults`

eth/query/block.go
```diff
@@ -79,14 +79,24 @@ func GetBlockByNumber(
 	blockInfo.Number = eth.EncInt(height)
 	bloomFilter := evmAuxStore.GetBloomFilter(uint64(height))
 	blockInfo.LogsBloom = eth.EncBytes(bloomFilter)
+	blockResults, err := blockStore.GetBlockResults(&height)
```
pathornteng: We only need to call `GetBlockResults` if `full` is true.

eth/query/block.go
```diff
@@ -79,14 +79,24 @@ func GetBlockByNumber(
...
+		return resp, err
+	}
 	for index, tx := range blockResult.Block.Data.Txs {
 		if full {
-			txResult, err := blockStore.GetTxResult(tx.Hash())
-			if err != nil {
-				return resp, errors.Wrapf(err, "cant find tx details, hash %X", tx.Hash())
+			var blockResultBytes []byte
+			if blockResults == nil {
+				blockResults, err := blockStore.GetTxResult(tx.Hash())
```
pathornteng: For this line, change `blockResults` to  `txResult`

eth/query/block.go
```diff
@@ -79,14 +79,28 @@ func GetBlockByNumber(
 	blockInfo.Number = eth.EncInt(height)
 	bloomFilter := evmAuxStore.GetBloomFilter(uint64(height))
 	blockInfo.LogsBloom = eth.EncBytes(bloomFilter)
+	var blockResults *ctypes.ResultBlockResults
+	if full {
+		blockResults, err = blockStore.GetBlockResults(&height)
+		if err != nil {
+			return resp, errors.Wrapf(err, "results for block %v", height)
+		}
```
enlight: An error here means the `blockResults` couldn't be loaded, if you return immediately then `blockResults == nil` always be `false` and the code will never fall back to using `blockStore.GetTxResult(tx.Hash())`.

PR #1448, opened 2019-08-28
Replace nonce handler feature flag with on-chain config

pathornteng:
Replace `IncrementNonceOnFailedTxFeature` feature flag with on-chain config so that we can toggle the new behavior of `NonceHandler`.  

go-loom: https://github.com/loomnetwork/go-loom/pull/451


auth/auth.go
```diff
@@ -119,7 +118,11 @@ func (n *NonceHandler) Nonce(
 		n.nonceCache = make(map[string]uint64)
 	}
 	var seq uint64
-	if state.FeatureEnabled(features.IncrementNonceOnFailedTxFeature, false) && !isCheckTx {
+
+	cfg := state.Config()
+	IncrementNonceOnFailedTx := cfg.GetNonceHandler().GetIncNonceOnFailedTx() > 0
```
enlight: `IncrementNonceOnFailedTx` -> `incrementNonceOnFailedTx`
And why do we need `cfg` if you only access it once?

auth/auth_test.go
```diff
@@ -13,10 +13,13 @@ import (
 	loom "github.com/loomnetwork/go-loom"
 	"github.com/loomnetwork/go-loom/auth"
 	"github.com/loomnetwork/loomchain"
-	"github.com/loomnetwork/loomchain/features"
 	"github.com/loomnetwork/loomchain/store"
 )
 
+const (
+	IncrementNonceOnFailedTx = "NonceHandler.IncNonceOnFailedTx"
```
enlight: `IncrementNonceOnFailedTx` -> `incrementNonceOnFailedTx`
No need to export this from a test.

auth/auth_test.go
```diff
@@ -62,7 +65,8 @@ func TestSignatureTxMiddlewareMultipleTxSameBlock(t *testing.T) {
 	ctx = context.WithValue(ctx, ContextKeyCheckTx, true)
 	kvStore := store.NewMemStore()
 	state := loomchain.NewStoreState(ctx, kvStore, abci.Header{Height: 27}, nil, nil)
-	state.SetFeature(features.IncrementNonceOnFailedTxFeature, true)
+	err = state.ChangeConfigSetting(IncrementNonceOnFailedTx, "1")
```
enlight: Can do `loomchain.NewStoreState(...).WithOnChainConfig(cfg)` instead, fewer hardcoded strings to deal with.

auth/auth_test.go
```diff
@@ -153,6 +161,12 @@ func TestRevertedTxNonceMiddleware(t *testing.T) {
 	})
 	require.NoError(t, err)
 
+	nonceTxBytes3, err := proto.Marshal(&NonceTx{
```
enlight: Move this down, closer to where it's actually used.

auth/auth.go
pathornteng: Fixed

auth/auth_test.go
pathornteng: Fixed

auth/auth_test.go
pathornteng: Fixed

auth/auth_test.go
pathornteng: Fixed

Issue #1444, opened 2019-08-27
Implement eth_getStorageAt on /eth endpoint


enlight:
We need to implement [`eth_getStorageAt`](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_getstorageat) on the `/eth` endpoint to support upgradable OpenZepplin contracts as requested in https://github.com/loomnetwork/loom-js/issues/306

pathornteng:
Probably a good reference: https://github.com/ethereum/go-ethereum/blob/54b271a86dd748f3b0bcebeaf678dc34e0d6177a/internal/ethapi/api.go#L727

pathornteng:
@eduardonunesp that's really useful articles. Thanks a lot.

PR #1449, opened 2019-08-28
Add eth_getStorageAt method to /eth endpoint

atchapcyp:
ref: https://github.com/loomnetwork/loomchain/issues/1444


e2e/tests/truffle/test/EthGetStorage.js
```diff
@@ -0,0 +1,41 @@
...
+    ethUrl = `http://${nodeAddr}/eth`
+    alice = accounts[1];
+    bob = accounts[2];
+  });
+
+  it('eth_getStorageAt', async () => {
+    const storeContract = await StoreTestContract.deployed();
+
+    result = await getStorageAt(ethUrl, storeContract.address, "0x00")
+    assert.equal(result, '0x000000000000000000000000000000000000000000000000000000000000000f')
```
pathornteng: Can you add comments on each `assert.equal` explaining what value from what variable do you expect?

e2e/tests/truffle/test/EthGetStorage.js
```diff
@@ -0,0 +1,41 @@
...
+contract('StoreTestContract', async (accounts) => {
+
+  beforeEach(async () => {
+    if (!process.env.CLUSTER_DIR) {
+      throw new Error('CLUSTER_DIR env var not defined');
+    }
+    let nodeAddr = fs.readFileSync(path.join(process.env.CLUSTER_DIR, '0', 'node_rpc_addr'), 'utf-8');
+    web3js = new Web3(new Web3.providers.HttpProvider(`http://${nodeAddr}/eth`));
+    ethUrl = `http://${nodeAddr}/eth`
+    alice = accounts[1];
```
pathornteng: Remove if not used

evm/evm.go
```diff
@@ -259,6 +259,11 @@ func (e Evm) GetCode(addr loom.Address) []byte {
 	return e.sdb.GetCode(common.BytesToAddress(addr.Local))
 }
 
+func (e Evm) GetStorageAt(addr loom.Address, key []byte) ([]byte, error) {
+	result := e.sdb.GetState(common.BytesToAddress(addr.Local), common.BytesToHash(key))
+	return result.Bytes(), nil
```
pathornteng: remove `error` if it is always nil

plugin/vm.go
```diff
@@ -217,6 +217,10 @@ func (vm *PluginVM) GetCode(addr loom.Address) ([]byte, error) {
 	return []byte{}, nil
 }
 
+func (vm *PluginVM) GetStorageAt(addr loom.Address, key []byte) ([]byte, error) {
+	return []byte{}, nil
```
pathornteng: ditto

e2e/tests/truffle/test/EthGetStorage.js
```diff
@@ -0,0 +1,41 @@
...
+    result = await getStorageAt(ethUrl, storeContract.address, "0x01")
+    assert.equal(result, '0x000000000000000000000000000004d20000000000000000000000000000429f')
+
+    result = await getStorageAt(ethUrl, storeContract.address,"0x02")
+    result = web3js.utils.hexToUtf8(result)
+    assert.equal(result, 'test1')
+
+    result = await getStorageAt(ethUrl, storeContract.address, "0x03")
+    result = web3js.utils.hexToUtf8(result)
+    assert.equal(result, "test1236", "invalid value get storage")
```
pathornteng: Did you test get map variable?

e2e/tests/truffle/contracts/StoreTestContract.sol
```diff
@@ -0,0 +1,45 @@
...
+
+	bytes16 string1 = 'test1';
+	bytes32 string2 = 'test1236';
+	string string3 = 'lets string something';
+
+	mapping (address => uint) uints1;
+	mapping (address => DeviceData) structs1;
+
+	uint[] uintarray;
+	DeviceData[] deviceDataArray;
```
pathornteng: Can you please simplify the contract, leaving only used variables?

rpc/query_server.go
```diff
@@ -1149,3 +1172,13 @@ func getTxByTendermintHash(blockStore store.BlockStore, hash []byte) (eth.JsonTx
 	txObj, _, err := query.GetTxObjectFromBlockResult(blockResult, txResults.TxResult.Data, int64(txResults.Index))
 	return txObj, err
 }
+
+func decodeHexToBytes(value string) ([]byte, error) {
```
pathornteng: Probably need to fix `DecDataToBytes` instead of creating a new method

rpc/eth/jsonrpc_conversion.go
```diff
@@ -406,6 +406,9 @@ func DecDataToBytes(value Data) ([]byte, error) {
 	if len(value) <= 2 || value[0:2] != "0x" {
 		return []byte{}, errors.Errorf("invalid data format: %v", value)
 	}
+	if len(value)%2 != 0 {
+		value = "0x0" + value[2:]
+	}
```
enlight: According to the web3 JSON-RPC spec the UNFORMATTED DATA value must always have an even number of digits https://github.com/ethereum/wiki/wiki/JSON-RPC#hex-value-encoding so if `len(value) % 2 != 0` we should return an error.

pathornteng: Umm, `0x0` is a valid param though. Take a look at eth_getStorageAt example here. https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_getstorageat

rpc/query_server.go
```diff
@@ -1054,6 +1054,29 @@ func (s *QueryServer) EthGetBalance(address eth.Data, block eth.BlockHeight) (et
...
 }
 
+// https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_getStorageAt
+func (s *QueryServer) EthGetStorageAt(local eth.Data, position eth.Data, block eth.BlockHeight) (eth.Data, error) {
+	address, err := eth.DecDataToAddress(s.ChainID, local)
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to decode address parameter %v", local)
+	}
+
+	positionBytes, err := eth.DecDataToBytes(position)
```
enlight: https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_getstorageat is a confusing mess, according to the spec the `position` should be an `eth.Quantity`, but it's not really used as a number, just as a hash... So lets make `position` a `string` and use `common.HexToHash(position)` to parse it, just like go-ethereum does https://github.com/ethereum/go-ethereum/blob/b8a94571397c1afd5579f1e319aff05a676c917d/internal/ethapi/api.go#L727

rpc/eth/jsonrpc_conversion.go
enlight: Good point, `DecDataToBytes` was actually the wrong thing to use. See https://github.com/loomnetwork/loomchain/pull/1449#discussion_r323075687

rpc/query_server.go
```diff
@@ -1054,6 +1056,25 @@ func (s *QueryServer) EthGetBalance(address eth.Data, block eth.BlockHeight) (et
 	return eth.EncBigInt(*amount.Int), nil
 }
 
+// https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_getStorageAt
+func (s *QueryServer) EthGetStorageAt(local eth.Data, position string, block eth.BlockHeight) (eth.Data, error) {
```
enlight: We shouldn't ignore the block height parameter, should accept "latest" or a number matching `snapshot.Block().Height`, return `fmt.Errors("unable to get storage at height %v", block)` for any another value.

enlight: An empty `block` parameter is also fine, in that case it should be assumed to be whatever the snapshot height is (i.e. "latest").

PR #1462, opened 2019-09-03
Tweak gateway migration

marktrs:

ref loomnetwork/transfer-gateway#59
tg : loomnetwork/transfer-gateway#83
- added unmarshaler to parse JSON config input
- added error handling for SwitchMainnetGateway function


migrations/gateway_migration.go
```diff
@@ -3,24 +3,33 @@
...
+	"github.com/loomnetwork/loomchain/plugin"
 	"github.com/loomnetwork/transfer-gateway/builtin/plugins/gateway"
 )
 
-// GatewayMigration upgrades the Gateway contracts to consensus-reactor based multi-sig.
 func GatewayMigration(ctx *MigrationContext, parameters []byte) error {
 	gwMigrationRequest := tgtypes.TransferGatewaySwitchMainnetGatewayRequest{}
-	err := proto.Unmarshal(parameters, &gwMigrationRequest)
+
+	unmarshaler, err := contractpb.UnmarshalerFactory(plugin.EncodingType_JSON)
```
enlight: Huh, why?

marktrs: input file will be like this in json format
```
{
  "gateway_name": "gateway",
  "mainnet_address": {
    "chain_id": "eth",
    "local": "7Oz8VDa9juDI6YN589md/ebhX7k="
  }
}
```
so we need to do unmarshalling it

enlight: So what was wrong with `proto.Unmarshal`?

marktrs: will get `unexpected EOF` error

enlight: Ah, so the migration input stuff didn't actually work before. Instead of `contractpb.UnmarshalerFactory` use `jsonpb.Unmarshaler` directly. And add a unit test to check encoding & decoding of the parameters (in GatewayMigration`) actually works.

Issue #1465, opened 2019-09-03
Fix Go contract emits events on failed txs


pathornteng:
Currently, if a tx passes in `CheckTx` but fails in `DeliverTx`, all the emitted events before the tx fails will still be saved to `event.db`. We need to clear out emitted events if the tx fails in `DeliverTx`. This problem seems to be tricky since we don't generate tx hash for Go tx like EVM tx.

enlight:
Should be resolved by https://github.com/loomnetwork/loomchain/pull/1467

PR #1467, opened 2019-09-04
Fix Go event emitting on failed txs

pathornteng:
Ref: https://github.com/loomnetwork/loomchain/issues/1465

### DIFF
--- a/app.go
+++ b/app.go
@@ -677,12 +677,9 @@ func (a *Application) processTx(txBytes []byte, isCheckTx bool) (TxHandlerResult
 
 	receiptHandler := a.ReceiptHandlerProvider.Store()
 	defer receiptHandler.DiscardCurrentReceipt()
-
+	defer a.EventHandler.Rollback()
 	r, err := a.TxHandler.ProcessTx(state, txBytes, isCheckTx)
 	if err != nil {
-		// clear event cache on failed txs
-		a.EventHandler.Purge()
-		storeTx.Rollback()
 		// TODO: save receipt & hash of failed EVM tx to node-local persistent cache (not app state)
 		return r, err
 	}
@@ -719,9 +716,6 @@ func (a *Application) processTx(txBytes []byte, isCheckTx bool) (TxHandlerResult
 			}
 		}
 		storeTx.Commit()
-	} else {
-		// clear event cache for CheckTx
-		a.EventHandler.Purge()
 	}
 	return r, nil
 }
--- a/app.go
+++ b/app.go
@@ -680,6 +680,8 @@ func (a *Application) processTx(txBytes []byte, isCheckTx bool) (TxHandlerResult
 
 	r, err := a.TxHandler.ProcessTx(state, txBytes, isCheckTx)
 	if err != nil {
+		// clear event cache on failed txs
+		a.EventHandler.Purge()
 		storeTx.Rollback()
 		// TODO: save receipt & hash of failed EVM tx to node-local persistent cache (not app state)
 		return r, err
@@ -718,8 +720,8 @@ func (a *Application) processTx(txBytes []byte, isCheckTx bool) (TxHandlerResult
 		}
 		storeTx.Commit()
 	} else {
-		// clear event cache
-		a.EventHandler.Purge(uint64(a.curBlockHeader.GetHeight()))
+		// clear event cache for CheckTx
+		a.EventHandler.Purge()
 	}
 	return r, nil
 }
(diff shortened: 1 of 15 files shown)


event.go
```diff
@@ -34,6 +35,7 @@ type EventDispatcher interface {
 type DefaultEventHandler struct {
 	dispatcher             EventDispatcher
 	stash                  *stash
+	stashCache             *stash
```
enlight: The `stash` has to support multiple blocks because we don't wait for events to be emitted before moving onto the next block, the cache on the other hand should only hold events for the current tx.

pathornteng: ok fixed

app.go
```diff
@@ -680,12 +680,18 @@ func (a *Application) processTx(txBytes []byte, isCheckTx bool) (TxHandlerResult
 
 	r, err := a.TxHandler.ProcessTx(state, txBytes, isCheckTx)
 	if err != nil {
+		// clear event cache on failed txs
+		a.EventHandler.Purge()
```
enlight: Should use the same commit/rollback pattern as the receipt handler.
Call `defer a.EventHandler.Rollback()` before calling `a.TxHandler.ProcessTx`.

event.go
```diff
@@ -67,10 +71,23 @@ func (ed *DefaultEventHandler) Post(height uint64, msg *types.EventData) error {
...
+	ed.eventCache = append(ed.eventCache, &eventData)
 	return nil
 }
 
+func (ed *DefaultEventHandler) Commit(height uint64) error {
+	for _, eventData := range ed.eventCache {
+		ed.stash.add(height, eventData)
+	}
+	// clear event cache
+	ed.eventCache = make([]*EventData, 0)
```
enlight: Just do `ed.eventCache = nil` anywhere you want to clear it, no point calling `make` if the capacity is zero.

event.go
```diff
@@ -43,6 +46,7 @@ func NewDefaultEventHandler(dispatcher EventDispatcher) *DefaultEventHandler {
 	return &DefaultEventHandler{
 		dispatcher:             dispatcher,
 		stash:                  newStash(),
+		eventCache:             make([]*EventData, 0),
```
enlight: Can omit this

app.go
pathornteng: Cool, fixed.

event.go
pathornteng: Done

event.go
pathornteng: Done

event.go
```diff
@@ -67,10 +70,22 @@ func (ed *DefaultEventHandler) Post(height uint64, msg *types.EventData) error {
 	}
 	// TODO: this is stupid, fix it
 	eventData := EventData(*msg)
-	ed.stash.add(height, &eventData)
+	ed.eventCache = append(ed.eventCache, &eventData)
 	return nil
 }
 
+func (ed *DefaultEventHandler) Commit(height uint64) error {
```
enlight: Since it never returns an error might as well simplify the function signature.

pathornteng: Done

event.go
```diff
@@ -67,10 +70,21 @@ func (ed *DefaultEventHandler) Post(height uint64, msg *types.EventData) error {
...
 	// TODO: this is stupid, fix it
 	eventData := EventData(*msg)
-	ed.stash.add(height, &eventData)
+	ed.eventCache = append(ed.eventCache, &eventData)
 	return nil
 }
 
+func (ed *DefaultEventHandler) Commit(height uint64) {
+	for _, eventData := range ed.eventCache {
+		ed.stash.add(height, eventData)
```
enlight: Seeing as we now add batches of events to the stash rather than one event at a time it'd be good to update the stash interface to accept an array of events to reduce locking.

app.go
```diff
@@ -677,15 +677,17 @@ func (a *Application) processTx(txBytes []byte, isCheckTx bool) (TxHandlerResult
 
 	receiptHandler := a.ReceiptHandlerProvider.Store()
 	defer receiptHandler.DiscardCurrentReceipt()
+	defer a.EventHandler.Rollback()
 
 	r, err := a.TxHandler.ProcessTx(state, txBytes, isCheckTx)
 	if err != nil {
-		storeTx.Rollback()
```
enlight: Hmm, why did this get removed?

pathornteng: It is unintentional. I will restore it.

enlight:
This PR is meant to address two issues with the event handling:
1. Previously, when a Go contract was called any event it emitted was stored in `events.db` & sent back to subscribers, even if the Go contract ultimately returned an error. However, since any state changes made within the Go contract are discarded on error the events emitted to reflect those state changes must not be stored nor sent to subscribers, these change ensure those events are discarded.
2. The event stash was constantly leaking memory because events were added to the stash in `CheckTx` at height N after the stash was already purged at height N (in `Application.Commit`), hence the stash kept accumulating events that were never discarded.

PR #1468, opened 2019-09-04
Log the errors from EvmAuxStore.SaveChildTxRefs

enlight:


app.go
```diff
@@ -720,7 +723,10 @@ func (a *Application) Commit() abci.ResponseCommit {
 		panic(err)
 	}
 
-	a.EvmAuxStore.SaveChildTxRefs(a.childTxRefs)
+	if err := a.EvmAuxStore.SaveChildTxRefs(a.childTxRefs); err != nil {
+		// TODO: consider panic instead
+		log.Error("Failed to save Tendermint -> EVM tx hash refs", "err", err)
```
pathornteng: Should we print out `a.childTxRefs` as well?

enlight: No, because there could be dozens of them, I will add the height though to make it easier to track down the cause should this ever actually fail.

PR #1475, opened 2019-09-09
Add GetValidator method for RPC endpoint

atchapcyp:
In order to support staking with `trustwallet`, We need to provide RPC endpoints to fetch the following data:
- Validators list
- Staking pool (if applicable)
- Current annual APR 

ref : https://developer.trustwallet.com/staking#step-3-define-rpc-endpoints-for-blockatlas

### DIFF
--- a/rpc/query_server.go
+++ b/rpc/query_server.go
@@ -7,9 +7,9 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/gorilla/websocket"
-
 	"github.com/gogo/protobuf/proto"
+	"github.com/gorilla/websocket"
+	"github.com/loomnetwork/loomchain/builtin/plugins/dposv3"
 	sha3 "github.com/miguelmota/go-solidity-sha3"
 	"github.com/phonkee/go-pubsub"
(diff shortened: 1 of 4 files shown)


rpc/query_server.go
```diff
@@ -1073,6 +1082,79 @@ func (s *QueryServer) EthAccounts() ([]eth.Data, error) {
...
+
+	candidates, err := dposv3.LoadCandidateList(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	getValidatorResponse := make([]blockatlas.Validator, 0)
+	chainID := ctx.Block().ChainID
+
+	for _, validator := range validators {
```
pathornteng: This loop has time complexity equal to n^2. We can do it better by using hash map.

First we loop to save validator address in hash map like
```
validatorList := map[string]loom.Address
for _, validator := range validators {
  validatorList[string(validator.Pubkey)] = loom.Address{ChainID: chainID, Local: loom.LocalAddressFromPublicKey(validator.PubKey)}
}
```
and then we can just iterate candidates that are in the validators list

PR #1481, opened 2019-09-13
Change tx hash generation algorithm

pathornteng:
Ref: https://github.com/loomnetwork/loomchain/issues/1240


e2e/tests/truffle/test/helpers.js
```diff
@@ -18,6 +18,12 @@ async function delay(delayInms) {
   });
 }
 
+function toHexString(byteArray) {
```
enlight: Does `Buffer.from(byteArray).toString('hex')` not work?

pathornteng: yes it works, fixed

PR #1484, opened 2019-09-16
Fix jailing jailed validator

pathornteng:
It's pointless to jail the validator that is already jailed.

### DIFF
--- a/builtin/plugins/dposv3/dpos.go
+++ b/builtin/plugins/dposv3/dpos.go
@@ -1521,7 +1521,10 @@ func UpdateDowntimeRecord(ctx contract.Context, downtimePeriod uint64, jailingEn
 	if ctx.FeatureEnabled(features.DPOSVersion3_4, false) {
 		jailOfflineValidator = jailingEnabled
 	}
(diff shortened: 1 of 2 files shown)


builtin/plugins/dposv3/dpos.go
```diff
@@ -1521,6 +1521,9 @@ func UpdateDowntimeRecord(ctx contract.Context, downtimePeriod uint64, jailingEn
 	if ctx.FeatureEnabled(features.DPOSVersion3_4, false) {
 		jailOfflineValidator = jailingEnabled
 	}
+	if ctx.FeatureEnabled(features.DPOSVersion3_6, false) {
+		jailOfflineValidator = jailOfflineValidator && !statistic.Jailed
```
enlight: I think it's fairly low risk to just do `if jailOfflineValidator && !statistic.Jailed` below, `emitJailEvent` doesn't change state itself, and it's highly unlikely to fail.

pathornteng: true, done

PR #1493, opened 2019-09-23
[WIP] Support blockatlas with GetBlockTxs endpoint 

atchapcyp:
We need to provide these endpoints to support `blockatlas`

-  query sent/received list of transactions for an account/address
-  query transactions in a block

This PR will add `ba_block_txs` endpoint that can get block information by block height.

ref: https://developer.trustwallet.com/wallet-core/rpc-requirements#for-block-atlas

### DIFF
--- a/rpc/blockatlas/jsonrpc_conversion.go
+++ b/rpc/blockatlas/jsonrpc_conversion.go
@@ -1,0 +1,237 @@
+package blockatlas
+
+import (
+	"encoding/hex"
+	"math/big"
+	"strconv"
+	"strings"
+
+	"github.com/loomnetwork/go-loom"
+	ltypes "github.com/loomnetwork/go-loom/types"
+	"github.com/pkg/errors"
+)
+
+// https://github.com/ethereum/wiki/wiki/JSON-RPC#hex-value-encoding
+// Eth JSON RPC define three types QUANTITIES, DATA and default block parameter.
+// All represented by strings.
+type Quantity string
+type Data string
+type BlockHeight string
+
+const (
+	ZeroedQuantity     Quantity = "0x0"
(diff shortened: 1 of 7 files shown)


rpc/blockatlas/jsonrpc_conversion.go
```diff
@@ -0,0 +1,249 @@
...
+	TransactionType  string      `json:"transactionType,omitempty"`
+	ContractName     string      `json:"contractName,omitempty"`
+	ContractMethod   string      `json:"contractMethod,omitempty"`
+	Nonce            Quantity    `json:"nonce,omitempty"`
+	BlockHash        Data        `json:"blockHash,omitempty"`
+	BlockNumber      Quantity    `json:"blockNumber,omitempty"`
+	TransactionIndex Quantity    `json:"transactionIndex,omitempty"`
+	From             string      `json:"from,omitempty"`
+	To               string      `json:"to"`
+	Value            interface{} `json:"value"`
```
atchapcyp: I still have problems with the type of `Value` field.

What I am trying to do is Decode message and assign the value differently depending on the transaction type 
(Ex `DelegateValue`, `ReDelegateValue` etc.. )

And when I test by query the block that has a trasaction. I got ..
```
{
  "jsonrpc": "2.0",
  "id": "",
  "error": {
    "code": -32603,
    "message": "Internal error",
    "data": "Error marshalling response: Unregistered interface interface {}"
  }
}
```

So Is there a way to add custom struct to these `value` field ?

enlight: Try using `json.RawMessage` instead of `interface{}` and `json.Marshal` the value before putting it in `JsonTxObject`

Issue #1482, opened 2019-09-13
Improve debug trancing of the EVM calls


eduardonunesp:
In order to achieve greater detail about the internals of the calls and to show more details information on loomchain block explorer a call `debug_traceTransaction` should be added to loomchain.

https://github.com/ethereum/go-ethereum/wiki/Management-APIs#debug_tracetransaction

mattkanwisher:
ideally we enable this only on specific nodes

mattkanwisher:
would this also enable debug on remix?

eduardonunesp:
I'm investigating, as far I can say right now yes there's a relation

https://i.stack.imgur.com/QkWqB.png

eduardonunesp:
Remix uses the following calls
- `debug_storageRangeAt`
- `debug_traceTransaction`

enlight:
`debug_storageRangeAt` isn't documented at https://github.com/ethereum/go-ethereum/wiki/Management-APIs#debug but the PR which added it to go-ethereum is here https://github.com/ethereum/go-ethereum/pull/14350

Sriep:
Implemented in https://github.com/loomnetwork/loomchain/pull/1573

PR #1496, opened 2019-09-23
Add support for Remix debugging of EVM call

Sriep:

To use this PR you need three other PRs.
loomchian PR1492 [Refactor the State object from loomchain into a separate package.](https://github.com/loomnetwork/loomchain/pull/1491)
transfer gateway PR95 [Move config to separate package to reduce cyclic dependencies](https://github.com/loomnetwork/transfer-gateway/pull/95)
Change to go-ethereum to access some internal objects https://github.com/loomnetwork/go-ethereum/pull/8

Note that the optimal second parameter is not supported. So include `{}` instead of leaving it out. This is because it is using existing `QuerySever` endpoint mapping code that does not support optional parameters. This feature can be added when added to a [separate private interface](https://loom-network.slack.com/archives/G9TCVAV16/p1568713126011100?thread_ts=1568706549.007600&cid=G9TCVAV16). 

issue https://github.com/loomnetwork/loomchain/issues/1482
Includes refactoring of middleware object to separate package
Implements `debug_traceTransaction` as outlined in [Management APIs](https://github.com/ethereum/go-ethereum/wiki/Management-APIs#debug_tracetransaction)
Based off https://github.com/loomnetwork/loomchain/pull/1491 .


Sriep:
There is a lot of refactoring needed to allow transactions to be run off-chain in separate threads. 

## Refactoring I
These changes do not change code except to move code into different or new packages. The aim is to prevent cyclic imports by reducing imports of loomchian root directory. And isolating code called, directly or indirectly, from loomchain root.
* Uses loomchain: PR91 to move the state into a separate package `loomchian\state`.
* `loomchain/config` move config structs into the config package. transfer-gatway:PR95 moves `gateway/config.go` into separate package `gatway/config/config.go`  to prevent import cycles including loomchian root.
* Move `loomchain/sequence.go` into the auth package, `loomchian/auth/sequence/sequence.go`. Seems to belong in auth.
* auth package, created new `lomchian/auth/keys` package for code exported to loomchiain root package.
* Created new txhandler package for the middleware code. The `loomchain/txhandler` root holding the interface and other exported objects. `loomchian/txhandler/middleware` the middleware. implementation.
* Moved `loomchian\tx_handler` package into `loomchian\txhandler` and renamed `migration`. 

## Refactoring II
Created a new TxHandlerFacotry. This object contains and replaces the TxHandler creation code from `loomchian\cmd\loom\loom.go`. It allows the creation of custom tx handles on demand.

## New Objects
*  loomchian\store new `iAVLVersionReader` . This is a `KVReader` that takes a version and an IAVLStore and allows you to view the data stored at in the IAVLStore at any version that still exists in the store.
* `splitStore` An `VersionedKVStore` bult from a KVReader and an empty store, using the KVReader to retrieve data and the empty store for any changes.

rpc/debug/trace.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/ethereum/go-ethereum/eth/ethapi"
+	"github.com/ethereum/go-ethereum/eth/tracers"
+	"github.com/pkg/errors"
+
+	"github.com/loomnetwork/loomchain/store"
+	"github.com/loomnetwork/loomchain/txhandler/middleware"
+)
+
+func TraceTransaction(
+	app middleware.InMemoryApp,
```
enlight: This should take the ABCI application interface as input (`abci.Application`).

rpc/debug/trace.go
```diff
@@ -0,0 +1,49 @@
...
+	blockstore store.BlockStore,
+	blockNumber, txIndex int64,
+	config eth.TraceConfig,
+) (interface{}, error) {
+	if err := app.RunUpTo(blockNumber, txIndex); err != nil {
```
enlight: Here you should replay the txs using the `abci.Application` interface, so you'd call `app.BeginBlock`, then `app.DeliverTx` for each tx that needs to be replayed.

store/multi_writer_app_store.go
```diff
@@ -354,3 +354,16 @@ func (s *multiWriterStoreSnapshot) Range(prefix []byte) plugin.RangeData {
 
 	return ret
 }
+
+func (m *MultiWriterAppStore) VersionExists(version int64) bool {
+	return m.appStore.VersionExists(version)
+}
+
+func (m *MultiWriterAppStore) RetrieveVersion(version int64) (VersionedKVStore, error) {
+	reader, err := newMultiWriterVersionReader(*m, version)
```
enlight: I don't think we need all these new stores (`iAVLVersionReader`, `multiWriterVersionReader`), you should be able to do something very similar to what `GetSnapshot()` does, in fact rather than adding new methods to `VersionedKVStore` you can probably modify `GetSnapshot()` -> `GetSnapshot(version int64)` and use `version == 0` to retrieve the last snapshot, and `version > 0` to retrieve a specific version.

`splitStore` is still going to be required to allow in-memory writes.

app.go
```diff
@@ -621,3 +606,30 @@ func (a *Application) ReadOnlyState() appstate.State {
...
+	}
+	if startVersion == 0 {
+		return nil, errors.Errorf("no saved version for height %d", blockNumber)
+	}
+	startStore, err := a.Store.RetrieveVersion(startVersion)
+	if err != nil {
+		return nil, err
+	}
+
+	return middleware.NewInMemoryApp(
```
enlight: This should really return a clone of `Application` with the snapshoted state, without all the irrelevant bits (e.g. `Application.QueryHandler`)

store/multi_writer_app_store.go
Sriep: You need `value, _, _ := s.IAVLStore.tree.GetVersionedWithProof(key, s.version)` somehwer as in
```to
func (s *SomeKVReader) Get(key []byte) []byte {
	value, _, _ := s.IAVLStore.tree.GetVersionedWithProof(key, s.version)
	return value
}
```

enlight: You need to obtain the `ImmutableTree` at the desired height/version and then use the regular Get/Has/Range like the snapshots do. This is basically was `MutableTree.GetVersionedWithProof` is doing every time you call it. Similarly snapshots work off the last saved `ImmutableTree`, and now we need to extend them to work with any previously saved `ImmutableTree` version.

Sriep: I have renamed iAVLVersionReader to a Snapshot and merged multiWriterVersionReader and newMultiWriterStoreSnapshot.

rpc/debug/trace.go
```diff
@@ -0,0 +1,87 @@
...
+		}
+		for i := 0; i < len(block.Block.Data.Txs); i++ {
+			if h != height || i != int(index) {
+				_ = app.DeliverTx(block.Block.Data.Txs[i])
+			} else {
+				return nil
+			}
+		}
+
+		_ = app.EndBlock(abci.RequestEndBlock{})
```
enlight: Should also call `app.Commit`, and pass in the real `RequestBeginBlock`/`RequestEndBlock`

replay/replayapp.go
```diff
@@ -0,0 +1,140 @@
...
+
+type ReplayApplication interface {
+	abci.Application
+	SetTracer(vm.Tracer) error
+}
+
+var _ abci.Application = (*replayApplication)(nil)
+var _ abci.Application = ReplayApplication(nil)
+
+type replayApplication struct {
```
enlight: This shouldn't exist, you should replay with a new instance of the `Application` that has the stores swapped out to in-memory ones, and mock event handlers / subscriptions. We can't just use a mock app for the replay, because we need all the state changing code in BeginBlock/EndBlock/Commit to run just like it did originally.

store/multi_writer_app_store.go
```diff
@@ -256,114 +261,60 @@ func (s *MultiWriterAppStore) Prune() error {
...
 	defer func(begin time.Time) {
 		getSnapshotDuration.Observe(time.Since(begin).Seconds())
 	}(time.Now())
-	appStoreTree := (*iavl.ImmutableTree)(atomic.LoadPointer(&s.lastSavedTree))
-	evmDbSnapshot := s.evmStore.GetSnapshot(appStoreTree.Version())
-	return newMultiWriterStoreSnapshot(evmDbSnapshot, appStoreTree)
+
+	if version == 0 {
+		version = (*iavl.ImmutableTree)(atomic.LoadPointer(&s.lastSavedTree)).Version()
+	}
```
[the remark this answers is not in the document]
Sriep: Just checking; Do we want to add an error return to VersionedKVStore.GetSnapshot?
```go
type VersionedKVStore interface {
         ...
	GetSnapshot(version int64) (Snapshot, error)
}
```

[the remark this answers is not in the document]
enlight: Hmm, `GetSnapshot(0)` should panic, but `GetSnapshot(N)` should probably return an error, I guess it'd be better to add a new function to the interface:
```golang
type VersionedKVStore interface {
  ...
  GetSnapshot() Snapshot
  GetSnapshotAt(version int64) (Snapshot, error)
}
```

store/iavlstore.go
```diff
@@ -262,9 +251,38 @@ func NewIAVLStore(db dbm.DB, maxVersions, targetVersion, flushInterval int64) (*
 }
 
 type iavlStoreSnapshot struct {
-	*IAVLStore
+	*iavl.ImmutableTree
 }
 
-func (s *iavlStoreSnapshot) Release() {
-	// noop
+func newIavlStoreSnapshot(store IAVLStore, version int64) (Snapshot, error) {
```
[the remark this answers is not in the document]
Sriep: What about when you somehow call `GetSnapshotAt` on an actual `IAVLStore` object?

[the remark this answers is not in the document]
enlight: Existing code assumes `IAVLStore` doesn't have real snapshots, so `IAVLStore.GetSnapshotAt` would only get called by the `MultiWriterAppStore` (which can just deal with the ImmutableTree directly instead), where else is it going to get called from?

rpc/debug/trace.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	result := app.DeliverTx(block.Block.Data.Txs[txIndex])
+
+	switch tracer := tracer.(type) {
+	case *vm.StructLogger:
+		return &ethapi.ExecutionResult{
+			Gas:         5,
+			Failed:      result.Code != abci.CodeTypeOK,
+			ReturnValue: fmt.Sprintf("%x", result),
```
enlight: Is `fmt.Sprintf` sufficient or is the resulting hex string required to have an even number of digits?

rpc/debug/trace.go
```diff
@@ -0,0 +1,115 @@
...
+
+func runUpTo(app *loomchain.Application, blockstore store.BlockStore, startHeight, height, index int64) error {
+	for h := startHeight; h <= height; h++ {
+		resultBlock, err := blockstore.GetBlockByHeight(&h)
+		if err != nil {
+			return err
+		}
+		_ = app.BeginBlock(requestBeginBlock(*resultBlock))
+
+		block, err := blockstore.GetBlockByHeight(&h)
```
enlight: This is going to return the same block as `resultBlock`

rpc/debug/trace.go
```diff
@@ -0,0 +1,115 @@
...
+		if err != nil {
+			return errors.Wrapf(err, "getting block information at height %v", h)
+		}
+		for i := 0; i < len(block.Block.Data.Txs); i++ {
+			if h != height || i != int(index) {
```
enlight: The height check only needs to be done once per block, not once per tx.

app.go
```diff
@@ -622,3 +621,53 @@ func (a *Application) ReadOnlyState() appstate.State {
 		a.GetValidatorSet,
 	)
 }
+
+func (a *Application) ReplayApplication(blockNumber uint64, blockstore store.BlockStore) (*Application, int64, error) {
```
enlight: So now that it's here instead of a separate package you should be able to eliminate most of the refactoring in this PR, the tx handlers don't need to move, the middleware doesn't need to move, the configs don't need to move. Probably easier to copy the relevant bits out to a fresh PR.

rpc/debug/trace.go
Sriep: Inspired by https://github.com/loomnetwork/go-ethereum/blob/loomchain/eth/api_tracer.go#L613 I am assuming it just all works?!

Issue #1495, opened 2019-09-23
Check open file handles limit on node startup


enlight:
Node should run `ulimit -n` first thing on startup and refuse to start if the limit is too low. This will prevent people from running into app hash mismatch errors when the node attempts to access all the LevelDB files.

PR #1503, opened 2019-09-26
Check Process limit on node start up

atchapcyp:
ref : https://github.com/loomnetwork/loomchain/issues/1495

### DIFF
--- a/cmd/loom/loom.go
+++ b/cmd/loom/loom.go
@@ -7,10 +7,12 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"os/exec"
 	"os/signal"
 	"path"
 	"path/filepath"
 	"sort"
+	"strconv"
 	"strings"
 	"syscall"
 	"time"
@@ -337,6 +339,17 @@ func newRunCommand() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "run [root contract]",
 		Short: "Run the blockchain node",
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			output, err := exec.Command("ulimit", "-n").Output()
+			if err != nil {
+				return err
+			}
+			if currentLimit, _ := strconv.ParseUint(string(output[:len(output)-1]), 10, 64); currentLimit < cfg.MinimumProcessNumber {
+				fmt.Printf("Require at least %d process limit, Only have %d", cfg.MinimumProcessNumber, currentLimit)
+				os.Exit(0)
+			}
+			return nil
+		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if err != nil {
 				return err
--- a/config/config.go
+++ b/config/config.go
@@ -141,6 +141,9 @@ type Config struct {
 
 	// Dragons
 	EVMDebugEnabled bool
+
+	// Minimum allowed of ProcessNumber
+	MinimumProcessNumber uint64
 }
 
 type Metrics struct {
@@ -388,6 +391,7 @@ func DefaultConfig() *Config {
 		CallEnabled:            true,
 		DPOSVersion:            3,
 		AllowNamedEvmContracts: false,
+		MinimumProcessNumber:   512,
 	}
 	cfg.TransferGateway = DefaultTGConfig(cfg.RPCProxyPort)
 	cfg.LoomCoinTransferGateway = DefaultLoomCoinTGConfig(cfg.RPCProxyPort)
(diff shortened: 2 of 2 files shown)


config/config.go
```diff
@@ -388,6 +391,7 @@ func DefaultConfig() *Config {
 		CallEnabled:            true,
 		DPOSVersion:            3,
 		AllowNamedEvmContracts: false,
+		MinimumProcessNumber:   512,
```
pathornteng: Should be name `MinimumFileDescriptorLimit`

cmd/loom/loom.go
```diff
@@ -337,6 +339,17 @@ func newRunCommand() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "run [root contract]",
 		Short: "Run the blockchain node",
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			output, err := exec.Command("ulimit", "-n").Output()
```
pathornteng: We should not rely on external linux command like ulimit. Instead, we should take a look at `syscall` library. Take a look at go-ethereum https://github.com/ethereum/go-ethereum/tree/fdb34b7a7c87142cde48a2ec092f77a56d66f773/common/fdlimit

Also, we should try to raise the file descriptor limit automatically instead of asking the users to manually do so.

config/config.go
```diff
@@ -141,6 +141,9 @@ type Config struct {
 
 	// Dragons
 	EVMDebugEnabled bool
+
+	// Minimum allowed of ProcessNumber
```
pathornteng: Minimum number of file descriptor limit

cmd/loom/loom.go
```diff
@@ -1288,6 +1295,18 @@ func startPushGatewayMonitoring(cfg *config.PrometheusPushGatewayConfig, log *lo
 	}
 }
 
+func checkFileDescriptorLimit(min uint64) error {
+	var rlimit syscall.Rlimit
+	if err := syscall.Getrlimit(syscall.RLIMIT_NOFILE, &rlimit); err != nil {
+		return err
+	}
+	if rlimit.Cur < min {
+		return errors.New("insufficient file descriptor")
```
pathornteng: Change this line to
```return fmt.Errorf("Current file descriptor limit (%d) is too low; minimum file descriptor limit is %d", rlimit.Cur, min)```

File: config/config.go [hunk 5]
```diff
@@ -383,11 +386,12 @@ func DefaultConfig() *Config {
...
-		DeployEnabled:          true,
-		CallEnabled:            true,
-		DPOSVersion:            3,
-		AllowNamedEvmContracts: false,
+		Oracle:                     "",
+		DeployEnabled:              true,
+		CallEnabled:                true,
+		DPOSVersion:                3,
+		AllowNamedEvmContracts:     false,
+		MinimumFileDescriptorLimit: 512,
```
pathornteng: Not sure 512 is enough but 500000 is definitely not practical.

cmd/loom/loom.go
```diff
@@ -337,6 +337,13 @@ func newRunCommand() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "run [root contract]",
 		Short: "Run the blockchain node",
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			if err := checkFileDescriptorLimit(cfg.MinimumFileDescriptorLimit); err != nil {
```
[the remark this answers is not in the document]
enlight: And skip the limit check if `MinimumFileDescriptorLimit == 0`

config/config.go
```diff
@@ -141,6 +141,9 @@ type Config struct {
 
 	// Dragons
 	EVMDebugEnabled bool
+
+	//Minimum number of file descriptor limit
+	MinimumFileDescriptorLimit uint64
```
enlight: Lets just call it `MinOpenFilesLimit`

cmd/loom/loom.go
```diff
@@ -1288,6 +1295,18 @@ func startPushGatewayMonitoring(cfg *config.PrometheusPushGatewayConfig, log *lo
 	}
 }
 
+func checkFileDescriptorLimit(min uint64) error {
+	var rlimit syscall.Rlimit
+	if err := syscall.Getrlimit(syscall.RLIMIT_NOFILE, &rlimit); err != nil {
+		return err
```
enlight: Print out the error and return `nil` in this case, we don't want nodes refusing to start if this call fails due to some OS quirk.

cmd/loom/loom.go
```diff
@@ -1288,6 +1295,18 @@ func startPushGatewayMonitoring(cfg *config.PrometheusPushGatewayConfig, log *lo
...
+	}
+	if rlimit.Cur < min {
+		return fmt.Errorf("Current file descriptor limit (%d) is too low; minimum file descriptor limit is %d", rlimit.Cur, min)
```
enlight: -> `"Current open file limit (%d) is too low, please set it to at least %d"`

PR #1505, opened 2019-09-26
Add Withdrawal Limit Feature Flag

phoorichet:

This PR adds `TGWithdrawalLimitFeature` feature flag.

### DIFF
--- a/cmd/loom/gateway/general_cmds.go
+++ b/cmd/loom/gateway/general_cmds.go
@@ -663,7 +663,7 @@ func newSetWithdrawFeeCommand() *cobra.Command {
 }
 
 func newSetWithdrawLimitCommand() *cobra.Command {
-	var totalLimit, accountLimit int64
+	var totalLimit, accountLimit, decimals int64
 	cmd := &cobra.Command{
 		Use:     "set-withdrawal-limit <gateway>",
 		Short:   "Sets maximum amount the gateway should allow withdrawal",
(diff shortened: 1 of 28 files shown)


cmd/loom/gateway/general_cmds.go
```diff
@@ -657,6 +661,76 @@ func newSetWithdrawFeeCommand() *cobra.Command {
 	return cmd
 }
 
+func newSetMaxWithdrawLimitCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "set-max-withdrawal-limit <total-amount-limit> <per-account-amount-limit> [gateway]",
```
enlight: It should be possible to change each limit individually:
```bash
set-withdrawal-limit [gateway] --total-limit <amount> --account-limit <amount>
```
So if the user specifies only `--total` then only the total limit should change.

jenkins.sh
```diff
@@ -5,6 +5,7 @@ set -ex
 PKG=github.com/loomnetwork/loomchain
 
 # setup temp GOPATH
+export GO111MODULE=off
```
enlight: What does this do?

phoorichet: The left over. I'll remove it.

cmd/loom/gateway/general_cmds.go
```diff
@@ -677,37 +678,21 @@ func newSetMaxWithdrawLimitCommand() *cobra.Command {
...
 
 			var name string
-			if len(args) <= 2 || strings.EqualFold(args[2], GatewayName) {
+			if len(args) == 0 || strings.EqualFold(args[0], GatewayName) {
 				name = GatewayName
-			} else if strings.EqualFold(args[2], LoomGatewayName) {
+			} else if strings.EqualFold(args[0], LoomGatewayName) {
 				name = LoomGatewayName
 			} else {
 				return errors.New("only Gateway or LoomCoin gateway is allowed to set max withdrawal limit")
```
enlight: LOOM can also be withdrawn through the Binance Gateway, so we should be able to set limits there too.

phoorichet: Right. I'll add it.

cmd/loom/gateway/general_cmds.go
```diff
@@ -657,6 +663,68 @@ func newSetWithdrawFeeCommand() *cobra.Command {
...
+			gatewayAddr, err := rpcClient.Resolve(name)
+			if err != nil {
+				return errors.Wrap(err, "failed to resolve DAppChain Gateway address")
+			}
+			gateway := client.NewContract(rpcClient, gatewayAddr.Local)
+
+			req := &tgtypes.TransferGatewaySetMaxWithdrawalLimitRequest{
+				MaxTotalDailyWithdrawalAmount:      &types.BigUInt{Value: *maxTotalAmount},
+				MaxPerAccountDailyWithdrawalAmount: &types.BigUInt{Value: *maxPerAccountAmount},
+			}
```
enlight: If I only specify one of the limits on the command line then the other will get reset to zero, that shouldn't happen, only the one I specify should change. Fetch the current limits, apply the requested changes, then submit them back to the cluster.

Issue #1501, opened 2019-09-25
Fix `eth_getBlockByNumber` & `eth_getBlockByHash`


enlight:
Both `eth_getBlockByNumber` & `eth_getBlockByHash` call `GetBlockByNumber` https://github.com/loomnetwork/loomchain/blob/294a1e46090c63a9c7ec54a22634544da63d1c4c/eth/query/block.go#L32

The tx hashes of EVM calls/deploys returned by this function change based on the `full` parameter, when `full == false` the tx hash for an EVM call/deploy is the hash of the marshalled `SignedTx`, but when `full == true` the tx hash for an EVM call/deploy is the one generated by LoomVm (https://github.com/loomnetwork/loomchain/blob/294a1e46090c63a9c7ec54a22634544da63d1c4c/evm/loomevm.go#L176) or the receipt handler (https://github.com/loomnetwork/loomchain/blob/294a1e46090c63a9c7ec54a22634544da63d1c4c/receipts/leveldb/leveldb.go#L57).

The Web3 API shouldn't be returning inconsistent tx hashes.

enlight:
Blockscout appears to primarily use `full == true`, so we need to fix the `full == false` case to return the same tx hashes as `full == true`.

enlight:
Should be resolved by https://github.com/loomnetwork/loomchain/pull/1506

PR #1506, opened 2019-09-27
Fix GetBlockByNumber

pathornteng:
Previously, `GetBlockByNumber` returns TM tx hash for EVM tx if `full` is `false`. This PR makes it return EVM tx hash for EVM tx.

Ref: https://github.com/loomnetwork/loomchain/issues/1501

### DIFF
--- a/eth/query/block.go
+++ b/eth/query/block.go
@@ -110,7 +110,15 @@ func GetBlockByNumber(
 			}
 			blockInfo.Transactions = append(blockInfo.Transactions, txObj)
 		} else {
-			blockInfo.Transactions = append(blockInfo.Transactions, eth.EncBytes(tx.Hash()))
+			txResult, err := blockStore.GetTxResult(tx.Hash())
+			if err != nil {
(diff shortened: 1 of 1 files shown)


enlight:
For future reference for EVM call & deploy txs we generate a tx hash that's different to the Tendermint tx hash. While for Go call & deploy txs the tx hash is the same as the Tendermint tx hash. The reason it works that way is because we want to allow a client to submit signed Ethereum txs to the DAppChain without having to mess around with protobuf encoding. So the client will submit a signed Ethereum tx to a Loom node, the node will encode it in a protobuf and Tendermint will generate a hash for that marshaled protobuf, but the client has no easy way to figure out what the Tendermint tx hash would be before it submits the signed Ethereum tx (not without doing protobuf encoding anyway), which is why the EVM tx hash is supposed to be generated from data that the client has before it submits the signed Ethereum tx. That's how it'll work eventually anyway, right now the EVM tx hash is generated in such a way that it's impossible for the client to predict it.

Issue #1494, opened 2019-09-23
Make a node panic if it detects an active feature flag not supported by the current build


enlight:
Currently the ChainConfigManager will cause a node to panic if it attempts to activate a feature that's not supported by the current build, however, if a node is restored from a jump-start on an old build the node will run until it hits an app hash mismatch... at which point it has to be restored from the jump-start again.

enlight:
Should probably do the same for on-chain config changes.

atchapcyp:
> Should probably do the same for on-chain config changes.

Currently, we don't have the build number collected in the on-chain config store.
It is impossible to check the build number of the config that already applied.
Need your suggestion.

enlight:
Suggested an approach in the PR

enlight:
Resolved by https://github.com/loomnetwork/loomchain/pull/1507

PR #1507, opened 2019-09-27
Make node panic on unsupported build number

atchapcyp:
issue ref : https://github.com/loomnetwork/loomchain/issues/1494


plugin/chainconfig_manager.go
```diff
@@ -84,3 +84,11 @@ func (c *ChainConfigManager) UpdateConfig() (int, error) {
 
 	return len(settings), nil
 }
+
+func (c *ChainConfigManager) CheckFeaturesBuildNumber(blockHeight int64) error {
+	err := chainconfig.CheckFeaturesBuildNumber(c.ctx, uint64(blockHeight), c.build)
```
pathornteng: just do
```
return chainconfig.CheckFeaturesBuildNumber(c.ctx, uint64(blockHeight), c.build)
```

app.go
```diff
@@ -538,6 +539,10 @@ func (a *Application) BeginBlock(req abci.RequestBeginBlock) abci.ResponseBeginB
 			// invalidate cached config so it's reloaded next time it's accessed
 			a.config = nil
 		}
+
+		if err := chainConfigManager.CheckFeaturesBuildNumber(a.height()); err != nil {
```
pathornteng: no need to pass `height`

builtin/plugins/chainconfig/chainconfig.go
```diff
@@ -407,6 +408,23 @@ func HarvestPendingActions(ctx contract.Context, buildNumber uint64) ([]*Action,
...
+	featureRange := ctx.Range([]byte(featurePrefix))
+	for _, m := range featureRange {
+		var f Feature
+		if err := proto.Unmarshal(m.Value, &f); err != nil {
+			return errors.Wrapf(err, "failed to unmarshal feature %s", string(m.Key))
+		}
+		if f.Status != FeatureEnabled {
+			continue
+		}
+		if f.BuildNumber > buildNumber && f.BlockHeight < blockHeight {
```
pathornteng: We don't need to check block height. Just do
```
if f.BuildNumber > buildNumber && f.Status == FeatureEnabled
```
We should also error out the minimum build number that can run on this data

app.go
```diff
@@ -344,6 +346,7 @@ type Application struct {
 	ReceiptsVersion             int32
 }
 
+var checkUnsupportedBuild bool
```
pathornteng: `checkUnsupportedBuild` will be shared across loomchain package. Although this should work fine since it is only used in this file, it is not a good practice as there might be other files in the future unintentionally changing or reusing this variable. So `checkUnsupportedBuild` should be a field of Application struct to make sure that only one application instance can access it.

app.go
```diff
@@ -46,6 +47,8 @@ type State interface {
 	WithContext(ctx context.Context) State
 	WithPrefix(prefix []byte) State
 	SetFeature(string, bool)
+	SetMinBuildNumber(uint64)
+	GetMinBuildNumber() uint64
```
pathornteng: Move `GetMinBuildNumber` to `ReadOnlyState`

plugin/chainconfig_manager.go
```diff
@@ -59,8 +61,19 @@ func (c *ChainConfigManager) EnableFeatures(blockHeight int64) error {
 		}
 		return err
 	}
-	for _, feature := range features {
+
+	supportBuild := c.state.GetMinBuildNumber()
+	if c.build < supportBuild {
```
pathornteng: We don't need to check min build in `EnableFeature`. If the build is not supported, `chainconfig.EnableFeatures` will return error and node panics. Instead we should check the minimum build in `loadApp` by querying data from appstore https://github.com/loomnetwork/loomchain/blob/5ea160014a59315fe44de92c40a07623c1d27500/cmd/loom/loom.go#L727

plugin/chainconfig_manager.go
```diff
@@ -70,6 +83,10 @@ func (c *ChainConfigManager) UpdateConfig() (int, error) {
 	if !c.state.FeatureEnabled(features.ChainCfgVersion1_3, false) {
 		return 0, nil
 	}
+	supportBuild := c.state.GetMinBuildNumber()
+	if c.build < supportBuild {
```
pathornteng: ditto

cmd/loom/loom.go
```diff
@@ -730,6 +733,17 @@ func loadApp(
 		return nil, err
 	}
 
+	if build := appStore.Get(minimumBuildKey); !bytes.Equal(build, []byte{}) {
+		minimumBuild := binary.BigEndian.Uint64(build)
+		currentBuild, err := strconv.ParseUint(loomchain.Build, 10, 64)
+		if err != nil {
```
enlight: I think we should return the error in this case, wrap it first though
```golang
return nil, errors.Wrap(err, "failed to parse loomchain build number")
```

cmd/loom/loom.go
```diff
@@ -730,6 +733,17 @@ func loadApp(
...
+			currentBuild = 0
+		}
+		if currentBuild < minimumBuild {
+			return nil, fmt.Errorf("current build not supported (%d), require (%d)", currentBuild, minimumBuild)
```
enlight:
```golang
return nil, fmt.Errorf("build %d is too old, upgrade to build %d or later", currentBuild, minimumBuild)
```

cmd/loom/loom.go
```diff
@@ -730,6 +733,17 @@ func loadApp(
...
+		}
+	}
+
```
enlight: Also lets add a setting to `loom.yml` to skip this code, we may need to use custom builds sometimes:
```yaml
SkipMinBuildCheck: true
```

PR #1514, opened 2019-10-01
Fix eth_getLogs 

pathornteng:
Currently, tx hashes on returned logs from eth_getLogs do not match the ones from eth_getBlockByNumber. The mismatch can happen if a tx is sent to a Go contract that calls EVM contract and the receipt is generated. The tx hash on receipts is EVM tx hash while it should be TM tx hash. This PR fixes the problem.


eth/query/eth.go
```diff
@@ -95,11 +95,28 @@ func getBlockLogs(
...
+					}
+					txHashList = append(txHashList, hash)
+				}
+			}
 
-			if err != nil {
-				return nil, errors.Wrapf(err, "txhash for block height %d", height)
+			// Try get txHashList from EvmAuxStore instead if txHashList is empty
+			if len(txHashList) == 0 {
+				txHashList, err = evmAuxStore.GetTxHashList(height)
```
enlight: So the issue here seems to be that the tx hash list has duplicate tx hashes because clusterkit didn't fix the hash lists when it recreated the receipts?

pathornteng: Done

eth/query/eth.go
```diff
@@ -95,11 +95,20 @@ func getBlockLogs(
...
 		if MatchBloomFilter(ethFilter, bloomFilter) {
-			txHashList, err := evmAuxStore.GetTxHashList(height)
-
-			if err != nil {
-				return nil, errors.Wrapf(err, "txhash for block height %d", height)
+			var txHashList [][]byte
+
+			// Get txHashList from BlockStore
+			txObject, err := GetBlockByNumber(blockStore, state, int64(height), false, evmAuxStore)
+			if err == nil {
```
enlight: If `err != nil` return an error

store/block_store.go
```diff
@@ -37,10 +37,15 @@ type BlockStore interface {
 }
 
 type MockBlockStore struct {
+	blocks       map[int64]*ctypes.ResultBlock
+	blockResults map[int64]*ctypes.ResultBlockResults
```
enlight: Move this to `mock_block_store.go`

eth/query/eth.go
pathornteng: Done

store/block_store.go
pathornteng: Moved

PR #1517, opened 2019-10-02
Use pull request builder plugin

adyromantika:

Use webhook so that we don't have to poll for changes.


adyromantika:
trigger rebuild

PR #1524, opened 2019-10-04
Complete receipts with data from state.db

pathornteng:

### DIFF
--- a/rpc/query_server.go
+++ b/rpc/query_server.go
@@ -783,10 +783,10 @@ func (s *QueryServer) EthGetTransactionReceipt(hash eth.Data) (*eth.JsonTxReceip
 	if err != nil {
 		if strings.Contains(errors.Cause(err).Error(), "not found") {
 			blockResults, err := s.BlockStore.GetBlockResults(&height)
-			if err != nil || len(blockResults.Results.DeliverTx) <= int(txReceipt.TransactionIndex) {
-				return nil, nil
-			}
-			if blockResults.Results.DeliverTx[txReceipt.TransactionIndex] == nil {
+			if err != nil ||
+				blockResults == nil ||
+				len(blockResults.Results.DeliverTx) <= int(txReceipt.TransactionIndex) ||
+				blockResults.Results.DeliverTx[txReceipt.TransactionIndex] == nil {
 				return nil, nil
 			}
 			return completeReceipt(
--- a/rpc/query_server.go
+++ b/rpc/query_server.go
@@ -1187,18 +1187,16 @@ func completeReceipt(txResults *ctypes.ResultTx, blockResult *ctypes.ResultBlock
 		}
 	}
 
+	if txResults.TxResult.Code == abci.CodeTypeOK {
+		txReceipt.Status = StatusTxSuccess
+	} else {
+		txReceipt.Status = StatusTxFail
+	}
+
 	jsonReceipt := eth.EncTxReceipt(*txReceipt)
-	if txResults != nil {
-		if txResults.TxResult.Code == abci.CodeTypeOK {
-			txReceipt.Status = StatusTxSuccess
-		} else {
-			txReceipt.Status = StatusTxFail
-		}
-		jsonReceipt = eth.EncTxReceipt(*txReceipt)
-		if txResults.TxResult.Info == utils.CallEVM && (jsonReceipt.To == nil || len(*jsonReceipt.To) == 0) {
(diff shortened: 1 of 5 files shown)


rpc/query_server.go
```diff
@@ -1167,21 +1167,25 @@ func getReceiptByTendermintHash(
 }
 
 func completeReceipt(txResults *ctypes.ResultTx, blockResult *ctypes.ResultBlock, txReceipt *types.EvmTxReceipt) *eth.JsonTxReceipt {
+	jsonReceipt := eth.EncTxReceipt(*txReceipt)
```
enlight: Moving it up here means the block time won't be set in the returned receipt, doesn't it?

rpc/query_server.go
```diff
@@ -782,8 +782,8 @@ func (s *QueryServer) EthGetTransactionReceipt(hash eth.Data) (*eth.JsonTxReceip
 	txResults, err := s.BlockStore.GetTxResult(blockResult.Block.Data.Txs[txReceipt.TransactionIndex].Hash())
 	if err != nil {
 		if strings.Contains(errors.Cause(err).Error(), "not found") {
-			// return nil response if cannot find hash
-			return nil, nil
+			// return receipts without blockresult info
```
enlight: If `blockStore.GetTxResult` doesn't have the data try `blockStore.GetBlockResults`.

rpc/query_server.go
pathornteng: Move it back

rpc/query_server.go
pathornteng: Done

rpc/query_server.go
```diff
@@ -782,8 +782,12 @@ func (s *QueryServer) EthGetTransactionReceipt(hash eth.Data) (*eth.JsonTxReceip
 	txResults, err := s.BlockStore.GetTxResult(blockResult.Block.Data.Txs[txReceipt.TransactionIndex].Hash())
 	if err != nil {
 		if strings.Contains(errors.Cause(err).Error(), "not found") {
-			// return nil response if cannot find hash
-			return nil, nil
+			txResults, err := s.BlockStore.GetTxResultByHeightAndIndex(&height, int(txReceipt.TransactionIndex))
+			if err != nil {
+				// return receipts without blockresult info
+				return completeReceipt(nil, blockResult, &txReceipt), nil
```
enlight: I think we should return nil in this case, like we used to. Creating a receipt with an inaccurate status seems worse than just returning nothing.

store/block_store_lru_cache.go
```diff
@@ -131,3 +132,24 @@ func (s *LRUBlockStoreCache) GetTxResult(txHash []byte) (*ctypes.ResultTx, error
 	}
 	return txResult, nil
 }
+
+func (s *LRUBlockStoreCache) GetTxResultByHeightAndIndex(height *int64, index int) (*ctypes.ResultTx, error) {
```
enlight: What's the point of this function, it doesn't check the cache, nor update it.

store/block_store.go
```diff
@@ -150,6 +156,27 @@ func (s *TendermintBlockStore) GetTxResult(txHash []byte) (*ctypes.ResultTx, err
...
+		return nil, ErrIndexOutOfRange
+	}
+
+	return &ctypes.ResultTx{
+		Index:  uint32(index),
+		Height: *height,
+		TxResult: abci.ResponseDeliverTx{
+			Code: blockResult.Results.DeliverTx[index].Code,
+			Data: blockResult.Results.DeliverTx[index].Data,
+			Info: blockResult.Results.DeliverTx[index].Info,
```
enlight: Remember the whole `Results.DeliverTx[index] == nil` incident?

store/block_store.go
```diff
@@ -150,6 +156,27 @@ func (s *TendermintBlockStore) GetTxResult(txHash []byte) (*ctypes.ResultTx, err
 	}, nil
 }
 
+func (s *TendermintBlockStore) GetTxResultByHeightAndIndex(height *int64, index int) (*ctypes.ResultTx, error) {
```
enlight: Adding this to the block store seems premature

rpc/query_server.go
pathornteng: Done

store/block_store.go
pathornteng: Check nil before dereferencing

store/block_store.go
pathornteng: Removed

Issue #1498, opened 2019-09-24
Add EVM builtin function to get ed25519 address mapped to secp256k1 address


enlight:
We should provide access to account mappings within Solidity contracts so devs can verify signed messages. Currently the address returned by `ecrecover` in a Solidity contract can't be matched to any DAppChain account (nor `msg.sender`).

It would look something like this:
```solidity
// get the address matching the secp256k1 private key that signed the message
address signer = ecrecover(_message, _v, _r, _s);
// get the DAppChain address of the signer
address account = loom_native_api.mappedAccount('eth', signer);
```

This would need to be implemented as a precompiled EVM contract, we've got an example of doing so in https://github.com/loomnetwork/loomchain/blob/1521b61bc8b4af38a0d9ebc552886f858f006bb3/evm/evm_test.go#L71 and of course `ecrecover` and friends https://github.com/ethereum/go-ethereum/blob/master/core/vm/contracts.go

enlight:
Going the other way (DAppChain address -> Eth address) would also be useful:
```solidity
// lookup the mapping of the DAppChain address of msg.sender to an Eth address
address ethAccount = loom_native_api.mappedAccount('loom', msg.sender, 'eth')
address signer = ecrecover(_message, _v, _r, _s);
// can now compare ethAccount to signer to verify msg.sender signed the message
```

Sriep:
Implemented in https://github.com/loomnetwork/loomchain/pull/1526

PR #1526, opened 2019-10-04
Add Solidity libary function to get ed25519 address mapped to secp256k1 address

Sriep:

Addresses https://github.com/loomnetwork/loomchain/issues/1498
Adds new LoomNativeApi Solidity libaray

```solidity
library LoomNativeApi {
    function MappedAccount(string memory from, address addr) public view returns (address) 
    function MappedAccount(string memory from, address addr, string memory to) public view returns (address) 
}
```


Sriep:
Adds new `loomchain/evm/precompiles` directory with new `EvmPrecompilerHandler` interface.
The `LoomNativeApi` library is at `loomchain/e2e/tests/truffle/contracts/LoomNativeApi.sol`.
Referenced https://blog.ricmoo.com/verifying-messages-in-solidity-50a94f82b2ca for striping prefix in truffle test.

e2e/tests/truffle/contracts/LoomNativeApi.sol
```diff
@@ -0,0 +1,61 @@
...
+
+    function callPFAssembly(address _addr, bytes memory _input, uint256 outSize) view internal returns (bytes20)    {
+        uint256 inSize = _input.length*4+1;
+        uint256 inLenght = _input.length;
+        bytes20 rtv;
+        assembly{
+            let start := add(_input, 0x20)
+            let p := mload(0x40)
+            mstore(p, inSize)
+            if iszero(staticcall(
```
enlight: Got a link to some docs or example that explains how this works?

Sriep: https://medium.com/@rbkhmrcr/precompiles-solidity-e5d29bd428c4 is the main guide I used, could not find much else.
Also, https://solidity.readthedocs.io/en/v0.5.12/assembly.html is a reference of the Solidity assembly commands, nothing about precompiles.

Sriep: `call(g, a, v, in, insize, out, outsize)` call contract at address a with input mem[in…(in+insize)) providing g gas and v wei and output area mem[out…(out+outsize)) returning 0 on error (eg. out of gas) and 1 on success


`staticcall(g, a, in, insize, out, outsize)`  identical to call(g, a, 0, in, insize, out, outsize) but do not allow state modifications

Where a = address is the number of our new precompiled functions. Need to take care to get the input and output data setup correctly.

e2e/tests/truffle/contracts/LoomNativeApi.sol
```diff
@@ -0,0 +1,73 @@
...
+            input[i] = byte(uint8(uint(addr) / (2**(8*(19 - i)))));
+        }
+        input[20] = byte(uint8(fromB.length));
+        for (uint j = 0; j < fromB.length; j++) {
+            input[21+j] = fromB[j];
+        }
+        for (uint k = 0; k < toB.length; k++) {
+            input[21+fromB.length+k] = toB[k];
+        }
+        return address(callPFAssembly(MapToAddress, input, 0x14));
```
enlight: Can you make `0x14` a named constant?

evm/precompiles/loomprecompiles/addressmapping.go
```diff
@@ -0,0 +1,115 @@
...
+		createAddressMapperCtx: createAddressMapperCtx,
+	}
+}
+
+type mapToLoomAddress struct {
+	state                  loomchain.State
+	createAddressMapperCtx func(loomchain.State) (contractpb.StaticContext, error)
+}
+
+func (ma mapToLoomAddress) RequiredGas(input []byte) uint64 {
```
enlight: Use a pointer receiver

evm/precompiles/loomprecompiles/addressmapping.go
```diff
@@ -0,0 +1,115 @@
...
+	return uint64(0)
+}
+
+func (ma mapToLoomAddress) Run(input []byte) ([]byte, error) {
```
enlight: Use a pointer receiver

evm/precompiles/loomprecompiles/addressmapping.go
```diff
@@ -0,0 +1,115 @@
...
+		createAddressMapperCtx: createAddressMapperCtx,
+	}
+}
+
+type mapToAddress struct {
+	state                  loomchain.State
+	createAddressMapperCtx func(loomchain.State) (contractpb.StaticContext, error)
+}
+
+func (ma mapToAddress) RequiredGas(input []byte) uint64 {
```
enlight: Use a pointer receiver

evm/precompiles/loomprecompiles/addressmapping.go
```diff
@@ -0,0 +1,115 @@
...
+	return uint64(0)
+}
+
+func (ma mapToAddress) Run(input []byte) ([]byte, error) {
```
enlight: Use a pointer receiver

evm/precompiles/loomprecompiles/addressmapping.go
```diff
@@ -0,0 +1,115 @@
...
+	strI := string(input)
+	_ = strI
```
enlight: Can remove this I think

evm/precompiles/loomprecompiles/addressmapping.go
```diff
@@ -0,0 +1,115 @@
...
+
+	if len(input) < 23 || uint(len(input)) < 22+uint(input[20]) {
+		return nil, errors.Errorf("mapBetweenAccounts input too short %x", input)
+	}
+	local := input[:20]
+	chainFromLen := uint(input[20])
+	chainFrom := string(input[21 : 21+chainFromLen])
+	chainTo := string(input[21+chainFromLen:])
```
enlight: Can we combine `mapToLoomAddress` and `mapToAddress` into one precompile? The only difference is whether or not the `to` chain ID is passed in which you can detect by checking for `len(input[21+chainFromLen:]) > 0`

evm/precompiles/loomprecompiles/testdata/TestLoomNativeApi.sol
```diff
@@ -0,0 +1,82 @@
...
+contract TestLoomNativeApi {
+    function TestMappedLoomAccount(string memory ethChainId, bytes32 hash, uint8 v, bytes32 r, bytes32 s) public view returns (address) {
+        return MappedAccount(ethChainId, ecrecover(hash, v, r, s));
+    }
+
+    function TestMappedAccount(string memory ethChainId, bytes32 hash, uint8 v, bytes32 r, bytes32 s, string memory toChain) public view returns (address) {
+        return MappedAccount(ethChainId, ecrecover(hash, v, r, s), toChain);
+    }
+
+    // LoomApi library
```
enlight: Why is this library code duplicated here?

cmd/loom/loom.go
```diff
@@ -833,7 +834,14 @@ func loadApp(
 					return nil, err
 				}
 			}
-			return evm.NewLoomVm(state, eventHandler, receiptHandlerProvider.Writer(), createABM, cfg.EVMDebugEnabled), nil
+			return evm.NewLoomVm(
+				state,
+				eventHandler,
+				receiptHandlerProvider.Writer(),
+				createABM,
+				loomprecompiles.NewLoomPrecompileHandler(getContractStaticCtx("addressmapper", vmManager)),
```
enlight: The new precompiles should be behind a feature flag, so need to call `state.FeatureEnabled` here, if it's not enabled don't pass through the precompiles.

e2e/tests/truffle/contracts/TestLoomNativeApi.sol
```diff
@@ -0,0 +1,52 @@
+pragma solidity >=0.4.21;
+
+import "./LoomNativeApi.sol";
+
+library TestLoomNativeApi {
+    function TestMappedLoomAccount(string memory fromChain, string memory hash, uint8 v, bytes32 r, bytes32 s) public view returns (address) {
+        return LoomNativeApi.MappedAccount(fromChain, ecrecover(stripPrefix(hash), v, r, s));
```
enlight: I suspect the reason you've had to add this `stripPrefix` thing here is that you're signing a hash of the message instead of the message itself:
```js
testHash = web3js.utils.sha3(msg);
sig = await wallet.signMessage(testHash)
```

If you did this instead I don't think you'll need to use `stripPrefix`:
```js
sig = await wallet.signMessage(msg);
testHash = web3js.utils.soliditySha3('\x19Ethereum Signed Message:\n32', msg);
```

e2e/tests/truffle/contracts/LoomNativeApi.sol
```diff
@@ -0,0 +1,73 @@
...
+
+    // Calls MapToLoomAddress precompiled EVM function.this.
+    // Returns the loom address mapped to the input local addr and fromChainId..
+    function MappedAccount(string memory fromChainId, address addr) public view returns (address) {
+        bytes memory fromB = bytes(fromChainId);
+        bytes memory input = new bytes(fromB.length + 20);
+        // Encode from chain id and local address into bytes object for passing to precompied function
+        // [<addr - 20 bytes>, <from chain id, rest of array>]"
+        for (uint i = 0; i < 20; i++) {
+            input[i] = byte(uint8(uint(addr) / (2**(8*(19 - i)))));
```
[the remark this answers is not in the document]
Sriep: https://ethereum.stackexchange.com/questions/884/how-to-convert-an-address-to-bytes-in-solidity
`function suggested by Solidity's author, chriseth:`

e2e/tests/truffle/contracts/TestLoomNativeApi.sol
enlight: Actually we just spent a bunch of time doing something very similar today, here's how to get it working:
```js
sig = await wallet.signMessage(msg);
testHash = ethers.utils.hashMessage(msg);
```

Then you just pass the `testHash` to the Solidity contract and do:
```solidity
ecrecover(testHash, v, r, s)
```

evm/precompiles/loomprecompiles/testdata/TestLoomNativeApi.sol
Sriep: Bearing in mind that the point is to test `addressmapping.go` and not the Solidity code, I found it easier to 'statically link' the LoomNativeApi and the test functions in this fashion rather than dynamically link the libraries. Dynamically linking the libraries involves modifying the TestLoomNativeApi.sol binary at runtime with the address of the library contract in the active EVM.

Sriep: I could add some utility code to allow dynamically linking Solidity libraries in unit tests.

PR #1532, opened 2019-10-09
Allow EVM state to be flushed to disk every Nth block instead of every block

pathornteng:
Currently, we flush Patricia trie every tx. This commit operation is really expensive and it causes a lot of IOPS. Furthermore, once the trie is committed to disk, web3/truffle normally queries the newly committed data so the app loads the recently committed trie from disk shortly after. 

This PR makes commit more efficient by holding multiple tries in memory and only flushing them every N blocks. This should significantly reduce the disk usage (less trie versions) and also IOPS (recent tries cached in memory)

Dependency: https://github.com/loomnetwork/go-ethereum/pull/9


app.go
```diff
@@ -829,6 +860,39 @@ func (a *Application) Commit() abci.ResponseCommit {
...
+
+	flushInterval := a.FlushInterval
+	if flushInterval == 0 {
+		cfg := state.Config()
+		flushInterval = int64(cfg.GetAppStore().GetIAVLFlushInterval())
+	} else if flushInterval == -1 {
+		flushInterval = 0
+	}
+
+	// Only commit Patricia tree every N blocks
```
pathornteng: This should be synchronised with IAVLFlushInterval

app.go
```diff
@@ -39,6 +41,7 @@ type ReadOnlyState interface {
 	Config() *cctypes.Config
 	EnabledFeatures() []string
 	GetMinBuildNumber() uint64
+	GetTrieDB() *trie.Database
```
enlight: I don't what the `trie.Database` exposed via `ReadOnlyState`/`State`

pathornteng: How about `EVMStateDB`?

app.go
```diff
@@ -41,7 +42,7 @@ type ReadOnlyState interface {
 	Config() *cctypes.Config
 	EnabledFeatures() []string
 	GetMinBuildNumber() uint64
-	GetTrieDB() *trie.Database
+	GetEVMStateDB() gstate.Database
```
enlight: Just how we don't expose the IAVL store via ReadOnlyState/State we shouldn't be exposing the low level EVM DB.

pathornteng: K, I will move it to `state`. This is not a low level database. EVMStateDB is a wrapper of loomState -> EthDB -> TrieDB -> EVMStateDB.

pathornteng:
According to my own performance test, this PR
1. Reduce the growth rate around 50%
2. Reduce processing time for block replay around 45%

rpc/query_server.go
```diff
@@ -327,7 +328,7 @@ func (s *QueryServer) EthGetCode(address eth.Data, block eth.BlockHeight) (eth.D
 	snapshot := s.StateProvider.ReadOnlyState()
 	defer snapshot.Release()
 
-	evm := levm.NewLoomVm(snapshot, nil, nil, nil, false)
+	evm := levm.NewLoomVm(snapshot, s.EvmStore, nil, nil, nil, false)
```
enlight: This doesn't look right to me, the EVM created to handle this query will access the same `trie.Database` as the one being written to in the tx/middleware handlers, the `EvmStore.trieDB` isn't a snapshot. The `trie.Database` looks thread-safe, but that may not be sufficient to guarantee correctness. Also the `EvmStore.trieDB` should probably be accessed via atom ops like the `MultiWriterAppStore.lastSavedTree`.

pathornteng:
rebuild sesame

store/evmstore.go
```diff
@@ -315,15 +311,20 @@ func (s *EvmStore) GetSnapshot(version int64) db.Snapshot {
...
 
 func NewEvmStoreSnapshot(snapshot db.Snapshot, rootHash []byte) *EvmStoreSnapshot {
-	return &EvmStoreSnapshot{
+	evmSnapshot := &EvmStoreSnapshot{
 		Snapshot: snapshot,
 		rootHash: rootHash,
 	}
+	ethDB := NewLoomEthDB(evmSnapshot, nil)
+	// This should be read only database
+	evmSnapshot.trieDB = trie.NewDatabase(ethDB)
```
enlight: The evm.db snapshot will only contain the state as of the last flushed block, so this `EvmStoreSnapshot` will not have any state changes made in-memory since the last flush to disk.

rpc/query_server.go
```diff
@@ -268,7 +269,9 @@ func (s *QueryServer) queryEvm(state loomchain.State, caller, contract loom.Addr
 			return nil, err
 		}
 	}
-	vm := levm.NewLoomVm(state, nil, nil, createABM, false)
+	version := state.Block().Height
+	evmSnapshot := s.EvmStore.GetSnapshot(version).(*store.EvmStoreSnapshot)
```
enlight: This casting looks silly, why doesn't `EvmStore.GetSnapshot()` return `*store.EvmStoreSnapshot` in the first place?

store/evmstore.go
pathornteng: Yeah I know, it is a snapshot so the uncommitted state should be excluded

pathornteng: Oh hmmmmm I see what you mean. Probably need to find a way to fix this.

pathornteng: Probably need to implement this committing state once per block https://github.com/loomnetwork/loomchain/issues/1568 in this PR.

rpc/query_server.go
pathornteng: Fixed

pathornteng:
rebuild sesame

pathornteng:
rebuild sesame

app.go
```diff
@@ -823,19 +840,28 @@ func (a *Application) deliverTx2(storeTx store.KVStoreTx, txBytes []byte) abci.R
...
 		commitBlockLatency.With(lvs...).Observe(time.Since(begin).Seconds())
 	}(time.Now())
+
+	// Commit EVM state
+	evmStateRoot, err := a.EVMState.Commit(true)
+	if err != nil {
+		panic(err)
+	}
+	a.EvmStore.SetVMRootKey(evmStateRoot[:])
+	a.EVMState.Reset(evmStateRoot)
```
enlight: Why is this stuff here instead of `EvmStore.Commit`?

app.go
```diff
@@ -368,6 +384,7 @@ type Application struct {
 	config                      *cctypes.Config
 	childTxRefs                 []evmaux.ChildTxRef // links Tendermint txs to EVM txs
 	ReceiptsVersion             int32
+	EVMState                    *gstate.StateDB
```
enlight: Does this `StateDB` instance really need to be in `Application`, why can't the `EvmStore` own it?

app.go
```diff
@@ -911,5 +942,5 @@ func (a *Application) ReadOnlyState() State {
 		a.lastBlockHeader,
 		nil, // TODO: last block hash!
 		a.GetValidatorSet,
-	)
+	).WithEVMState(snapshotEVMState)
```
enlight: This should probably be something like:
```golang
).WithEVMState(a.EvmStore.GetSnapshot().StateDB())
```
Also `a.EvmStore.GetVMRootKey()` as currently implemented is not go-routine-safe.

app.go
pathornteng: `EvmStore` is at lower level. `EVMState` is a layer on top of `EVMStore` just like `State` is on top of  AppStore.

app.go
pathornteng: The name is misleading. StateDB is actually a state not DB. The wrapper is like StateDB -> TrieDB -> EvmStore.

evm/loomevm.go
```diff
@@ -209,18 +183,15 @@ func (lvm LoomVm) Create(caller loom.Address, code []byte, value *loom.BigUInt)
...
 	levm, err := NewLoomEvm(lvm.state, lvm.accountBalanceManager(false), logContext, lvm.debug)
 	if err != nil {
 		return nil, err
 	}
+	snapshot := lvm.state.EVMState().Snapshot()
 	_, err = levm.Call(caller, addr, input, value)
-	if err == nil {
-		_, err = levm.Commit()
+	if err != nil {
+		lvm.state.EVMState().RevertToSnapshot(snapshot)
```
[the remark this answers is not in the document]
pathornteng: For `create`, evm calls SetNonce even though the tx reverts. So this is necessary or we will get consensus failure. https://github.com/loomnetwork/go-ethereum/blob/6128fa1a8c767035d3da6ef0c27ebb7778ce3713/core/vm/evm.go#L366

evm/loomevm.go
```diff
@@ -230,6 +201,7 @@ func (lvm LoomVm) Call(caller, addr loom.Address, input []byte, value *loom.BigU
 			events = lvm.receiptHandler.GetEventsFromLogs(
 				levm.sdb.Logs(), lvm.state.Block().Height, caller, addr, input,
 			)
+			levm.sdb.ResetLogs()
```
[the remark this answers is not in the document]
pathornteng: Logs get clear every `State.Commit` but since now state is shared by many txs, we have to clear ou logs or we will end up having duplicates logs if there are multiple txs in one block.

store/evmstore.go
```diff
@@ -270,10 +319,11 @@ func (s *EvmStore) GetSnapshot(version int64) db.Snapshot {
 }
 
 func NewEvmStoreSnapshot(snapshot db.Snapshot, rootHash []byte) *EvmStoreSnapshot {
-	return &EvmStoreSnapshot{
+	evmSnapshot := &EvmStoreSnapshot{
 		Snapshot: snapshot,
 		rootHash: rootHash,
 	}
+	return evmSnapshot
```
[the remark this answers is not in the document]
pathornteng: accidentally

store/evmstore.go
```diff
@@ -240,6 +276,19 @@ func (s *EvmStore) Version() ([]byte, int64) {
...
+	return s.trieDB
+}
+
+// GetVMRootKey returns the lastest root of EVM state trie
+func (s *EvmStore) GetVMRootKey() []byte {
+	return s.rootHash
+}
+
+func (s *EvmStore) SetVMRootKey(root []byte) {
+	s.Set(util.PrefixKey(vmPrefix, rootKey), root)
```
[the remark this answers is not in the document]
pathornteng: true, we don't need to set it to DB anymore.

app.go
enlight: Yeah, so maybe we should have something like this?
```golang
struct EVMState {
  sdb StateDB
  evmStore *EvmStore

  Commit() error {
    evmStateRoot, err := sdb.Commit(true)
    if err != nil {
      return nil, err
    }
    evmStore.SetVMRootKey(evmStateRoot[:])
    sdb.Reset(evmStateRoot)
  }

  GetSnapshot(version int64) (*EVMState, error) {
    evmStateRoot, err := evmStore.GetRootAt(version)
    if err != nil {
      return err
    }
    stateDB, err := gstate.New(gcommon.BytesToHash(evmStateRoot), sdb.Database())
    if err != nil {
      return nil, err
    }
    return &EVMState{ sdb: stateDB, evmStore: evmStore }
  }
}
```

app.go
```diff
@@ -911,5 +936,5 @@ func (a *Application) ReadOnlyState() State {
 		a.lastBlockHeader,
 		nil, // TODO: last block hash!
 		a.GetValidatorSet,
-	)
+	).WithEVMState(a.EvmStore.GetSnapshot(a.lastBlockHeader.Height).StateDB(a.EVMState.Database()))
```
enlight: Should call `s.EvmStore.GetSnapshot(storeStateSnapshot.Block().Height)` to ensure both snapshots will be for the same height/version.

cmd/loom/loom.go
```diff
@@ -836,8 +842,18 @@ func loadApp(
...
+
+		ethDB := store.NewLoomEthDB(evmStore, nil)
+		evmRoot, _ := evmStore.Version()
+		stateDB := state.NewDatabase(ethDB)
+		stateDB.SetTrieDB(evmStore.TrieDB())
+		sdb, err := state.New(gcommon.BytesToHash(evmRoot), stateDB)
+		if err != nil {
+			return nil, err
+		}
+		evmState = loomchain.NewEVMState(evmStore, sdb)
```
enlight: This stuff is pretty confusing, I think it'd be better to move this into `NewEVMState`:
```golang
evmState = loomchain.NewEVMState(evmStore)
```

Which will then look something like this:
```golang
func NewEVMState(evmStore *store.EvmStore) (*EVMState, error) {
  stateDatabase := state.NewDatabaseWithTrieDB(evmStore.TrieDB())
  evmRoot, _ := evmStore.Version()
  sdb, err := state.New(gcommon.BytesToHash(evmRoot), stateDatabase)
  if err != nil {
    return nil, err
  }
  return &EVMState{
    evmStore: evmStore,
    sdb:      sdb,
  }
}
```

app.go
```diff
@@ -903,6 +967,10 @@ func (a *Application) height() int64 {
 }
 
 func (a *Application) ReadOnlyState() State {
+	evmStateSnapshot, err := a.EVMState.GetSnapshot(a.lastBlockHeader.Height)
```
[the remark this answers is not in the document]
pathornteng: done

cmd/loom/loom.go
pathornteng: I can only do this.
```
func NewEVMState(evmStore *store.EvmStore) (*EVMState, error) {
	ethDB := store.NewLoomEthDB(evmStore, nil)
	stateDB := state.NewDatabase(ethDB).WithTrieDB(evmStore.TrieDB())
	evmRoot, _ := evmStore.Version()
	sdb, err := state.New(gcommon.BytesToHash(evmRoot), stateDB)
	if err != nil {
		return nil, err
	}
	return &EVMState{
		evmStore: evmStore,
		sdb:      sdb,
	}, nil
}
```

pathornteng:
rebuild sesame

registry/registry.pb.go
```diff
@@ -32,7 +32,7 @@ func (m *Record) Reset()         { *m = Record{} }
 func (m *Record) String() string { return proto.CompactTextString(m) }
 func (*Record) ProtoMessage()    {}
 func (*Record) Descriptor() ([]byte, []int) {
-	return fileDescriptor_registry_8f33b09e6c22f9ce, []int{0}
+	return fileDescriptor_registry_db594c69f58cdc2f, []int{0}
```
enlight: These changes are irrelevant in this PR, revert them.

app.go
```diff
@@ -914,7 +873,7 @@ func (a *Application) Commit() abci.ResponseCommit {
 
 	// Update the last block header before emitting events in case the subscribers attempt to access
 	// the latest committed state as soon as they receive an event.
-	a.lastBlockHeader = a.curBlockHeader
+	atomic.StorePointer(&a.lastBlockHeader, unsafe.Pointer(&a.curBlockHeader))
```
enlight: This doesn't look right, I'm pretty sure `&a.curBlockHeader` is going to remain the same, meaning the last block header will always reference the current block header.

pathornteng: good catch

evm/loomevm.go
```diff
@@ -54,44 +49,24 @@ type LoomEvm struct {
...
-		return nil, err
-	}
-
+	p := &LoomEvm{}
 	var abm *evmAccountBalanceManager
+	var err error
 	if accountBalanceManager != nil {
 		abm = newEVMAccountBalanceManager(accountBalanceManager, loomState.Block().ChainID)
-		p.sdb, err = newLoomStateDB(abm, common.BytesToHash(oldRoot), state.NewDatabase(p.db))
+		p.sdb, err = newLoomStateDB(abm, loomState.EVMState().StateDB())
```
enlight: `LoomEvm` no longer reads or writes from the `MultiWriterAppStore` because the `StateDB` here no longer wraps `loomchain.State`, so we can eliminate all the reads & writes of `EvmStore` from `MultiWriterAppStore`, right?

pathornteng: yes

evm_state.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// GetSnapshot returns the EVMState instance containing the state as it was at the given version.
+// NOTE: Do not call Commit on the returned instance.
+func (s *EVMState) GetSnapshot(version int64) (*EVMState, error) {
+	stateDB, err := state.New(common.BytesToHash(s.evmStore.GetRootAt(version)), s.sdb.Database())
+	if err != nil {
+		return nil, err
+	}
+	return &EVMState{sdb: stateDB, evmStore: s.evmStore}, nil
```
enlight: Writing/reading the `EvmStore` is not thread-safe (mostly due to the cache), the EvmStore snapshots can be safely read while the store is being written to, but since we don't use `EvmStore.GetSnapshot` here the new `EVMState` will allow the QueryServer to read from the `EvmStore` while the Application writes to it.

pathornteng:
rebuild sesame

e2e/tests/truffle/test/EvmSnapshot.js
```diff
@@ -0,0 +1,92 @@
...
+            // a local address (not an eth address)
+            {
+                from: localAddr.local.toString()
+            }
+        );
+    })
+    it('SnapshotTest', async () => {
+        for (var i = 0; i < 50; i++) {
+            contract.methods.set(7777).send().then()
+            contract.methods.get().call().then()
```
enlight: I think `then()` is unnecessary since you don't need to wait for the promise to be resolved.

store/evmstore.go
```diff
@@ -236,23 +190,11 @@ func (s *EvmStore) Commit(version int64) []byte {
...
-	batch := s.evmDB.NewBatch()
-	for key, item := range s.cache {
-		if !item.Deleted {
-			batch.Set([]byte(key), item.Value)
-		} else {
-			batch.Delete([]byte(key))
-		}
-	}
-	batch.Write()
-	s.cache = make(map[string]cacheItem)
```
enlight: Hmm, so no cache is fine, but no batching less so. Should still batch the writes to the DB.

e2e/tests/truffle/test/EvmSnapshot.js
pathornteng: Yes we don't wait for the promise to be resolved but `then()` is necessary here or the promise will never be resolved.

store/evmstore.go
pathornteng: We only write to evmstore in `TrieDB.Commit()` which already uses ethDB.batch so it is fine. https://github.com/loomnetwork/go-ethereum/blob/6128fa1a8c767035d3da6ef0c27ebb7778ce3713/trie/database.go#L620

enlight: You're right, there is already batching in `LoomEthDB`, but it's crap because each key in the batch is written out individually to `evm.db` by the `EvmStore`. Should make `EvmStore` expose a batch API so `LoomEthDB` can batch writes properly.

store/loomethdb.go
```diff
@@ -111,13 +116,15 @@ func (b *batch) Write() error {
 		return bytes.Compare(b.cache[j].key, b.cache[k].key) < 0
```
enlight: This cache is no longer needed, should just write to the batch directly. Only the `LogBatch` needs the cache & sorting to ensure log entries are written out in deterministic order.

store/loomethdb.go
```diff
@@ -192,7 +184,7 @@ func (s *LoomEthdb) NewLogBatch(logContext *ethdbLogContext) ethdb.Batch {
 	if !loggerStarted {
 		file, err := os.Create(b.params.LogFilename)
 		if err != nil {
-			return &b.batch
+			return b
```
enlight: This error handling doesn't make much sense to me

store/loomethdb.go
```diff
@@ -146,9 +133,16 @@ type LogParams struct {
...
+	value []byte
+}
 type LogBatch struct {
-	batch  batch
-	params LogParams
+	parentStore *LoomEthDB
+	size        int
+	params      EthDBLogParams
+	cache       []kvPair
+	lock        sync.RWMutex
```
enlight: This lock appears to be completely useless, there's only ever going to be a single writer, and the lock isn't used properly in the batch anyway.

store/loomethdb.go
```diff
@@ -11,79 +9,92 @@ import (
...
-type kvPair struct {
-	key   []byte
-	value []byte
-}
 
 type batch struct {
-	cache       []kvPair
-	parentStore *LoomEthdb
-	size        int
+	levelDBbatch dbm.Batch
```
enlight: `levelDBbatch` -> `dbBatch`

store/loomethdb.go
```diff
@@ -93,46 +104,22 @@ func (b *batch) ValueSize() int {
...
-	})
-
-	for _, kv := range b.cache {
-		if kv.value == nil {
-			b.parentStore.Delete(kv.key)
-		} else {
-			b.parentStore.Put(kv.key, kv.value)
-		}
-	}
+	b.levelDBbatch.WriteSync()
```
enlight: This wasn't a sync write in the EvmStore, so probably shouldn't be sync here either.

store/loomethdb.go
```diff
@@ -241,21 +241,41 @@ func (b *LogBatch) Write() error {
...
+	for _, kv := range b.cache {
+		if kv.value == nil {
+			levelDBbatch.Delete(util.PrefixKey(vmPrefix, kv.key))
+		} else {
+			levelDBbatch.Set(util.PrefixKey(vmPrefix, kv.key), kv.value)
+		}
 	}
-	err := b.batch.Write()
+	b.lock.Unlock()
+	levelDBbatch.WriteSync()
```
enlight: See comment on `batch.Write`

store/loomethdb.go
pathornteng: panic instead then

store/loomethdb.go
pathornteng: Yeah I think so but did not want to touch it in the first place. Removed

store/iavlstore.go
```diff
@@ -216,6 +216,13 @@ func (s *IAVLStore) GetSnapshot() Snapshot {
 	}
 }
 
+func (s *IAVLStore) GetSnapshotAt(version int64) (Snapshot, error) {
+	// This isn't an actual snapshot obviously, and never will be, but lets pretend...
+	return &iavlStoreSnapshot{
+		IAVLStore: s,
+	}, nil
+}
```
enlight: Actually, this should probably just `panic("not implemented")`

pathornteng:
rebuild sesame

enlight:
Previously the EVM accessed data like so `EVM -> VersionedCachingStore -> MultiWriterAppStore -> EvmStore -> evm.db`, this PR changes it to `EVM -> EVMState -> EvmStore -> evm.db`, which means that the `VersionedCachingStore` is completely bypassed for EVM reads & writes, leaving only LevelDB caching.

Newer versions of go-ethereum (https://github.com/ethereum/go-ethereum/releases/tag/v1.8.19 or later) have added a cache to the `trie.Database` (https://github.com/ethereum/go-ethereum/pull/18087) created by the `EvmStore`, this trie cache should compensate for the loss of `VersionedCachingStore` so we should update our go-ethereum fork.

https://github.com/ethereum/go-ethereum/releases/tag/v1.8.19 is the release in which the trie cache was initially added, it was optimized in https://github.com/ethereum/go-ethereum/releases/tag/v1.9.0 (https://github.com/ethereum/go-ethereum/pull/19307), and then again in https://github.com/ethereum/go-ethereum/releases/tag/v1.9.8 (https://github.com/ethereum/go-ethereum/pull/19971). Need to figure out if we can just bump our go-ethereum to v1.9.8.

enlight:
There's also https://github.com/ethereum/go-ethereum/pull/19953 that was merged in go-ethereum v1.9.4 that may impact the `trie.Database` performance, will need to review it to make sure our usage of `trie.Database` is compatible with the changes in that PR.

enlight:
In a follow up PR:

- Eliminate call to `EvmStore.GetSnapshot()` from `MultiWriterAppStore.GetSnapshotAt()`. In fact we can probably eliminate `GetSnapshot()` from all the stores, only `GetSnapshotAt()` is being used now.
- Eliminate writes to the `EvmStore` from `MultiWriterAppStore`.
- Eliminate `Get/Has/Set/Delete` from `EvmStore`.
- Fix `VersionedCachingStore.GetSnapshotAt()` to be atomic.
- Remove the hack from `VersionedCachingStore.SaveVersion()` and fix `Application.Commit()` instead. This should be done last probably since it's not that important.

enlight:
> Previously the EVM accessed data like so `EVM -> VersionedCachingStore -> MultiWriterAppStore -> EvmStore -> evm.db`, this PR changes it to `EVM -> EVMState -> EvmStore -> evm.db`, which means that the `VersionedCachingStore` is completely bypassed for EVM reads & writes, leaving only LevelDB caching.

Actually this data access flow is only accurate when considering queries. Tx handlers never read from `VersionedCachingStore`, they only write to it, which means that the loss of the `VersionedCachingStore` shouldn't affect performance of validator and sentry nodes that aren't queried often... in fact we probably should've disabled the `VersionedCachingStore` by default on all those nodes a long time ago to save a bit of CPU & RAM.

Issue #1538, opened 2019-10-16
Blocktime isn't available on events from JSON RPC interface


eduardonunesp:
The `blockTime` property which is available on legacy `GRPC` events aren't available on events dispatched from `JSON RPC` (`/eth`)

Tested with build `1313`

enlight:
Resolved by https://github.com/loomnetwork/loomchain/pull/1539

PR #1539, opened 2019-10-16
Add Blocktime to subscription Events

Sriep:

Add blocktime to subscription events. Addresses https://github.com/loomnetwork/loomchain/issues/1538 .

### DIFF
--- a/eth/subs/subs_set.go
+++ b/eth/subs/subs_set.go
@@ -59,9 +59,7 @@ func (s *EthSubscriptionSet) EmitEvent(data types.EventData) error {
 	if err != nil {
 		return errors.Wrapf(err, "marshaling event %v", data)
 	}
-	jLog := eth.EncEvent(data)
-	jLog.BlockTime = eth.EncInt(data.BlockTime)
-	s.logsHub.Publish(pubsub.NewMessage(string(ethMsg), jLog))
+	s.logsHub.Publish(pubsub.NewMessage(string(ethMsg), eth.EncEvent(data)))
 	return nil
 }
 
--- a/eth/subs/subs_set.go
+++ b/eth/subs/subs_set.go
@@ -59,7 +59,9 @@ func (s *EthSubscriptionSet) EmitEvent(data types.EventData) error {
 	if err != nil {
 		return errors.Wrapf(err, "marshaling event %v", data)
 	}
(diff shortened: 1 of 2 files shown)


rpc/eth/jsonrpc_conversion.go
```diff
@@ -188,6 +187,13 @@ func EncEvent(log types.EventData) JsonLog {
 	return jLog
 }
 
+func EncSubscriptionEvent(log types.EventData) JsonLog {
```
enlight: Since this 3-liner appears to be only called from one place just inline the code at the callsite.

eth/subs/subs_set.go
```diff
@@ -59,7 +59,9 @@ func (s *EthSubscriptionSet) EmitEvent(data types.EventData) error {
 	if err != nil {
 		return errors.Wrapf(err, "marshaling event %v", data)
 	}
-	s.logsHub.Publish(pubsub.NewMessage(string(ethMsg), eth.EncEvent(data)))
+	jLog := eth.EncEvent(data)
+	jLog.BlockTime = eth.EncInt(data.BlockTime)
```
enlight: Any particular reason you didn't set the block time in `eth.EncEvent` itself?

Sriep: When writing out the list of events in a transaction receipt, the blocktime is left out.

enlight: Right, but why?

Sriep: Not sure. My guess FWIW is that there are technical difficulties in go-ethereum in returning the blocktime with receipts, possibly because ethereum tends to rerun transactions to get receipts rather than save them.

https://github.com/loomnetwork/go-ethereum/blob/loomchain/core/types/receipt.go#L46
https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_gettransactionreceipt

enlight: The block time on events is a Loom extension, so what go-ethereum does in this case doesn't matter. The old `/query` endpoint for fetching receipts returns the block time in the event logs, so we should do the same with the `/eth` endpoint.

Sriep: Ah! OK! I'll add the block time to logs internal to receipts as well.

Issue #1540, opened 2019-10-16
QueryServer.EThCall: Caller chain Id should be assumed to be "eth"


Sriep:
EThCall: [Caller chain Id should be assumed to be "eth"](https://github.com/loomnetwork/loomchain/blob/master/rpc/query_server.go#L281)

enlight:
Resolved by https://github.com/loomnetwork/loomchain/pull/1542

PR #1542, opened 2019-10-16
Modify eth_call to expect the From parameter to be an Ethereum address

Sriep:

https://github.com/loomnetwork/loomchain/issues/1540
If `from` address entered, set to use "eth" chain id or give an error if eth not supported.
If there is no `from` address, use loom.RootAddress(s.ChainID).

### DIFF
--- a/rpc/query_server.go
+++ b/rpc/query_server.go
@@ -295,6 +295,7 @@ func (s *QueryServer) EthCall(query eth.JsonTxCallObject, block eth.BlockHeight)
 	if err != nil {
 		return resp, err
 	}
+
 	bytes, err := s.queryEvm(caller, contract, data)
 	return eth.EncBytes(bytes), err
 }
--- a/rpc/query_server.go
+++ b/rpc/query_server.go
@@ -277,16 +277,11 @@ func (s *QueryServer) queryEvm(caller, contract loom.Address, query []byte) ([]b
 
 // https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_call
 func (s *QueryServer) EthCall(query eth.JsonTxCallObject, block eth.BlockHeight) (resp eth.Data, err error) {
-	var caller loom.Address
-	// TODO: This is wrong, the caller chain ID should be assumed to be "eth:"
-	if len(query.From) > 0 {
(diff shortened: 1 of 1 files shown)


rpc/query_server.go
```diff
@@ -1133,6 +1125,21 @@ func (s *QueryServer) getBlockHeightFromHash(hash []byte) (uint64, error) {
 	}
 }
 
+func (s *QueryServer) getEthAccount(address eth.Data) (loom.Address, error) {
+	snapshot := s.StateProvider.ReadOnlyState()
+	defer snapshot.Release()
```
enlight: The snapshot should be passed in as a parameter so we don't create multiple snapshots to handle a single query.

rpc/query_server.go
```diff
@@ -1009,6 +1009,11 @@ func (s *QueryServer) EthUnsubscribe(id eth.Quantity) (unsubscribed bool, err er
 // The input address is assumed to be an Ethereum account address, so it'll be mapped to a local
 // account, and the transaction count returned will be for that local account.
 func (s *QueryServer) EthGetTransactionCount(address eth.Data, block eth.BlockHeight) (eth.Quantity, error) {
+	resolvedAddr, err := s.getEthAccount(address)
+	if err != nil {
+		return eth.Quantity("0x0"), errors.Wrap(err, "failed to resolve account address")
```
enlight: Do the `error.Wrap` in `getEthAccount`

enlight: And lets return `eth.ZeroedQuantity`

rpc/query_server.go
```diff
@@ -277,15 +277,17 @@ func (s *QueryServer) queryEvm(caller, contract loom.Address, query []byte) ([]b
...
 	if len(query.From) > 0 {
-		caller, err = s.getEthAccount(query.From)
+		caller, err = s.getEthAccount(snapshot, query.From)
 		if err != nil {
 			return resp, err
 		}
 	} else {
 		caller = loom.RootAddress(s.ChainID)
 	}
+	snapshot.Release()
```
enlight: Replace this with a `defer snapshot.Release()` at the top of this function, and pass the snapshot into `s.queryEvm`. Probably will need to change `s.queryPlugin` to also take a snapshot as input.

PR #1545, opened 2019-10-18
Add pprof profiling data http server

pathornteng:
Currently, s304 is taking up much more memory compared to s303. We need to investigate what is wrong with it so this PR adds pprof which is a http server used for showing runtime data. 

Ref: https://golang.org/pkg/net/http/pprof/


pathornteng:
We already have this endpoint. So close this PR.

Issue #1535, opened 2019-10-10
Cached Multi-Writer AppStore


pathornteng:
Currently, we have a `VersionedCachingStore` used by` QueryService`. However, our application does not have cached layer so it hits DB every time to query data. We can significantly reduce IOPS if we implement cached layer.

pathornteng:
LevelDB seems to has already had in-memory cache. This is no longer needed.

Issue #1310, opened 2019-07-08
Cache Feature Flag and Cfg Setting 


pathornteng:
Currently, `Feature Flag` and `Cfg Setting` are read from `app.db` every block. We should cache them in memory to reduce IOPS. The caching can be done in `MultiWriterAppStore`

pathornteng:
LevelDB seems to have already had a in-memory cache. This is no longer an issue.

PR #1552, opened 2019-10-24
Implement CachingKVStore

pathornteng:
Currently, our chain has to access a number of feature flags and on-chain config every block. To reduce IOPS, we should introduce a new caching layer for kv store. 

Issue: https://github.com/loomnetwork/loomchain/issues/1535 and https://github.com/loomnetwork/loomchain/issues/1310


pathornteng:
This PR has been already deployed and tested. Even though the cache hit rate is 95%, the read operation rate is not reduced. LevelDB seems to has already had in-memory cache implemented. Close this PR as it does not work as expected.

PR #1554, opened 2019-10-25
Add block range limit log poll

pathornteng:

### DIFF
--- a/eth/polls/polllogs.go
+++ b/eth/polls/polllogs.go
@@ -23,9 +23,10 @@ type EthLogPoll struct {
 	lastBlockRead uint64
 	evmAuxStore   *evmaux.EvmAuxStore
 	blockStore    store.BlockStore
+	maxBlockRange uint64
 }
 
-func NewEthLogPoll(filter string, evmAuxStore *evmaux.EvmAuxStore, blockStore store.BlockStore) (*EthLogPoll, error) {
+func NewEthLogPoll(filter string, evmAuxStore *evmaux.EvmAuxStore, blockStore store.BlockStore, maxBlockRange uint64) (*EthLogPoll, error) {
 	ethFilter, err := utils.UnmarshalEthFilter([]byte(filter))
 	if err != nil {
 		return nil, err
@@ -33,8 +34,8 @@ func NewEthLogPoll(filter string, evmAuxStore *evmaux.EvmAuxStore, blockStore st
 	p := &EthLogPoll{
 		filter:        ethFilter,
 		lastBlockRead: uint64(0),
-		evmAuxStore:   evmAuxStore,
 		blockStore:    blockStore,
+		maxBlockRange: maxBlockRange,
 	}
 	return p, nil
 }
@@ -55,6 +56,10 @@ func (p *EthLogPoll) Poll(
 		return p, nil, errors.New("Filter FromBlock is greater than ToBlock")
 	}
 
+	if end-start > p.maxBlockRange {
+		return p, nil, fmt.Errorf("max allowed block range (%d) exceeded", p.maxBlockRange)
+	}
+
 	if start <= p.lastBlockRead {
 		start = p.lastBlockRead + 1
 		if start > end {
@@ -71,6 +76,9 @@ func (p *EthLogPoll) Poll(
 	newLogPoll := &EthLogPoll{
 		filter:        p.filter,
 		lastBlockRead: end,
+		evmAuxStore:   p.evmAuxStore,
+		blockStore:    p.blockStore,
+		maxBlockRange: p.maxBlockRange,
 	}
 	return newLogPoll, eth.EncLogs(eventLogs), nil
 }
@@ -87,6 +95,10 @@ func (p *EthLogPoll) AllLogs(state loomchain.ReadOnlyState,
 	}
 	if start > end {
 		return nil, errors.New("Filter FromBlock is greater than ToBlock")
+	}
+
+	if end-start > p.maxBlockRange {
+		return nil, fmt.Errorf("max allowed block range (%d) exceeded", p.maxBlockRange)
 	}
 
 	eventLogs, err := query.GetBlockLogRange(
@@ -116,6 +128,11 @@ func (p *EthLogPoll) LegacyPoll(state loomchain.ReadOnlyState,
 			return p, nil, fmt.Errorf("filter start after filter end")
 		}
 	}
+
+	if end-start > p.maxBlockRange {
+		return p, nil, fmt.Errorf("max allowed block range (%d) exceeded", p.maxBlockRange)
+	}
+
 	eventLogs, err := query.GetBlockLogRange(
 		p.blockStore, state, start, end, p.filter.EthBlockFilter, readReceipts, p.evmAuxStore,
 	)
@@ -127,6 +144,7 @@ func (p *EthLogPoll) LegacyPoll(state loomchain.ReadOnlyState,
 		lastBlockRead: end,
 		evmAuxStore:   p.evmAuxStore,
 		blockStore:    p.blockStore,
+		maxBlockRange: p.maxBlockRange,
 	}
 
 	blocksMsg := types.EthFilterEnvelope_EthFilterLogList{
--- a/rpc/eth/config.go
+++ b/rpc/eth/config.go
@@ -5,10 +5,13 @@ package eth
 type Web3Config struct {
 	// GetLogsMaxBlockRange specifies the maximum number of blocks eth_getLogs will query per request
 	GetLogsMaxBlockRange uint64
+	// PollLogsMaxBlockRange specifies the maximum number of blocks eth_getFilterChanges will query per request
+	PollLogsMaxBlockRange uint64
 }
 
 func DefaultWeb3Config() *Web3Config {
 	return &Web3Config{
-		GetLogsMaxBlockRange: 20,
+		GetLogsMaxBlockRange:  20,
+		PollLogsMaxBlockRange: 20,
 	}
 }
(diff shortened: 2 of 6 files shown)


eth/polls/polllogs.go
```diff
@@ -89,6 +97,10 @@ func (p *EthLogPoll) AllLogs(state loomchain.ReadOnlyState,
 		return nil, errors.New("Filter FromBlock is greater than ToBlock")
 	}
 
+	if end-start > p.maxBlockRange {
+		return nil, fmt.Errorf("max allowed block range (%d) exceeded", p.maxBlockRange)
+	}
```
enlight: The issue here is that `EthLogPoll.AllLogs` is called from `QueryServer.EthGetFilterLogs` but when you call `EthGetFilterLogs` you can't specify the block range, you can only specify the block range when you call `QueryServer.EthNewFilter`... which makes all this rather frustrating to figure out for end users since the `latest` block is always increasing. So, instead of returning an error the end of the range should be auto-adjusted in `EthLogPoll.AllLogs`.

eth/polls/polllogs.go
```diff
@@ -55,6 +56,10 @@ func (p *EthLogPoll) Poll(
 		return p, nil, errors.New("Filter FromBlock is greater than ToBlock")
 	}
 
+	if end-start > p.maxBlockRange {
+		return p, nil, fmt.Errorf("max allowed block range (%d) exceeded", p.maxBlockRange)
+	}
```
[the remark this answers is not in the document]
pathornteng: I think it can be here. Less complex validation should always come first as you told me before.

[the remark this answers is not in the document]
enlight: The issue here is not complexity but rather correctness. The start of the range may be adjusted by the code below so the actual block range that will be queried may be less than the max, in which case returning this error here is a bug.

rpc/eth/config.go
```diff
@@ -5,10 +5,13 @@ package eth
 type Web3Config struct {
 	// GetLogsMaxBlockRange specifies the maximum number of blocks eth_getLogs will query per request
 	GetLogsMaxBlockRange uint64
+	// PollLogsMaxBlockRange specifies the maximum number of blocks eth_getFilterChanges will query per request
+	PollLogsMaxBlockRange uint64
```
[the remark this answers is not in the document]
pathornteng: Done

eth/polls/polllogs.go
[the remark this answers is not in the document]
pathornteng: Done

eth/polls/polllogs.go
pathornteng: Done

pathornteng:
> Also it looks like we need to fix up `EthTxPoll` and `EthBlockPoll` in a similar way.

`EthTxPoll` and `EthBlockPoll` do not allow users to specify filter though. It just queries last read block to latest block. However, we can auto-adjust block range. Should we go for it?

enlight:
> `EthTxPoll` and `EthBlockPoll` do not allow users to specify filter though. It just queries last read block to latest block. However, we can auto-adjust block range. Should we go for it?

Yeah, should auto-adjust the range.

eth/polls/polllogs.go
```diff
@@ -98,7 +98,7 @@ func (p *EthLogPoll) AllLogs(state loomchain.ReadOnlyState,
 	}
 
 	if end-start > p.maxBlockRange {
-		return nil, fmt.Errorf("max allowed block range (%d) exceeded", p.maxBlockRange)
+		start = end - p.maxBlockRange
```
enlight: I think it's adjusting the wrong end of the range, if you adjust the start it means the polling will miss blocks.

eth/polls/polltx.go
```diff
@@ -35,6 +37,9 @@ func (p *EthTxPoll) Poll(
 	if p.lastBlockRead+1 > uint64(state.Block().Height) {
 		return p, nil, nil
 	}
+	if uint64(state.Block().Height)-p.lastBlockRead > p.maxBlockRange {
+		p.lastBlockRead = uint64(state.Block().Height) - p.maxBlockRange
```
enlight: Here too polling would end up missing blocks, the end of the range should be set to `p.lastBlockRead + maxBlockRange`

eth/polls/polltx.go
```diff
@@ -46,7 +51,7 @@ func (p *EthTxPoll) Poll(
 func (p *EthTxPoll) AllLogs(
 	state loomchain.ReadOnlyState, id string, readReceipts loomchain.ReadReceiptHandler,
 ) (interface{}, error) {
-	_, results, err := getTxHashes(state, p.startBlock, readReceipts, p.evmAuxStore)
+	_, results, err := getTxHashes(state, uint64(state.Block().Height)-p.maxBlockRange, readReceipts, p.evmAuxStore)
```
enlight: Ditto

eth/polls/polltx.go
```diff
@@ -73,7 +78,9 @@ func (p *EthTxPoll) LegacyPoll(
 	if p.lastBlockRead+1 > uint64(state.Block().Height) {
 		return p, nil, nil
 	}
-
+	if uint64(state.Block().Height)-p.lastBlockRead > p.maxBlockRange {
+		p.lastBlockRead = uint64(state.Block().Height) - p.maxBlockRange
```
enlight: Ditto

eth/polls/polllogs.go
enlight: Actually this is fine as is, eth_getFilterLogs is not supposed to keep track of the last block read, it just re-evaluates the filter each time it's called.

eth/polls/polltx.go
pathornteng: Done

eth/polls/polltx.go
pathornteng: Done

eth/polls/polltx.go
pathornteng: Done

eth/polls/pollblocks.go
```diff
@@ -35,25 +38,36 @@ func (p *EthBlockPoll) Poll(
...
+	if toBlock-startBlock > p.maxBlockRange {
+		startBlock = toBlock - p.maxBlockRange
+	}
+	_, results, err := getBlockHashes(p.blockStore, toBlock, startBlock)
 	return eth.EncBytesArray(results), err
 }
 
 func getBlockHashes(
-	blockStore store.BlockStore, state loomchain.ReadOnlyState, lastBlockRead uint64,
+	blockStore store.BlockStore, toBlock, lastBlockRead uint64,
```
enlight: Can you swap around the order of `toBlock` and `lastBlockRead`?

eth/polls/pollblocks.go
```diff
@@ -35,25 +38,36 @@ func (p *EthBlockPoll) Poll(
...
 
+// AllLogs pull logs from last N blocks limited by p.maxBlockRange
 func (p *EthBlockPoll) AllLogs(
 	state loomchain.ReadOnlyState, id string, readReceipts loomchain.ReadReceiptHandler,
 ) (interface{}, error) {
-	_, results, err := getBlockHashes(p.blockStore, state, p.startBlock)
+	// Only pull logs from latest height - maxBlockRange
+	toBlock := uint64(state.Block().Height)
+	startBlock := p.startBlock
+	if toBlock-startBlock > p.maxBlockRange {
```
enlight: I think this should be
```golang
if toBlock-(startBlock+1) > p.maxBlockRange {
```
because blocks will be read starting from `startBlock+1`

eth/polls/pollblocks.go
```diff
@@ -35,25 +38,36 @@ func (p *EthBlockPoll) Poll(
...
+		startBlock = toBlock - p.maxBlockRange
```
enlight: And this would then be
```golang
startBlock = toBlock - p.maxBlockRange - 1
```

eth/polls/polltx.go
```diff
@@ -35,25 +37,35 @@ func (p *EthTxPoll) Poll(
...
+	startBlock := p.startBlock
+	if toBlock-startBlock > p.maxBlockRange {
+		startBlock = toBlock - p.maxBlockRange
+	}
+	_, results, err := getTxHashes(toBlock, startBlock, readReceipts, p.evmAuxStore)
 	return eth.EncBytesArray(results), err
 }
 
-func getTxHashes(state loomchain.ReadOnlyState, lastBlockRead uint64,
+func getTxHashes(toBlock uint64, lastBlockRead uint64,
```
enlight: Here too the order of `toBlock` & `lastBlockRead` should be swapped

eth/polls/polltx.go
```diff
@@ -35,25 +37,35 @@ func (p *EthTxPoll) Poll(
...
 	readReceipts loomchain.ReadReceiptHandler, evmAuxStore *evmaux.EvmAuxStore) (uint64, [][]byte, error) {
 	var txHashes [][]byte
-	for height := lastBlockRead + 1; height < uint64(state.Block().Height); height++ {
+	for height := lastBlockRead + 1; height < toBlock; height++ {
```
enlight: I think this should be `height <= toBlock`?

eth/polls/polltx.go
```diff
@@ -73,9 +85,12 @@ func (p *EthTxPoll) LegacyPoll(
...
 		return p, nil, nil
 	}
-
+	toBlock := uint64(state.Block().Height)
+	if toBlock-p.lastBlockRead > p.maxBlockRange {
+		toBlock = p.lastBlockRead + p.maxBlockRange
+	}
 	var txHashes [][]byte
-	for height := p.lastBlockRead + 1; height < uint64(state.Block().Height); height++ {
+	for height := p.lastBlockRead + 1; height < toBlock; height++ {
```
enlight: Same here

eth/polls/pollblocks.go
pathornteng: Done

PR #1556, opened 2019-10-28
Prune old EVM keys every N blocks

pathornteng:
Currently, pruning EVM keys every block forces Tendermint to create a new block faster so Blockscout indexer cannot catch up the speed and start to fall behind. This PR adds a new on-chain config, `PruneEvmKeysInterval`, which can be used to set pruning interval.   

Issue: https://github.com/loomnetwork/ops/issues/781

### DIFF
--- a/store/multi_writer_app_store.go
+++ b/store/multi_writer_app_store.go
@@ -225,8 +225,12 @@ func (s *MultiWriterAppStore) pruneOldEVMKeys() error {
 	}
 
 	maxKeysToPrune := cfg.GetAppStore().GetNumEvmKeysToPrune()
-	pruneInterval := cfg.GetAppStore().GetPruneEvmKeysInterval() // Prune old EVM Keys every N blocks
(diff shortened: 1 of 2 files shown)


store/multi_writer_app_store.go
```diff
@@ -225,7 +225,8 @@ func (s *MultiWriterAppStore) pruneOldEVMKeys() error {
 	}
 
 	maxKeysToPrune := cfg.GetAppStore().GetNumEvmKeysToPrune()
-	if maxKeysToPrune > 0 {
+	pruneInterval := cfg.GetAppStore().GetPruneEvmKeysInterval() // Prune old EVM Keys every N blocks
+	if maxKeysToPrune > 0 && (pruneInterval == 0 || s.Version()%int64(pruneInterval) == 0) {
```
enlight: I think this would be a bit clearer?
```golang
// If pruneInterval is set then only prune old EVM Keys every N blocks
if (pruneInterval != 0) && (s.Version()%int64(pruneInterval)) != 0) {
  maxKeysToPrune = 0
}
if maxKeysToPrune {
```

Issue #1561, opened 2019-10-29
Contract events shouldn't be emitted to subscribers in DeliverTx


enlight:
This should be fixed:
https://github.com/loomnetwork/loomchain/blob/3b78a6598c6a3746fc6a0788e201b9057e9e34d5/app.go#L810

enlight:
Resolved by https://github.com/loomnetwork/loomchain/pull/1562

PR #1562, opened 2019-10-30
Emit tx events after block commit

pathornteng:
Currently, events are emitted immediately after tx is processed successfully. We should delay it since the state won't be available until the block is committed.

Issue: https://github.com/loomnetwork/loomchain/issues/1561

### DIFF
--- a/app.go
+++ b/app.go
@@ -341,6 +341,11 @@ type ValidatorsManagerFactoryFunc func(state State) (ValidatorsManager, error)
 
 type ChainConfigManagerFactoryFunc func(state State) (ChainConfigManager, error)
 
+type CommittedTx struct {
+	result TxHandlerResult
+	txHash []byte
+}
+
 type Application struct {
 	lastBlockHeader abci.Header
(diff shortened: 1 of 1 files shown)


app.go
```diff
@@ -843,7 +842,18 @@ func (a *Application) Commit() abci.ResponseCommit {
...
+			if err := a.EventHandler.LegacyEthSubscriptionSet().EmitTxEvent(tx.result.Data, tx.result.Info); err != nil {
+				log.Error("Emit Tx Event error", "err", err)
+			}
+
+			if len(tx.txHash) > 0 {
+				if err := a.EventHandler.EthSubscriptionSet().EmitTxEvent(tx.txHash); err != nil {
+					log.Error("failed to emit tx event to subscribers", "err", err)
+				}
+			}
+		}
```
enlight: Looks ok at first look, I wonder if the block even should be emitted first though... probably should check ganache.

pathornteng: Yeah, checking with ganache, block events are emitted before tx events. Fixed.

PR #1563, opened 2019-10-31
Add a setting to disable mempool.wal

pathornteng:
Issue: https://github.com/loomnetwork/ops/issues/783

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -52,8 +52,8 @@ type Config struct {
 	// AppHash changes. Defaults to true.
 	CreateEmptyBlocks bool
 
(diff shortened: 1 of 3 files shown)


pathornteng:
rebuild sesame

config/config.go
```diff
@@ -52,6 +52,9 @@ type Config struct {
 	// AppHash changes. Defaults to true.
 	CreateEmptyBlocks bool
 
+	// MempoolWalDir sets directory path of mempool.wal. If empty, mempool.wal is disabled
+	MempoolWalDir string
```
enlight: I think we should just have `MempoolWalEnabled bool`, and it should be `false` by default

pathornteng: Done

PR #1573, opened 2019-11-07
Impliments debug_traceTransaction

Sriep:

issue https://github.com/loomnetwork/loomchain/issues/1482
Implements `debug_traceTransaction` as outlined in [Management APIs](https://github.com/ethereum/go-ethereum/wiki/Management-APIs#debug_tracetransaction)

Needs two other PRs
* Transfer gatway modified to with new NewLoomVm trace paramter https://github.com/loomnetwork/transfer-gateway/pull/118
* Change to go-ethereum to access some internal objects https://github.com/loomnetwork/go-ethereum/pull/8

### DIFF
--- a/app.go
+++ b/app.go
@@ -8,8 +8,11 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/ethereum/go-ethereum/core/vm"
 	"github.com/loomnetwork/go-loom/config"
 	"github.com/loomnetwork/go-loom/util"
+	"github.com/pkg/errors"
+
 	"github.com/loomnetwork/loomchain/eth/utils"
 	"github.com/loomnetwork/loomchain/features"
 	"github.com/loomnetwork/loomchain/registry"
@@ -20,14 +23,15 @@ import (
 	cctypes "github.com/loomnetwork/go-loom/builtin/types/chainconfig"
 	"github.com/loomnetwork/go-loom/plugin"
 	"github.com/loomnetwork/go-loom/types"
-	"github.com/loomnetwork/loomchain/log"
-	"github.com/loomnetwork/loomchain/store"
-	blockindex "github.com/loomnetwork/loomchain/store/block_index"
-	evmaux "github.com/loomnetwork/loomchain/store/evm_aux"
 	stdprometheus "github.com/prometheus/client_golang/prometheus"
 	abci "github.com/tendermint/tendermint/abci/types"
 	"github.com/tendermint/tendermint/libs/common"
 	ttypes "github.com/tendermint/tendermint/types"
+
+	"github.com/loomnetwork/loomchain/log"
+	"github.com/loomnetwork/loomchain/store"
+	blockindex "github.com/loomnetwork/loomchain/store/block_index"
+	evmaux "github.com/loomnetwork/loomchain/store/evm_aux"
 )
 
 type ReadOnlyState interface {
@@ -302,6 +306,11 @@ type TxHandler interface {
 	ProcessTx(state State, txBytes []byte, isCheckTx bool) (TxHandlerResult, error)
 }
 
+type TxHandlerFactory interface {
+	TxHandler(tracer vm.Tracer, metrics bool) (TxHandler, error)
+	Copy(newStore store.VersionedKVStore) TxHandlerFactory
+}
+
 type TxHandlerFunc func(state State, txBytes []byte, isCheckTx bool) (TxHandlerResult, error)
 
 type TxHandlerResult struct {
@@ -348,6 +357,7 @@ type Application struct {
 	Store           store.VersionedKVStore
 	Init            func(State) error
 	TxHandler
+	TxHandlerFactory
 	QueryHandler
 	EventHandler
 	ReceiptHandlerProvider
@@ -577,17 +587,19 @@ func (a *Application) EndBlock(req abci.RequestEndBlock) abci.ResponseEndBlock {
 	}
 
 	// TODO: receiptHandler.CommitBlock() should be moved to Application.Commit()
-	storeTx := store.WrapAtomic(a.Store).BeginTx()
-	receiptHandler := a.ReceiptHandlerProvider.Store()
-	if err := receiptHandler.CommitBlock(a.height()); err != nil {
-		storeTx.Rollback()
-		// TODO: maybe panic instead?
-		log.Error(fmt.Sprintf("aborted committing block receipts, %v", err.Error()))
-	} else {
-		storeTx.Commit()
+	if a.ReceiptHandlerProvider != nil {
+		storeTx := store.WrapAtomic(a.Store).BeginTx()
+		receiptHandler := a.ReceiptHandlerProvider.Store()
+		if err := receiptHandler.CommitBlock(a.height()); err != nil {
+			storeTx.Rollback()
+			// TODO: maybe panic instead?
+			log.Error(fmt.Sprintf("aborted committing block receipts, %v", err.Error()))
+		} else {
+			storeTx.Commit()
+		}
 	}
 
-	storeTx = store.WrapAtomic(a.Store).BeginTx()
+	storeTx := store.WrapAtomic(a.Store).BeginTx()
 	state := NewStoreState(
 		context.Background(),
 		storeTx,
@@ -774,15 +786,18 @@ func (a *Application) deliverTx2(storeTx store.KVStoreTx, txBytes []byte) abci.R
 		a.GetValidatorSet,
 	).WithOnChainConfig(a.config)
 
-	receiptHandler := a.ReceiptHandlerProvider.Store()
-	defer receiptHandler.DiscardCurrentReceipt()
-	defer a.EventHandler.Rollback()
+	if a.ReceiptHandlerProvider != nil {
+		defer a.ReceiptHandlerProvider.Store().DiscardCurrentReceipt()
+	}
+	if a.EventHandler != nil {
+		defer a.EventHandler.Rollback()
+	}
 
 	r, txErr := a.TxHandler.ProcessTx(state, txBytes, false)
 
 	// Store the receipt even if the tx itself failed
 	var receiptTxHash []byte
-	if a.ReceiptHandlerProvider.Reader().GetCurrentReceipt() != nil {
+	if a.ReceiptHandlerProvider != nil && a.ReceiptHandlerProvider.Reader().GetCurrentReceipt() != nil {
 		receiptTxHash = a.ReceiptHandlerProvider.Reader().GetCurrentReceipt().TxHash
 		txHash := ttypes.Tx(txBytes).Hash()
 		// If a receipt was generated for an EVM tx add a link between the TM tx hash and the EVM tx hash
@@ -793,7 +808,7 @@ func (a *Application) deliverTx2(storeTx store.KVStoreTx, txBytes []byte) abci.R
 				ChildTxHash:  receiptTxHash,
 			})
 		}
-		receiptHandler.CommitCurrentReceipt()
+		a.ReceiptHandlerProvider.Store().CommitCurrentReceipt()
 	}
 
 	if txErr != nil {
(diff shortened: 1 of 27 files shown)


rpc/query_server_test.go
```diff
@@ -101,6 +101,10 @@ func (s *stateProvider) ReadOnlyState() loomchain.State {
 	)
 }
 
+func (s *stateProvider) ReplayApplication(uint64, store.BlockStore) (*loomchain.Application, int64, error) {
+	return nil, 0, fmt.Errorf("Not implimented")
```
enlight: -> `"not implemented"`

store/versioned_cachingstore.go
```diff
@@ -396,14 +407,6 @@ func newVersionedCachingStoreSnapshot(snapshot Snapshot, cache *versionedBigCach
 	}
 }
 
-func (c *versionedCachingStoreSnapshot) Delete(key []byte) {
```
enlight: Why were these functions removed?

store/versioned_cachingstore.go
```diff
@@ -371,11 +370,23 @@ func (c *versionedCachingStore) SaveVersion() ([]byte, int64, error) {
...
+}
+
+func (c *versionedCachingStore) GetSnapshotAt(version int64) (Snapshot, error) {
+	if version == 0 {
+		return newVersionedCachingStoreSnapshot(
+			c.VersionedKVStore.GetSnapshot(),
+			c.cache, c.version-1, c.logger,
+		), nil
+	} else {
+		return c.VersionedKVStore.GetSnapshotAt(version)
```
enlight: Why doesn't this use `newVersionedCachingStoreSnapshot`?

evm/loomevm.go
```diff
@@ -121,20 +127,22 @@ type LoomVm struct {
...
+	tracer         ethvm.Tracer
 }
 
 func NewLoomVm(
 	loomState loomchain.State,
-	eventHandler loomchain.EventHandler,
 	receiptHandler loomchain.WriteReceiptHandler,
 	createABM AccountBalanceManagerFactoryFunc,
 	debug bool,
+	tracer ethvm.Tracer,
```
enlight: Since the tracer is optional add a new function to set it instead of adding it to the constructor:
```golang
func (lvm LoomVm) WithTracer(tracer ethvm.Tracer) LoomVm {
  lvm.tracer = tracer
  return lvm
}
```
And it doesn't look like the `debug` parameter is ever set to anything other than `false`, so I think you can remove it from the construction function as well.

e2e/tests/truffle/test/DebugFunctions.js
```diff
@@ -0,0 +1,57 @@
...
+const { getLoomEvmTxHash, waitForXBlocks } = require('./helpers');
+const {
+    SpeculativeNonceTxMiddleware, SignedTxMiddleware, Client,
+    LocalAddress, CryptoUtils, LoomProvider
+} = require('loom-js');
+const TxHashTestContract = artifacts.require('TxHashTestContract');
+
+contract('debug_traceTransaction', async (accounts) => {
+    let contract, fromAddr, nodeAddr, txHashTestContract, web3, web3js;
+    const StructLogsLength = 87;
```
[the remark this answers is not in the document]
Sriep: When you do a trace you get an array of data. The data is not particularly human-readable, but if it is returning anything it's probably the correct stuff. Maybe I should just check if greater than zero, 87 is just how many entries happen to be in the trace log.
We should do an integration test using remix, as that seems the best way to confirm that the data returned and make sure the data format does not need tweaking in some way.

store/versioned_cachingstore.go
Sriep: `versionedCachingStoreSnapshot` is only used as a  `Snapshot` in GetSnapshot. 
A snapshot only has `Get, Range, Has and Release` members. So the extra versionedCachingStoreSnapshot members can't be used and only seem to confuse.

store/versioned_cachingstore.go
Sriep: It could do, but the version caching wrapper did not seem necessary for a read-only store used for a replay.

evm/loomevm.go
Sriep: The evm debugging flag could be considered redundant with the new trace functionality. And as it is not actually used, do you want me to remove the evmDebugging functionality - which is just set a trace log - can be done with tracer - and write out a warning.

Sriep: NewLoomVm returns a `vm.VM` not a `LoomVm` object so we need to change something to use `loomVm.WithTracer`. Probably best to change the return type of `NewLoomVm` to `LoomVm` so SetTracer can be run on it. Could also add `WithTracer` to VM interface.

enlight: > The evm debugging flag could be considered redundant with the new trace functionality and could be removed. It sets a trace log - can be done with tracer - and writes out a warning.
> Evm Debug can be set from loom.yaml at https://github.com/loomnetwork/loomchain/blob/master/cmd/loom/loom.go#L837 . I can replace the config `EVMDebugEnabled` with an `EVMTraceEnable` option which can add a trace to the `main app.TxHandlerFactory` and dumps the result to the log file?

Yeah, the EVM debug flag doesn't seem to actually do anything unless the tracer is set, so lets replace it with an `EVMTracerEnabled` setting.

e2e/tests/truffle/test/DebugFunctions.js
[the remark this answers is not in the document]
Sriep: I changed to check an error and whether there was non-empty output. I also add code to the loomchain for `debug_traceTransaction` to give an [error if the replay gives different output to the original tx](https://github.com/loomnetwork/loomchain/blob/ethdebug/rpc/debug/trace.go#L60).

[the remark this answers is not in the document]
Sriep: Looking into remix.

evm/evm.go
```diff
@@ -299,31 +302,22 @@ func defaultChainConfig(enableConstantinople bool) params.ChainConfig {
...
 		}
-		debug = true
+		tracer = vm.NewStructLogger(&logCfg)
 	}
-	logger := vm.NewStructLogger(&logCfg)
+
 	return vm.Config{
 		// Debug enabled debugging Interpreter options
-		Debug: debug,
+		Debug: tracer != nil,
```
enlight: `tracer` will never be `nil` here because when `createVmConfig(nil)` is called a new tracer is created using `vm.NewStructLogger(&logCfg)`.

evm/loomevm.go
```diff
@@ -74,11 +76,14 @@ func NewLoomEvm(
 		return nil, err
 	}
 
-	p.Evm = NewEvm(p.sdb, loomState, abm, debug)
+	p.wrappedEVM, err = NewEvm(p.sdb, loomState, abm, tracer)
+	if err != nil {
+		return nil, errors.Wrap(err, "creating tracer")
```
enlight: -> `errors.Wrap(err, "failed to create EVM")`

store/splitstore.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	return ss.KVReader.Get(key)
+}
+
+func (ss *splitStore) Range(prefix []byte) plugin.RangeData {
+	readerRange := ss.KVReader.Range(prefix)
+	updateRange := ss.VersionedKVStore.Range(prefix)
+	for _, re := range updateRange {
+		if !ss.KVReader.Has(re.Key) && !ss.deleted[string(re.Key)] {
+			readerRange = append(readerRange, re)
```
pathornteng: This `range` function does not seem right to me. The logic is appending the data instead of updating it. However, I believe it is not used in evm logic but probably we should try to get it right.

store/multi_writer_app_store.go
```diff
@@ -262,12 +264,30 @@ func (s *MultiWriterAppStore) Prune() error {
...
-	appStoreTree := (*iavl.ImmutableTree)(atomic.LoadPointer(&s.lastSavedTree))
+
+	var err error
+	var appStoreTree *iavl.ImmutableTree
+	if version == 0 {
+		appStoreTree = (*iavl.ImmutableTree)(atomic.LoadPointer(&s.lastSavedTree))
+	} else {
+		appStoreTree, err = s.appStore.tree.GetImmutable(version)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to load immutable tree for version %v", version)
```
pathornteng: Should we panic instead of returning the error? It would simplify the code quite a bit.

Sriep: It makes sense for `GetSnapshot` to panic as there should always be a lastest version. 
However, there are legitimate reasons for there not to be data available for an arbitrary version, as we can prune versions. 
In Application.ReplayApplication https://github.com/loomnetwork/loomchain/blob/ethdebug/app.go#L950 we take account of this.

store/splitstore.go
```diff
@@ -0,0 +1,93 @@
...
+	KVReader
+	VersionedKVStore
+	deleted map[string]bool
+	version int64
+}
+
+func NewSplitStore(full KVReader, empty VersionedKVStore, version int64) VersionedKVStore {
+	return &splitStore{
+		KVReader:         full,
+		VersionedKVStore: empty,
```
enlight: Seeing as this mutable `VersionedKVStore` is supposed to be empty you should just create it here instead of passing it in.

store/splitstore.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	if ss.deleted[string(key)] {
+		return nil
+	}
+	return ss.KVReader.Get(key)
+}
+
+func (ss *splitStore) Range(prefix []byte) plugin.RangeData {
+	readerRange := ss.KVReader.Range(prefix)
+	updateRange := ss.VersionedKVStore.Range(prefix)
```
enlight: Seeing as `ss.VersionedKVStore` is really a `MemStore` and `MemStore.Range` iterates over a `map`... this is going to be non-deterministic.

rpc/debug/trace.go
```diff
@@ -0,0 +1,196 @@
...
+					if includeEnd {
+						_ = app.DeliverTx(resultBlock.Block.Data.Txs[i])
+					}
+					return resultBlock, nil
+				}
+			}
+		}
+
+		_ = app.EndBlock(requestEndBlock(h))
+		_ = app.Commit()
```
enlight: Can you add a TODO comment here to say that we should really be matching the app hash returned here with the one that was stored in the next block, otherwise we can't really know if the replay was successful or not.

app.go
```diff
@@ -701,6 +715,9 @@ func (a *Application) DeliverTx(txBytes []byte) abci.ResponseDeliverTx {
 	} else {
 		r = a.deliverTx(storeTx, txBytes)
 	}
+	if a.EVMTracer != nil {
+		log.Debug("evm trace", "trace", a.EVMTracer)
+	}
```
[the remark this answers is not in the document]
Sriep: Actually no. This prints out the log if you have setup the trace option in loom.yaml. [`a.EVMTracer` is set from the loom.yaml settings](https://github.com/loomnetwork/loomchain/blob/ethdebug/cmd/loom/loom.go#L793), the same tracer is then built into the application's vmManager.

store/splitstore.go
Sriep: Added a sort to splitstore.Range. I could add the sort to MemStore.Range?

store/splitstore.go
```diff
@@ -33,6 +36,12 @@ func (ss *splitStore) Get(key []byte) []byte {
...
 	updateRange := ss.VersionedKVStore.Range(prefix)
+
+	// VersionedKVStore comes from a MemStore, hence updateRange is not deterministic
+	sort.Slice(updateRange, func(i, j int) bool {
+		return bytes.Compare(updateRange[i].Key, updateRange[j].Key) < 0
+	})
+
 	for _, re := range readerRange {
 		if !ss.VersionedKVStore.Has(re.Key) && !ss.deleted[string(re.Key)] {
 			updateRange = append(updateRange, re)
```
enlight: Seems like this doesn't ensure a consist order in the final `updateRange`, probably need to move the sort after this loop, not before it.

Issue #1574, opened 2019-11-07
Add new QueryServer RPC method to fetch total staked amount


enlight:
We need a new `dpos_total_staked` RPC method that returns a simple JSON response containing the total staked amount. The response should be cached so the dpos contract doesn't need to be queried too often (perhaps once a minute at most).

pathornteng:
Dashboard calculates total staked using the sum of each validator total delegation plus its whitelist amount but excluding whitelist amount of bootstrapping nodes. The problem is the bootstrapping node addresses are hard-coded here https://github.com/loomnetwork/dashboard/blob/b2ceea2b21f7620ca6a0b19e38ae8cd05d521f58/src/config/production.ts#L25
Unfortunately, DPOSv3 has no idea which validators are bootstrapping validators.

Here are some of the solutions:
1. make dpos_total_staked read a list of bootstrapping node addresses configurable from loom.yaml and exclude these validators' whitelist amounts.
2. add a bootstrapping field in DPOS candidate protobuf and add a command to set this flag

enlight:
Yeah, just add the list of bootstrap nodes to `loom.yml`, in PlasmaChain builds default the list to the current PlasmaChain boostrap nodes.

enlight:
Resolved by https://github.com/loomnetwork/loomchain/pull/1575

PR #1575, opened 2019-11-08
Add dpos_total_staked Query RPC endpoint

pathornteng:
Issue: https://github.com/loomnetwork/loomchain/issues/1574


rpc/query_server.go
```diff
@@ -599,6 +614,35 @@ func (s *QueryServer) GetContractRecord(contractAddrStr string) (*types.Contract
 	return k, nil
 }
 
+func (s *QueryServer) DposTotalStaked() (*types.DposTotalStakedResponse, error) {
+	snapshot := s.StateProvider.ReadOnlyState()
+	defer snapshot.Release()
+	if s.totalStakedAmount != nil {
+		if time.Since(s.totalStakedAmount.createAt) <= time.Minute {
```
enlight: Should probably make the cache interval configurable via loom.yml. The amount is only going to change when an election runs right?

builtin/plugins/dposv3/dpos.go
```diff
@@ -2658,3 +2606,98 @@ func Initialize(ctx contract.Context, initState *InitializationState) error {
...
+	}
+
+	candidates, err := LoadCandidateList(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	totalStaked := &types.BigUInt{Value: *loom.NewBigUIntFromInt(0)}
+	for _, candidate := range candidates {
+		response, err := listDelegations(ctx, &ListDelegationsRequest{Candidate: candidate.Address})
```
enlight: Do we really need to load all delegations? We don't keep track of this amount per-validator?

rpc/query_service.go
```diff
@@ -62,8 +62,8 @@ type QueryService interface {
 	EthAccounts() ([]eth.Data, error)
 
 	ContractEvents(fromBlock uint64, toBlock uint64, contract string) (*types.ContractEventsResult, error)
-
 	GetContractRecord(contractAddr string) (*types.ContractRecordResponse, error)
+	DposTotalStaked() (*types.DposTotalStakedResponse, error)
```
enlight: -> `DPOSTotalStaked`

rpc/query_server.go
```diff
@@ -599,6 +614,35 @@ func (s *QueryServer) GetContractRecord(contractAddrStr string) (*types.Contract
 	return k, nil
 }
 
+func (s *QueryServer) DposTotalStaked() (*types.DposTotalStakedResponse, error) {
```
enlight: `DPOSTotalStaked`

rpc/query_server.go
```diff
@@ -599,6 +614,35 @@ func (s *QueryServer) GetContractRecord(contractAddrStr string) (*types.Contract
...
+	resp, err := dposv3.TotalStaked(dposCtx, s.BootstrapNodes)
+	if err != nil {
+		return nil, err
+	}
+	s.totalStakedAmount = &totalStakedAmount{
+		createAt: time.Now(),
+		amount:   *resp,
+	}
+
+	return &types.DposTotalStakedResponse{
```
enlight: This RPC method should return a simple JSON response, not a protobuf

rpc/query_server.go
```diff
@@ -343,6 +353,11 @@ func (s *QueryServer) EthGetCode(address eth.Data, block eth.BlockHeight) (eth.D
 	return eth.EncBytes(code), nil
 }
 
+// Attempts to construct the context of the DPOSv3 contract.
+func (s *QueryServer) createDposV3Ctx(state loomchain.State) (contractpb.StaticContext, error) {
```
enlight: Inline this at the call-site.

rpc/query_server.go
pathornteng: No, the amount changes all the time from the delegation, undelegation, and reward claims. I can make it configurable via loom.yaml. However, since we are using load-balancer for routing requests, the amount might be going up and down depending on the cache of nodes. We should consider cachong it on CDN instead.

builtin/plugins/dposv3/dpos.go
pathornteng: Unfortunately, we don't.

rpc/query_server.go
enlight: Well, if the amount is going to change all the time anyway it doesn't really matter if nodes are returning different values at a particular point in time.

builtin/plugins/dposv3/dpos.go
enlight: Well, this is going to load ALL the delegations N times over, once for each candidate, should only load those once.

pathornteng: This should be fine. ListDelegation is designed to be called this way. It only loads only delegations of a particular candidate, not all delegations.

rpc/query_service.go
pathornteng: Done

rpc/query_server.go
pathornteng: Done

rpc/query_server.go
pathornteng: Done

builtin/plugins/dposv3/dpos.go
enlight: `loadDelegationList` skips loading the individual delegation records for irrelevant candidates, but it still loads a giant "index" of all the delegations, each and every time unless you use the cached storage.

pathornteng: Fine, the logic was copied from `ListAllDelegations` but yeah it can be improved. Done

config/config.go
```diff
@@ -787,6 +791,21 @@ RootDir: "{{ .RootDir }}"
...
+#
+# Configuration of DPOSv3
+#
+DPOS:
+  # Specifies addresses of bootstrap nodes
+  BootstrapNodes:
+  {{- range .DPOS.BootstrapNodes}}
+    - "{{. -}}"
+  {{- end}}
+  # Time duration in seconds in which TotolStaked is valid
```
enlight: -> `How long (in seconds?) the response from the dpos_total_staked RPC method should be cached.`

builtin/plugins/dposv3/config.go
```diff
@@ -0,0 +1,30 @@
+package dposv3
```
enlight: This doesn't really have anything to do with the contract... I think it's better to put this in the RPC package.

builtin/plugins/dposv3/dpos.go
```diff
@@ -165,6 +166,10 @@ type (
 	GetRequestBatchTallyRequest = dtypes.GetRequestBatchTallyRequest
 )
 
+type DPOSTotalStaked struct {
```
enlight: This too doesn't belong here, it should be in the QueryServer.

builtin/plugins/dposv3/config.go
pathornteng: Putting it in RPC package will cause cyclic dependencies, I will put in it config package.

config/config.go
pathornteng: Done

builtin/plugins/dposv3/dpos.go
pathornteng: Done

rpc/query_server.go
```diff
@@ -599,6 +609,37 @@ func (s *QueryServer) GetContractRecord(contractAddrStr string) (*types.Contract
 	return k, nil
 }
 
+type DPOSTotalStaked struct {
```
enlight: `DPOSTotalStaked` -> `DPOSTotalStakedResponse`

config/config.go
```diff
@@ -169,6 +172,33 @@ func DefaultFnConsensusConfig() *FnConsensusConfig {
...
+	BootstrapNodes           []string
+	TotalStakedCacheDuration int64
+}
+
+func DefaultDPOSConfig() *DPOSConfig {
+	return &DPOSConfig{
+		BootstrapNodes: []string{
+			"default:0x0e99fc16e32e568971908f2ce54b967a42663a26",
+			"default:0xac3211caecc45940a6d2ba006ca465a647d8464f",
+			"default:0x69c48768dbac492908161be787b7a5658192df35",
```
enlight: This doesn't seem to match any candidate, where did you get the address from?

pathornteng: https://github.com/loomnetwork/dashboard/blob/b2ceea2b21f7620ca6a0b19e38ae8cd05d521f58/src/config/production.ts#L25

PR #1582, opened 2019-11-13
Implement eth_getStorageSize method

pathornteng:
Implement `eth_getStorageSize` method which returns the number of keys in the storage trie of an account. Each key/value has a fixed length which is 32 bytes. So the total storage trie space can be calculated from `number of keys * 32 * 2` bytes excluding metadata


evm/evm.go
```diff
@@ -264,6 +266,30 @@ func (e Evm) GetStorageAt(addr loom.Address, key []byte) ([]byte, error) {
...
 
+func (e Evm) GetStorageSize(addr loom.Address) (uint64, error) {
+	numKeys := uint64(0)
+	done := make(chan bool, 1)
+	// Count the number of keys in the storage trie with limit
+	go func() {
+		e.sdb.ForEachStorage(common.BytesToAddress(addr.Local), func(key, value common.Hash) bool {
+			numKeys++
+			// Limit the number of key iteration so that the response does not take too long
+			if numKeys >= storageKeyCountLimit {
```
enlight: So this won't actually give an accurate result for accounts with a lot of storage, and it'll take forever to iterate through the storage tree and load up the values we don't actually care about. Need to find a much quicker way to do this.

pathornteng:
There is no efficient way to find storage trie size since the underlying trie implementation does not have metadata that keeps tracking the number of nodes. Close this PR for now.

PR #1597, opened 2019-12-13
Skip a botched DPOS election on extdev

enlight:
DPOS contract had insufficient funds to process a rewards claim during the election in block `9856178` so we have to delay this election to resume the chain. Once the chain is back up and running we can transfer some funds to the DPOS contract so the next election goes through.


enlight:
rebuild sesame

PR #1610, opened 2020-01-30
Disable the VersionedCachingStore by default

enlight:
The caching provided by the `VersionedCachingStore` is of no benefit to nodes that aren't queried by clients because the transaction handlers bypass the cache. Since the vast majority of nodes, regardless of whether or not they're validators don't respond to client queries it makes more sense to disable the caching by default.


enlight:
rebuild sesame

Issue #1611, opened 2020-02-06
Add DPOS.UnbondAll


enlight:
QA needs a quick and easy way to unbond all delegations when testing, so we need to add a new admin-only `UnbondAll` method to the DPOS contract.

enlight:
Resolved in https://github.com/loomnetwork/loomchain/pull/1613

PR #1613, opened 2020-02-06
Add DPOS contract method UnbondAll

atchapcyp:
ref :  https://github.com/loomnetwork/loomchain/issues/1611
ref2 : https://github.com/loomnetwork/go-loom/pull/466

### DIFF
--- a/builtin/plugins/dposv3/dpos.go
+++ b/builtin/plugins/dposv3/dpos.go
@@ -1280,7 +1280,6 @@ func Elect(ctx contract.Context) error {
 	if err != nil {
 		return err
 	}
-
 	ctx.Logger().Debug("DPOSv3 Elect", "delegationResults", len(delegationResults))
 
 	validatorCount := int(state.Params.ValidatorCount)
@@ -2065,7 +2064,7 @@ func distributeDelegatorRewards(ctx contract.Context, cachedDelegations *CachedD
 			}
 			err = coin.Transfer(loom.UnmarshalAddressPB(delegation.Delegator), &delegation.UpdateAmount.Value)
 			if err != nil {
-				transferFromErr := fmt.Sprintf("Failed coin Transfer - distributeDelegatorRewards, %v, %s", delegation.Delegator.String(), delegation.UpdateAmount.Value.String())
+				transferFromErr := fmt.Sprintf("Failed coin TransferFrom - Delegate, %v, %s", delegator.String(), req.Amount.Value.String())
 				return nil, logDposError(ctx, err, transferFromErr)
 			}
 		} else if delegation.State == REDELEGATING {
--- a/builtin/plugins/dposv3/dpos.go
+++ b/builtin/plugins/dposv3/dpos.go
@@ -737,6 +737,8 @@ func (c *DPOS) Unbond(ctx contract.Context, req *UnbondRequest) error {
 	return c.emitDelegatorUnbondsEvent(ctx, delegation)
 }
 
+// UnbondAll unbonds the full amount on all currently bonded delegations.
+// This method can only be called by the DPOS oracle (admin).
 func (c *DPOS) UnbondAll(ctx contract.Context, req *UnbondAllRequest) error {
 	if !ctx.FeatureEnabled(features.DPOSVersion3_7, false) {
 		return errors.New("DPOS v3.7 is not enabled")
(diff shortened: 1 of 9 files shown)


builtin/plugins/dposv3/dpos.go
```diff
@@ -736,6 +737,54 @@ func (c *DPOS) Unbond(ctx contract.Context, req *UnbondRequest) error {
...
+	for _, di := range delegationIndexes {
+		delegation, err := GetDelegation(ctx, di.Index, *di.Validator, *di.Delegator)
+		if err == contract.ErrNotFound {
+			return logDposError(ctx, errors.New(fmt.Sprintf("delegation not found: %s %s", di.Validator, di.Delegator)), req.String())
+		} else if err != nil {
+			return errors.Wrap(err, "failed to load delegation")
+		}
+
+		if delegation.State != BONDED {
+			return logDposError(ctx, errors.New("Existing delegation not in BONDED state."), req.String())
```
enlight: Shouldn't return an error in this case, check if `UpdateAmount` different to `Amount` - if it is set `UpdateAmount` to match `Amount`.

builtin/plugins/dposv3/test_helpers.go
```diff
@@ -349,6 +349,14 @@ func (dpos *testDPOSContract) Unbond(ctx *plugin.FakeContext, validator *loom.Ad
 	return err
 }
 
+func (dpos *testDPOSContract) UnbondAll(ctx *plugin.FakeContext) error {
```
enlight: I don't think this abstraction is worth it, just makes the test harder to follow.

builtin/plugins/dposv3/dpos_test.go
```diff
@@ -495,6 +495,140 @@ func TestDelegate(t *testing.T) {
...
+
+	dpos, err := deployDPOSContract(pctx, &Params{
+		ValidatorCount: 21,
+		OracleAddress:  oracleAddr.MarshalPB(),
+	})
+	require.Nil(t, err)
+
+	// transfer coins to reward fund
+	amount := big.NewInt(1)
+	amount.Mul(amount, big.NewInt(1e18))
```
enlight: So the DPOS contract is only given 1 LOOM here, that's probably not enough to pay out all the rewards in this test.

builtin/plugins/dposv3/dpos_test.go
```diff
@@ -495,6 +495,140 @@ func TestDelegate(t *testing.T) {
...
+	require.Nil(t, err)
+
+	delegationAmount := big.NewInt(100)
+	err = coinContract.Approve(contractpb.WrapPluginContext(coinCtx.WithSender(addr1)), &coin.ApproveRequest{
+		Spender: dpos.Address.MarshalPB(),
+		Amount:  &types.BigUInt{Value: *loom.NewBigUInt(delegationAmount)},
+	})
+	require.Nil(t, err)
+
+	response, err := coinContract.Allowance(contractpb.WrapPluginContext(coinCtx.WithSender(oracleAddr)), &coin.AllowanceRequest{
```
enlight: You can remove this check, we don't need to test `coin.Approve` here, we've got other tests for that already.

builtin/plugins/dposv3/dpos_test.go
```diff
@@ -495,6 +495,140 @@ func TestDelegate(t *testing.T) {
...
+	require.Nil(t, err)
+
+	err = coinContract.Approve(contractpb.WrapPluginContext(coinCtx.WithSender(addr1)), &coin.ApproveRequest{
+		Spender: dpos.Address.MarshalPB(),
+		Amount:  &types.BigUInt{Value: *loom.NewBigUInt(delegationAmount)},
+	})
+	require.Nil(t, err)
+
+	// total rewards distribution should equal 0 before elections run
+	totalRewardDistribution, err := dpos.CheckRewards(pctx.WithSender(addr1))
```
enlight: You can get rid of all the rewards checks before calling `UnbondAll`, they aren't relevant in this test.

builtin/plugins/dposv3/dpos_test.go
```diff
@@ -495,6 +495,140 @@ func TestDelegate(t *testing.T) {
...
+	assert.True(t, common.IsPositive(*totalRewardDistribution))
+
+	pctx.SetFeature(features.DPOSVersion3_7, true)
+	
+	err = dpos.UnbondAll(pctx.WithSender(oracleAddr))
+	require.Nil(t, err)
+
+	require.NoError(t, elect(pctx.WithSender(addr1), dpos.Address))
+
+	_, delegatedAmount, _, err = dpos.CheckDelegation(pctx, &addr1, &delegatorAddress1)
```
enlight: Probably better to check if there are any delegations left at all at this point, I'd expect none to be left... though I could be wrong.

builtin/plugins/dposv3/dpos.go
```diff
@@ -736,6 +737,54 @@ func (c *DPOS) Unbond(ctx contract.Context, req *UnbondRequest) error {
...
+	}
+
+	// ensure that function is only executed when called by oracle
+	if state.Params.OracleAddress == nil || sender.Compare(loom.UnmarshalAddressPB(state.Params.OracleAddress)) != 0 {
+		return logDposError(ctx, errOnlyOracle, req.String())
+	}
+
+	delegationIndexes, err := loadDelegationList(ctx)
+	if err != nil {
+		return logDposError(ctx, errors.New("Existing delegation not in BONDED state."), req.String())
```
enlight: Error message is incorrect.

builtin/plugins/dposv3/dpos.go
```diff
@@ -736,6 +737,56 @@ func (c *DPOS) Unbond(ctx contract.Context, req *UnbondRequest) error {
...
+	for _, di := range delegationIndexes {
+
+		delegation, err := GetDelegation(ctx, di.Index, *di.Validator, *di.Delegator)
+		if err == contract.ErrNotFound {
+			return logDposError(ctx, errors.New(fmt.Sprintf("delegation not found: %s %s", di.Validator, di.Delegator)), req.String())
+		} else if err != nil {
+			return errors.Wrap(err, "failed to load delegation")
+		}
+
+		if delegation.State != UNBONDING {
```
enlight: Actually we should only process the `BONDED` delegations, updating delegations that are in any other state is probably going to introduce a bug.

cmd/loom/dposV3_commands.go
```diff
@@ -757,6 +757,27 @@ func UnbondCmdV3() *cobra.Command {
 	return cmd
 }
 
+const unboundAllCmdExample = `
```
enlight: `unboundAllCmdExample` -> unbondAllCmdExample

Issue #1615, opened 2020-02-19
Node panics if it's queried before first call to Application.Commit


enlight:
```
Panic in RPC HTTP handler module=query-server err="runtime error: invalid memory address or nil pointer dereference" stack="goroutine 222 [running]:
    runtime/debug.Stack(0xc08379c968, 0x1a2d520, 0x2dad660)
	/usr/local/go-1.12/src/runtime/debug/stack.go:24 +0x9d
    github.com/loomnetwork/loomchain/vendor/github.com/tendermint/tendermint/rpc/lib/server.RecoverAndLogHandler.func1.1(0xc07e7ca720, 0x1ff0e80, 0xc08339a1d0, 0xbf8b61752a96049a, 0x3b5d42981, 0x2e16260, 0xc08333ef00)
	/tmp/gopath-jenkins-loom-sdk-pipeline3-1363/src/github.com/loomnetwork/loomchain/vendor/github.com/tendermint/tendermint/rpc/lib/server/http_server.go:134 +0x552
    panic(0x1a2d520, 0x2dad660)
	/usr/local/go-1.12/src/runtime/panic.go:522 +0x1b5
    github.com/loomnetwork/loomchain.(*Application).ReadOnlyState(0xc0002a9200, 0x2a, 0xc083302270)
	/tmp/gopath-jenkins-loom-sdk-pipeline3-1363/src/github.com/loomnetwork/loomchain/app.go:938 +0x57
    github.com/loomnetwork/loomchain/rpc.(*QueryServer).Query(0xc083374b40, 0xc0833021e0, 0x30, 0xc083302210, 0x2a, 0xc083300be0, 0xa, 0xc, 0xc000000000, 0x0, ...)
	/tmp/gopath-jenkins-loom-sdk-pipeline3-1363/src/github.com/loomnetwork/loomchain/rpc/query_server.go:171
```

It appears that `Application.ReadOnlyState()` can get called from the QueryServer before `Application.lastBlockHeader` is set to something other than `nil`. This likely happens when the node has just started up, before the first call to `Application.Commit()` where `Application.lastBlockHeader` is set to something sensible. So we probably just need to initialize `Application.lastBlockHeader` on startup instead of leaving it as `nil`.

enlight:
Resolved by https://github.com/loomnetwork/loomchain/pull/1620

PR #1620, opened 2020-02-25
Load last block header when app starts up before responding to queries

enlight:
This should fix most cases where https://github.com/loomnetwork/loomchain/issues/1615 occurs in practice.


enlight:
rebuild sesame

PR #1621, opened 2020-02-26
[Testing] Load latest version of IAVL tree right after flushed

atchapcyp:
For IAVL Testing only

### DIFF
--- a/store/iavlstore_test.go
+++ b/store/iavlstore_test.go
@@ -3,6 +3,7 @@ package store
 import (
 	"bytes"
 	"os"
+	"strconv"
 	"testing"
 
 	"github.com/stretchr/testify/require"
@@ -142,6 +143,28 @@ func testNormal(t *testing.T) {
 		return false
 	})
 	diskDb.Close()
+}
+
+func TestLoadIAVLVersion(t *testing.T) {
+	flushInterval = int64(5)
+	memDB := db.NewMemDB()
+	store, err := NewIAVLStore(memDB, 0, 0, flushInterval)
+	require.NoError(t, err)
+	var lastVersion int64
+	var s string
+	for i := int64(1); i <= flushInterval; i++ {
+		s = strconv.FormatInt(i, 10)
+		store.Set([]byte("key"+s), []byte("value"+s))
+		_, lastVersion, err = store.tree.SaveVersion()
+		require.NoError(t, err)
+		require.Equal(t, i, lastVersion)
+	}
+	for i := lastVersion; i > 0; i-- {
+		it, err := store.tree.GetImmutable(i)
+		require.NoError(t, err)
+		require.NotNil(t, it)
+		t.Logf(it.String())
+	}
 }
 
 func testFlush(t *testing.T) {


store/iavlstore_test.go
```diff
@@ -144,6 +146,43 @@ func testNormal(t *testing.T) {
...
+		for i := int64(1); i <= flushInterval*10000; i++ {
+			s = strconv.FormatInt(i, 10)
+			store.Set([]byte("key"+s), []byte("value"+s))
+
+			// Broken Unpredictably
+			_, lastVersion, err = store.SaveVersion(nil)
+
+			// Works fine
+			// _, lastVersion, err = store.tree.SaveVersion()
+			ch <- lastVersion
```
enlight: I want this test to only load the versions before the ones flushed out, i.e. only versions that satisfy the condition:
```golang
version % flushInterval == (flushInterval - 1)
```

atchapcyp:
rebuild sesame

store/iavlstore.go
```diff
@@ -161,6 +162,10 @@ func (s *IAVLStore) SaveVersion(opts *VersionedKVStoreSaveOptions) ([]byte, int6
 	if flushInterval == 0 || ((oldVersion+1)%flushInterval == 0) {
 		if flushInterval != 0 {
 			log.Info("[IAVLStore] Flushing mem to disk", "version", oldVersion+1)
+			s.previousTree, err = s.tree.GetImmutable(oldVersion)
```
enlight: Probably need to check that `oldVersion > 0` first.
Also `s.previousTree` must be accessed atomically, similarly to the `lastSavedTree` in the MultiWriterAppStore

store/multi_writer_app_store.go
```diff
@@ -239,7 +239,13 @@ func (s *MultiWriterAppStore) GetSnapshotAt(version int64) (Snapshot, error) {
 	var appStoreTree *iavl.ImmutableTree
 	if version == 0 {
 		appStoreTree = (*iavl.ImmutableTree)(atomic.LoadPointer(&s.lastSavedTree))
+		fmt.Println("Get latest tree of version ", appStoreTree.Version())
+	} else if s.appStore.previousTree != nil &&
+		(*iavl.ImmutableTree)(s.appStore.previousTree).Version() == version {
```
enlight: `previousTree` must be loaded using `atomic.LoadPointer` first

store/iavlstore.go
```diff
@@ -201,6 +209,23 @@ func (s *IAVLStore) Prune() error {
 	return nil
 }
 
+func (s *IAVLStore) setPreviousTree(version int64) error {
+	var err error
+	var tree *iavl.ImmutableTree
+
+	if version == 0 {
+		tree = iavl.NewImmutableTree(nil, 0)
```
enlight: I think it's probably better to leave the previousTree `nil` in this case

store/iavlstore_test.go
```diff
@@ -144,6 +147,81 @@ func testNormal(t *testing.T) {
...
+
+	}
+	flushedVersion = latestVersion
+	store.Set([]byte("k"), []byte("v"))
+	_, latestVersion, err = store.SaveVersion(nil)
+	require.NoError(t, err)
+	//
+	it, err := store.tree.GetImmutable(flushedVersion)
+	require.NoError(t, err)
+	require.Equal(t, flushedVersion, it.Size())
```
enlight: Why is this `it.Size()` check here?

store/iavlstore_test.go
```diff
@@ -144,6 +147,81 @@ func testNormal(t *testing.T) {
...
+
+	// Trying to retrieve a tree on one version after flushed version.
+	it, err = store.tree.GetImmutable(latestVersion)
+	require.NoError(t, err)
+	require.Equal(t, latestVersion, it.Size())
```
enlight: And here?

store/iavlstore_test.go
```diff
@@ -144,6 +147,81 @@ func testNormal(t *testing.T) {
...
+
+	// Trying to retrieve a tree on one version before flushed version.
+	it, err = store.tree.GetImmutable(flushedVersion - 1)
+	require.EqualError(t, err, "version does not exist")
+	require.Nil(t, it)
+
+	diskDb.Close()
```
enlight: Should probably do `defer diskDB.Close()` right after creating `diskDB`

store/iavlstore_test.go
```diff
@@ -144,6 +147,81 @@ func testNormal(t *testing.T) {
 	diskDb.Close()
 }
 
+func testGetTreeAfterFlush(t *testing.T) {
+	diskDb := getDiskDb(t, "testGetTreeAfterFlush")
```
enlight: -> `diskDB`

store/iavlstore_test.go
```diff
@@ -144,6 +147,81 @@ func testNormal(t *testing.T) {
...
+	// Trying to retrieve a tree on one version before flushed version.
+	it, err = store.tree.GetImmutable(flushedVersion - 1)
+	require.EqualError(t, err, "version does not exist")
+	require.Nil(t, it)
+
+	diskDb.Close()
+}
+
+func testGetPreviousTree(t *testing.T) {
+	diskDb := getDiskDb(t, "testGetPreviousTree")
```
enlight: -> `diskDB`

store/multi_writer_app_store_test.go
```diff
@@ -59,6 +60,111 @@ func (m *MultiWriterAppStoreTestSuite) TestMultiWriterAppStoreSnapshotFlushInter
...
+		require.NoError(err)
+		require.Equal(i, latestVersion)
+	}
+
+	// make sure previousTree has been set after flush store to disk.
+	require.NotNil((*IAVLStore)(store.appStore.previousTree))
+	_, err = store.GetSnapshotAt(4)
+	require.NoError(err)
+
+	snap, err := store.GetSnapshotAt(0)
```
enlight: Add a comment to explain what this does

store/iavlstore_test.go
atchapcyp: My intention is to check a number of elements in the Immutable tree.
Since we set only one key-value on each SaveVersion, Tree version should equal to numbers of the element.
Is this irrelevant to the test?

enlight: Relevant enough I suppose, add a comment to the test to explain this.

Issue #1624, opened 2020-03-19
Add account_balances RPC method to unsafe RPC interface on nodes


enlight:
The `account_balances` RPC method should accept a list of contract addresses (on the DAppChain) as input. `eth` and `loom` should be accepted as aliases for the built-in contracts, `eth` will be resolved to `ethcoin`, and `loom` to `coin`.

The RPC method should:
1. Retrieve all the account mappings from the address mapper contract.
2. For each account mapping query how many tokens the DAppChain account has for each of the specified contracts.
3. Return a plain JSON response containing all the balances in this format:
   ```json
   "accounts": {
     "eth:0xdeadbeef": {
       "eth": "1000000000",
       "loom": "20000",
       "eth:0xwhatever": "0"
     },
     "eth:0xdeadfeeb": {
       "eth": "0",
       "loom": "10000000",
       "eth:0xwhatever": "100"
     }
   }
   ```
   Aside from `eth` and `loom` the contract addresses in the response should be the foreign address, this means the input contract addresses will need to be mapped to a foreign address via the contract mappings stored in the gateway contract.

PR #1625, opened 2020-03-19
Add Unsafe account_balances RPC method

atchapcyp:
ref: https://github.com/loomnetwork/loomchain/issues/1624

### DIFF
--- a/rpc/query_server.go
+++ b/rpc/query_server.go
@@ -686,6 +686,13 @@ func (s *QueryServer) GetCanonicalTxHash(block, txIndex uint64, evmTxHash eth.Da
 	return eth.EncBytes(blockResult.Block.Data.Txs[index].Hash()), nil
 }
 
+type AccountsBalanceResponse struct {
+}
+
+func (s *QueryServer) GetAccountBalance(address string) (*AccountsBalanceResponse, error) {
+	return nil, errors.Errorf("GetAccountBalance Called")
+}
+
 // Takes a filter and returns a list of data relative to transactions that satisfies the filter
 // Used to support eth_getLogs
 // https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_getlogs
--- a/rpc/query_service.go
+++ b/rpc/query_service.go
@@ -65,6 +65,7 @@ type QueryService interface {
 	GetContractRecord(contractAddr string) (*types.ContractRecordResponse, error)
 	DPOSTotalStaked() (*DPOSTotalStakedResponse, error)
 	GetCanonicalTxHash(block, txIndex uint64, evmTxHash eth.Data) (eth.Data, error)
+	GetAccountBalance(address string) (*AccountsBalanceResponse, error)
 
 	// deprecated function
 	EvmTxReceipt(txHash []byte) ([]byte, error)
@@ -213,7 +214,7 @@ func MakeEthQueryServiceHandler(logger log.TMLogger, hub *Hub, routes map[string
 }
 
 // MakeUnsafeQueryServiceHandler returns a http handler for unsafe RPC routes
-func MakeUnsafeQueryServiceHandler(logger log.TMLogger) http.Handler {
+func MakeUnsafeQueryServiceHandler(svc QueryService, logger log.TMLogger) http.Handler {
 	codec := amino.NewCodec()
 	mux := http.NewServeMux()
 	routes := map[string]*rpcserver.RPCFunc{}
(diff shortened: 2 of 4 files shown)


rpc/query_server.go
```diff
@@ -686,6 +686,13 @@ func (s *QueryServer) GetCanonicalTxHash(block, txIndex uint64, evmTxHash eth.Da
 	return eth.EncBytes(blockResult.Block.Data.Txs[index].Hash()), nil
 }
 
+type AccountsBalanceResponse struct {
+}
+
+func (s *QueryServer) GetAccountBalance(address string) (*AccountsBalanceResponse, error) {
```
enlight: -> `GetAccountBalances(contracts []string)`

rpc/query_service.go
```diff
@@ -226,6 +227,8 @@ func MakeUnsafeQueryServiceHandler(logger log.TMLogger) http.Handler {
 	routes["unsafe_stop_cpu_profiler"] = rpcserver.NewRPCFunc(rpccore.UnsafeStopCPUProfiler, "")
 	routes["unsafe_write_heap_profile"] = rpcserver.NewRPCFunc(rpccore.UnsafeWriteHeapProfile, "filename")
 
+	routes["account_balances"] = rpcserver.NewRPCFunc(svc.GetAccountBalance, "contract_address")
```
enlight: -> `rpcserver.NewRPCFunc(svc.GetAccountBalance, "contracts")`

Which should then be queried like so:
```
curl 'localhost:26657/accounts_balances?contracts=["eth","loom","0xdeadbeef"]'
```

builtin/plugins/address_mapper/address_mapper.go
```diff
@@ -250,4 +250,28 @@ func SignIdentityMapping(from, to loom.Address, key *ecdsa.PrivateKey, sigType e
 	return evmcompat.GenerateTypedSig(hash, key, sigType)
 }
 
+func ListMapping(ctx contract.StaticContext) (*ListMappingResponse, error) {
```
enlight: No need to duplicate this function, can call existing one like so:
```golang
import am "github.com/loomnetwork/go-loom/builtin/types/address_mapper"
mapper := &am.AddressMapper{}
r, err := mapper.ListMapping(ctx)
```

builtin/plugins/coin/coin.go
```diff
@@ -423,6 +423,11 @@ func (c *Coin) transferFrom(ctx contract.Context, req *TransferFromRequest) erro
 	return emitTransferEvent(ctx, from, to, &amount)
 }
 
+func BalanceOf(ctx contract.StaticContext, owner loom.Address) (*loom.BigUInt, error) {
```
enlight: Don't really need this, can just use the existing function:
```golang
c := &coin.Coin{}
resp, err := c.BalanceOf(blah)
```

rpc/query_server_gateway.go
```diff
@@ -0,0 +1,132 @@
...
+		return nil, err
+	}
+
+	gatewayCtx, err := s.createStaticContractCtx(snapshot, "gateway")
+	if err != nil {
+		return nil, err
+	}
+
+	accounts := make(map[string]map[string]string, len(resp.Mappings))
+	for _, mp := range resp.Mappings {
```
enlight: Since mappings are bi-directional there'll be 2 entries for each account, so you'll be fetching all the balances twice. Should check if `accounts` already has an entry for this account before fetching balances.

rpc/query_server_gateway.go
```diff
@@ -0,0 +1,132 @@
...
+		return nil, errors.Wrapf(err, "error getting balance of %s", address.Local.String())
+	}
+	if amount == nil {
+		return glcommon.BigZero(), nil
+	}
+
+	return amount, nil
+}
+
+func getLoomBalance(ctx contractpb.StaticContext, address loom.Address) (*gtypes.BigUInt, error) {
```
enlight: This should just return `glcommon.BigUInt` like `getEthBalance`, don't see any reason for them to be inconsistent

rpc/query_server_gateway.go
```diff
@@ -0,0 +1,132 @@
...
+			localAddr = loom.UnmarshalAddressPB(mp.To)
+			foreignAddr = loom.UnmarshalAddressPB(mp.From)
+		}
+		account := make(map[string]string, len(resp.Mappings))
+		for _, contract := range contracts {
+			switch contract {
+			case "eth":
+				ethBal, err := getEthBalance(ethcoinCtx, localAddr)
+				if err != nil {
+					account[contract] = err.Error()
```
enlight: Return the errors rather than continuing, otherwise the caller would have to guess if the account balance is a valid number or an error.

rpc/query_server_gateway.go
```diff
@@ -0,0 +1,132 @@
...
+				erc20Ctx := gateway.NewERC20StaticContext(gatewayCtx, loom.UnmarshalAddressPB(erc20ContractAddr))
+				erc20Bal, err := gateway.BalanceOf(erc20Ctx, localAddr)
+				if err != nil {
+					account[contract] = err.Error()
+					continue
+				}
+				account[contract] = erc20Bal.String()
+			}
+		}
+		accounts[foreignAddr.ChainID+":"+foreignAddr.Local.String()] = account
```
enlight: -> `accounts[foreignAddr.String()] = account`

PR #1629, opened 2020-04-09
Update chain-cfg list-validators cmd to show individual validator powers

enlight:
The power displayed by this command for each validator is the percentage of the validator's current voting power out of the total combined voting power of all currently active validators (known to Tendermint).


enlight:
rebuild sesame

enlight:
rebuild sesame

enlight:
rebuild sesame

enlight:
rebuild sesame

PR #1638, opened 2020-06-04
parselintreport.sh: Grammar fixes

andreipope:


adyromantika:

PR #1643, opened 2020-06-12
address_mapper.go: Improve comments

andreipope:


adyromantika:

builtin/plugins/address_mapper/address_mapper.go
```diff
@@ -68,8 +65,7 @@ func (am *AddressMapper) Init(ctx contract.Context, req *InitRequest) error {
 	return nil
 }
 
-// AddIdentityMapping adds a mapping between a DAppChain account and a Mainnet account.
-// The caller must provide proof of ownership of the Mainnet account.
+// AddIdentityMapping adds a mapping between a Loom account and a foreign account. The caller must provide proof of ownership of the foreign account.
```
andreipope, line 68: @enlight Quick question: is this how we should name them ("Loom account" and "foreign account")?

enlight: Should probably use more consistent naming... local account and foreign account.

andreipope: @enlight Gotcha, thanks!

PR #1664, opened 2020-07-06
plasma_cash/config/config.go: Grammar

andreipope:
plasma_cash/config/config.go: Grammar


enlight:
Plasma cash code is ancient, not worth changing now.

PR #1666, opened 2020-07-08
plasma_cash/plasma_cash.go: Grammar

andreipope:
plasma_cash/plasma_cash.go: Grammar


enlight:
Plasma cash code is ancient, not worth changing now.

PR #1704, opened 2021-05-16
fix parsing on GetEvmBlockByNumber

NightingaleAsh:


enlight (approved):
Probably a bit too lenient now, but that could be fixed later.

PR #1705, opened 2021-05-17
Add ability to copy out the currently used state from the app DB to a new DB

enlight:
To copy the currently used state at a particular height set `AppStore.CloneStateAtHeight` in the on-chain config, when the app reaches that height it will copy out the current state from the `app.db` to `app_v2.db` (omitting ancient keys that are no longer needed), and will save all further IAVL tree versions to the new `app_v2.db`.

Each time a node starts it will check if `app_v2.db` contains the most recent state, if it does it will rename:
1. `app.db` -> `old_app_timestamp.db`
2. `app_v2.db` -> `app.db`
And will then proceed to use `app.db` as usual.

Depends on:
https://github.com/loomnetwork/iavl/pull/15
https://github.com/loomnetwork/go-loom/pull/477

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -300,12 +300,12 @@
   version = "v0.14.0"
 
 [[projects]]
-  branch = "mutable-tree-with-version"
+  branch = "loomchain2"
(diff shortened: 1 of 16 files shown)


NightingaleAsh:
👍

NightingaleAsh (approved):
looks good thanks

